PR #62, opened 2015-09-30
Correct fortios to uppercase in driver_mapping()

StefanLindblom:


dbarrosop:
Thanks! That's already fixed in the 0.2 version which is about to be released:

https://github.com/spotify/napalm/blob/0.2/napalm/__init__.py

So no need of merging this PR :)

PR #64, opened 2015-10-07
Make Junos timeout configurable

ogarraux:
We have SRX clusters that take a long time to commit (~90+ seconds).  This change, along with host variables in ansible and some tweaks to napalm_install_config let us override the timeout on a per-host basis.  

This would make the arguments for the Junos driver different than the other drivers, which might be a bit ugly, though nothing should break since there's a 60 second default.

I haven't included the napalm_install_config changes here, since they always send a timeout argument...which won't work for anything other than this Junos driver.  (We're only using NAPALM for Junos at the moment).

### DIFF
--- a/napalm/junos.py
+++ b/napalm/junos.py
@@ -27,16 +27,17 @@ from utils import string_parsers
 
 class JunOSDriver(NetworkDriver):
 
-    def __init__(self, hostname, username, password):
+    def __init__(self, hostname, username, password, timeout=60):
         self.hostname = hostname
         self.username = username
         self.password = password
+        self.timeout = timeout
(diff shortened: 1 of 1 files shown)


fooelisa:
Great idea! @dbarrosop and I are looking into adding timeout support for the other vendors, in the spirit of a unified interface ;) We'll merge this in with the other changes soon!

dbarrosop:
@ogarraux could you actually cancel the PR and send it to the branch 0.30 instead of to the master? As @fooelisa mentioned we will add support for the rest of vendors as well.

Thanks!

ogarraux:
Sure, will do.  Thank you!

PR #65, opened 2015-10-09
Make Junos timeout configurable

ogarraux:
Update of #64 against the 0.30 branch.


ogarraux:
Given your comments about adding a timeout argument to the other drivers, I've also added my changes to napalm_install_config here.

PR #66, opened 2015-10-09
Module for IBM/Lenovo RackSwitch series.

kderynski:
This is the first version of module for IBM/Lenovo RackSwitch series. It uses netconf via library provided by Lenovo (mentioned in official manual). Module supports only configuration management features for now. There are couple of issues with netconf implementation on these devices and it will be described in documentation. Unit tests have been performed. Obviously there are 3 errors (getters are not implemented yet) and 1 error for "test_merge_configuration_typo_and_rollback". I can detect syntax error only during commit operations. I do not have problem with "test_replacing_config_with_typo" because commit statement is under "try" in this test. Maybe it should be under "try" also in case of merge ?
I am open for discussion :)

### DIFF
--- a/test/unit/TestIBMDriver.py
+++ b/test/unit/TestIBMDriver.py
@@ -12,25 +12,22 @@
 # License for the specific language governing permissions and limitations under
 # the License.
 
-from eos import EOSDriver
-from iosxr import IOSXRDriver
-from junos import JunOSDriver
-from fortios import FortiOSDriver
-from ibm import IBMDriver
+import unittest
 
-def get_network_driver(vendor):
-    driver_mapping = {
-        'EOS': EOSDriver,
-        'ARISTA': EOSDriver,
-        'IOS-XR': IOSXRDriver,
-        'IOSXR': IOSXRDriver,
-        'JUNOS': JunOSDriver,
-        'JUNIPER': JunOSDriver,
-        'FORTIOS': FortiOSDriver,
-        'IBM': IBMDriver,
-    }
-    try:
-        return driver_mapping[vendor.upper()]
-    except KeyError:
-        raise Exception('Vendor/OS not supported: %s' % vendor)
+from napalm.ibm import IBMDriver
+from base import TestNetworkDriver
 
+
+class TestIBMDriver(unittest.TestCase, TestNetworkDriver):
+
+    @classmethod
+    def setUpClass(cls):
+        hostname = '10.20.18.253'
+        username = 'admin'
+        password = 'admin'
+        cls.vendor = 'ibm'
+
+        cls.device = IBMDriver(hostname, username, password, timeout=60)
+        cls.device.open()
+        cls.device.load_replace_candidate(filename='%s/initial.conf' % cls.vendor)
+        cls.device.commit_config()
--- a/docs/ibm.rst
+++ b/docs/ibm.rst
@@ -1,0 +1,13 @@
+IBM Networking Operating System
+-------
+
+Rollback
+~~~~~~~~
+
+Rollback is simply implemented by reading current running configuration before any load actions. Rollback function executes load replace and commit.
+
+
+Atomic Changes
+~~~~~~~~~~~~~~
+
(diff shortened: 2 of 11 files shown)


dbarrosop:
Great work here : )

Could you add the "caveats" in the documentation, please? I would like to have it all together in the same PR so we don't forget it.

Regarding the unit tests, not sure if I understood your issue with the test **test_replacing_config_with_typo**. Or most probably the base wasn't clear enough with what you had to do there : )

You have to raise the exception *ReplaceConfigException* or *MergeConfigException* when doing the *load* or the *commit* operation. Some devices can detect the issue during the load and some others during the commit. That is expected. The idea of the test is to make sure the driver can figure out that there was something wrong with the configuration and that no changes were performed. That's why the "except" block retrieves compares the running config with the initial config. To make sure no change was performed.

kderynski:
Thanks :) I've added ibm.rst file to the documentation and it contains all my comments.

Regarding to unit tests it seems that I understood correctly the test named **test_replacing_config_with_typo** and I throw *ReplaceConfigException* under commit function. However I have a problem with  **test_merge_configuration_typo_and_rollback** because I throw *MergeConfigException* but there is no commit function under try in this test as you can see below:

    
    def test_merge_configuration_typo_and_rollback(self):
        result = False
        try:
            self.device.load_merge_candidate(filename='%s/merge_typo.conf' % self.vendor)
            diff = self.device.compare_config()
            raise Exception("We shouldn't be here")
        except exceptions.MergeConfigException:
    
Test named **test_replacing_config_with_typo** has commit under try statement:

    
 def test_replacing_config_with_typo(self):
        result = False
        try:
            self.device.load_replace_candidate(filename='%s/new_typo.conf' % self.vendor)
            self.device.commit_config()
    
Correct me if I'm wrong but in my opinion there should be commit function also in the first test function :)

dbarrosop:
You are 100% right! Do you want to implement your proposed change? Looks good to me. Good thing you spotted that : )

kderynski:
Good to hear that :) Yes, I would like to implement this change and after that I will test module one more time. Currently I have limited access to the network, so I will commit this change at the end of this week.

dbarrosop:
Thanks!

test/unit/TestIBMDriver.py
```diff
@@ -0,0 +1,33 @@
...
+# License for the specific language governing permissions and limitations under
+# the License.
+
+import unittest
+
+from napalm.ibm import IBMDriver
+from base import TestNetworkDriver
+
+
+class TestIOSXRDriver(unittest.TestCase, TestNetworkDriver):
```
thomas-mangin: pedantic, should really be called IBM and not IOSXR (copy and paste :wink:)

kderynski: @thomas-mangin you are right. copy and past will kill me someday ;) Thanks for letting me know :)

kderynski:
I have added this small fix to the merge unit test. It works as expected. Now I have only 3 errors from getters (not implemented). I have fixed also two additional issues in my code (variable initialization and copy&paste bug ;-)

dbarrosop:
I have a few comments/questions:

1. Can you squash all commits in your PR to a single one? I want to start keeping the history a bit cleaner. Squashing commits is something we have been very bad at until know : )
1. I am super surprised on how they implement the "replace" operation. Is this something you could discuss with them? Happy to help, I just don't have any contact.

If you can squash all the commits into one I will merge the PR : )

kderynski:
Answers below: 
1. I can squash all commits. I have never done it before but git rebase function looks quite straightforward.
2. I have to modify library provided by Lenovo for timeout support. I am looking at their code and I am wondering where I could do this (probably I will pass the timeout value to paramiko functions). I have also  concerns about timeouts in paramiko. I have read in documentation that timeout is related to TCP connection only and by default is 3600s. Long story short I will try to do this but I do not know how it will work at least for now ;)
3. Definitely I will try to discuss netconf implementation on these devices. I have already started a discussion with Lenovo but without any results so far. 

If you have any hints about timeouts in paramiko library please let me know :)

dbarrosop:
1. Let me know if you need help.
2. I tweaked pyfg library (which uses paramiko as well) to accept the timeout keyword a few days ago. You can check what I did in the following link: https://github.com/spotify/pyfg/commit/600df70f700f454c3b49c766ae98624f0048381f.
Worst case you can just accept the timeout parameter and just ignore it. We could fix it later.
3. Same as 1.

kderynski:
I tried to squash commits but I screwed up something badly. Sorry for the mess with this PR. I definitely need help :-) Any hints how to get out from this mess will be appreciated.  I checked also your code in pyfg and there is slightly different situation inside library from Lenovo. They used Transport class with start_client() method and there is no timeout parameter. I will re-implement this library but until that time I would like to accept timeout and ignore it.

dbarrosop:
I will try to be around on IRC to help out with the PR.

PR #73, opened 2015-10-28
bypass diff generation in Ansible module for install_config

jedelman8:
Added a parameter to bypass generating diffs because some platforms don't support this, i.e. the Arista version I'm using 4.14.7M.  So, setting `get_diffs` to `False` will always apply the config and not attempt to "compare" the configs.

### DIFF
--- a/ansible/napalm_install_config
+++ b/ansible/napalm_install_config
@@ -64,13 +64,13 @@ options:
     diff_file:
         description: A file where to store the "diff" between the running configuration and the new configuration. If
                      it's not set the diff between configurations is not saved.
-        choices: ['true', 'false']
         required: False
     get_diffs:
         description:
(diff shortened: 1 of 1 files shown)


fooelisa:
Awesome, thanks for tinkering around :) 
Would anything speak against checking the diff_file parameter and assume no file provided means no diffing? Just to spare the additional parameter. What do you think?

jedelman8:
That was going to be my first option too, but maybe some people want diffs without a file? Totally open.

dbarrosop:
I think it's cleaner if we have a different variable. As it's been pointed out, maybe you don't care about the diff itself but you want to check if there was some change. The variable ``get_diffs`` disables that check so it would prevent that behavior.

PR #79, opened 2015-10-29
Timeout, documentation revamp and IBMDriver (beta)

dbarrosop:


napalm/__init__.py
```diff
@@ -16,6 +16,7 @@
 from iosxr import IOSXRDriver
 from junos import JunOSDriver
 from fortios import FortiOSDriver
+#from ibm import IBMDriver
```
fooelisa: this should not be commented out?

napalm/__init__.py
```diff
@@ -26,6 +27,7 @@ def get_network_driver(vendor):
         'JUNOS': JunOSDriver,
         'JUNIPER': JunOSDriver,
         'FORTIOS': FortiOSDriver,
+#        'IBM': IBMDriver,
```
fooelisa: this should not be commented out?

napalm/iosxr.py
```diff
@@ -23,11 +23,12 @@
 
 class IOSXRDriver(NetworkDriver):
 
-    def __init__(self, hostname, username, password):
+    def __init__(self, hostname, username, password, timeout=60):
         self.hostname = hostname
         self.username = username
         self.password = password
-        self.device = IOSXR(hostname, username, password)
+        self.timeout = timeout
```
fooelisa: I guess there is not really a reason to assign timeout to self here, as I am only using it right there

napalm/__init__.py
dbarrosop: As the commit message says: IBMDriver disabled because documentation and requirements are missing.

Because the requirements are not complete importing napalm fails.

napalm/iosxr.py
dbarrosop: Maybe someone wants to check it? It doesn't hurt.

napalm/__init__.py
fooelisa: Ok, just making sure we won't merge it like this

kderynski: I'll try to create pip package for the Lenovo/IBM library used in my module. After that we can easily add it to requirements file. I'll also prepare documentation. It should be ready soon ~around weekend :)

dbarrosop: Great! thanks! For the documentation I just need to know where to get the library from (your repo URL) and the minimum OS version, I think what you did already is good enough at least to release now. More documentation will be good but I don't want to block the release : )

kderynski: Ok, great :) Library with my bug fix is available on https://github.com/kderynski/blade-netconf-python-client , any unix like system with python 2.7 should be enough. If I understand correctly you will add this info or should I do this ?

fooelisa: It would be great if this was available via pip and added to the requirements file, so the dependency will be installed automagically. Is the original client available via pip?

kderynski: Unfortunately it isn't available via pip and it seems library is abandoned (official information from the vendor). That is why I would like to make pip package. BTW I think I will rewrite module completely in the future because current one is prepared as cli tool not a library :)

fooelisa: Yeah, make a pip yourself then, like we did with pyiosxr, and we'll reference it in the requirements. I followed this guide: http://peterdowns.com/posts/first-time-with-pypi.html

dbarrosop: As @fooelisa said, package it with pip and I will fix the requirements and the link. Don't worry about that. By the way, with minimum OS version I meant the network operating system that the device has to run.

kderynski: @fooelisa thanks, I will try to use this guide :) 
@dbarrosop  Sorry for misunderstanding, module has been tested on 7.11.2 version and it should work on all 7.x.x version. There is newest branch with number 8.x but it was prepared only for Lenovo branded device. I'm sure that all from 7.x.x will work but I have to confirm what about 8.x :)

kderynski: Package has been created (bnclient). It is available now via pip.

PR #82, opened 2015-11-15
First draft at NXOS module

mzbenami:
Hello all,

This is a rough shot at an NXOS module. Although all test cases pass, there are some issues, and I'm willing to make whatever changes you suggest.

Issues: 
* As you can see in the ``test/unit/nxos/`` directory, the ``.conf`` files look a little weird. They aren't normal config files but special "checkpoint" files. That's because on NXOS the only way to replace a config without reboot is to rollback to a checkpoint (which could be a file). But these files explicitly list a lot of normally implicit config lines, some of them starting with ``!#``. The ``!#`` isn't necessary for the rollback to work, but leaving these lines out can cause erratic behavior. For example, leaving out a ``shutdown`` or ``no shutdown`` line will cause the switch to toggle the up/down state of an interface, depending on it's current state. The normal expected behavior would be to have no effect, I think. Another example is that `` !#switchport trunk allowed vlan 1-4094`` is required even if the switchport is in mode ``switchport access``. However if `` !#switchport trunk allowed vlan 1-4094`` is included with ``no switchport``, the rollback will fail.

* NXOS doesn't support candidate configs, but it does support taking a diff of a checkpoint file compared to the current running config. The diffs consist of the commands that would be entered to take the switch from it's current state to the checkpoint file's state. Take a look at ```test/unit/nxos/new_good.diff`` as an example. As you can see, it's a very unituitive format, and there seem to be a lot of unnecessary/redundant lines.

* Merges are currently implemented by simply applying the the merge config line by line. This doesn't use the checkpoint/rollback functionality, so there is no native diff support. As of now, if a merge config is loaded, and a ``compare_config()`` is run, you'll simply get a ``MergeConfigException``. The file ``merge_good.diff`` represents the diff that would take the switch from the state it would be in after the lines in ``merge_good.conf`` were executed to the initial state. As you can see, the diff is way larger than the the two lines in ``merge_good.conf``.

Is this checkpoint file format feasible?
Is there any advice on how to go about diffs differently, given the precious nature of checkpoint files and the diffs generated by the switch?


dbarrosop:
Awesome job!

> Is this checkpoint file format feasible?

Yes. I don't see why not. Just make sure it's documented. I have two questions though. Are the checkpoint number and the time importantly? Can they be faked?

> Is there any advice on how to go about diffs differently, given the precious nature of checkpoint files and the diffs generated by the switch?

Keep it as it is. That's what they provide. If someone has complaints they should be brought to Cisco not to us. However, the merge diff, if you can't fabricate it return the same commands you were provided with. Don't raise an exception as that might break some code.

Could you add documentation explaining all the tiny details (caveats)? What you put already in this PR will be a good starting point. Could you also fill the table that summarizes what is supported and if changes are atomic, etc?

The unit tests, were they run against a physical device or a virtual one?

As I said before, awesome job! This is going to be big. NX-OS is a quite popular OS.

jedelman8:
Very nice work!  

This is only from an Ansible perspective, but worth pointing out as a reminder.  Regarding calculating diffs (or not calculating them in this case for merges)...  I ran into this a few weeks ago with Arista since I was running an older version of EOS that didn't support diffs, and added the `get_diffs` param to the **napalm_install_config** module to mitigate `compare_config()` from ever being called. 

Really looking forward to using this :)

mzbenami:
Since my last note:

* Added documentation
* ``compare_config()`` for merges returns the merge_candidate (no test case failures)
* ``get_checkpoint_file()`` function for convenience
* fixed issues when replacement config is string instead of file

dbarrosop:
Excellent work!

PR #87, opened 2015-11-27
Major enhancements to tests

dbarrosop:
* Now you can test configuration methods and getter methods independently.
* Configuration methods can't be mocked. Main reason is because I think it's interesting to test how the code interacts with the internals of a real device.
* Getter methods can be run against a live device or against a mocked device. Just set the mock flag in the driver test to whatever you want to do.
* To mock a device call is as simple as just creating a file inside mock_data where filename is the call you are mocking and the content is whatever the real device would respond.
* Juniper can be equally mocked and it doesn't matter if you are using RPC calls or the junos_views functionality in your code. You just have to mock the RPC call you are making and that will work for both raw RPC calls or for the junos_views wrapper.


kderynski:
Great idea ! I think this approach will speed up work on getters part :)

PR #88, opened 2015-11-27
Started adding IOS support

GGabriele:
Hi there.
This started as an 'exercise' for myself but I decided to make the PR since I'd like to contribute to this awesome project.

Working with IOS is a pain since it doesn't have native API, but I tried to do my best to meet the project requirements. It's far to be completed yet, but if you think this could be useful I'll keep working on it (like adding compare function and fine tuning the whole thing).

Let me know :)

### DIFF
--- a/napalm/ios.py
+++ b/napalm/ios.py
@@ -108,8 +108,7 @@ class IOSDriver(NetworkDriver):
                     output = self.device.send_command(command)
                     new_prompt = self.device.find_prompt()
                     if (index > 2 and self.first_touch) or (self.first_touch is False):
-                        if '^' in output:
-                            self.candidate_config = commands_dict
+                        self.candidate_config = commands_dict
                         if new_prompt == prompt or 'hostname' in command:
                             commands_dict[command] = "unchanged"
                         else:
--- a/napalm/ios.py
+++ b/napalm/ios.py
@@ -38,7 +38,7 @@ class IOSDriver(NetworkDriver):
         """Closes the connection to the device."""
         self.device.disconnect() 
 
-    def discard_config(self):
+    def rollback(self):
         if self.candidate_config is not None:
             if isinstance(self.candidate_config, dict):
                 for command in self.candidate_config_commands:
@@ -86,8 +86,12 @@ class IOSDriver(NetworkDriver):
                 for command in commands:
                     print "+ {0}".format(command)
             else:
-                for command in commands:
+                for command in self.candidate_config:
                     print "+ {0}".format(command)
+
+    def discard_config(self):
+        if self.candidate_config is not None:
+            self.candidate_config = None
 
     def commit_config(self):
         if self.candidate_config is not None:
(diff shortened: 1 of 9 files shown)


dbarrosop:
Hello @GGabriele,
Great job! Of course I think it would be useful : ) The only comment I will make about the code is don't do things like:

     def compare_config(self):
         pass

As you are not supporting config_replace I would suggest for starters to return whatever the config_merge is doing. Improving on that and returning the changes would be nice but given the limitations of the platform I am happy with compare_config returning just that. Either that or don't implement the method at all. That will trigger a NotImplementedException allowing the user of napalm to modify the workflow of his/her application depending on if a method for a particular platform is implemented or not.

Regards
David

GGabriele:
Ehy @dbarrosop,
I made some changes as you suggested. Now the workflow is this:

- load_merge_candidate loads a candidate configuration (without sending/applying any commands)

- compare_config shows what changes would be applied from load_merge

- commit_config sends commands and save running configuration

- doing so, discard_config would simply look like a list of commands not yet applied and if it is called it empties the list. Anyway, if any error occours during the commit (like a typo into the configuration) the discard_config function is called and it will remove the commands that have been already applied with the 'no' prefix before the actual command

I've also added the get_lldp_neighbor function. Anyway, I did all my tests in GNS3 and since I didn't manage to find an IOS image supporting LLDP I used a textfile with a sample 'show lldp neighbors' output for my test.

napalm/ios.py
```diff
@@ -0,0 +1,435 @@
...
+            device_type='cisco_ios',
+            ip=self.hostname,
+            username=self.username,
+            password=self.password)
+
+    def close(self):
+        """Closes the connection to the device."""
+        self.device.disconnect() 
+
+    def discard_config(self):
```
dbarrosop: Quick question about this method. Are you trying to revert changes after they have been committed? If yes, you should do that with the rollback method. Discard should just discard the candidate configuration before commit. For example:

    device.open()
    device.load_merge_configuration(filename='my_changes.txt')
    # I changed my mind and I don't want to commit:
    device.discard_config()
    
    # Let's try this other instead
    device.load_merge_configuration(filename='my_other_changes.txt')
    device.commit()
    
    # Oh shit! I broke it : )
    device.rollback()
    device.close()

dbarrosop:
Could you work on some documentation as well? Something that explains the specifics of the driver you wrote (considerations, caveats, anything you would like to know if you weren't the author and wanted to use the library). You can check others drivers' caveats.

GGabriele:
Okay @dbarrosop, it definitely makes sense. 
I'm fixing it and adding docs :)

dbarrosop:
Thanks! : )

GGabriele:
I made the needed changes to the code and added some docs.
Let me know if everything is fine and if you need anything else :)

dbarrosop:
Awesome work : )

fooelisa:
Very nice!! :) +1

GGabriele:
Thank you :) happy to contribute

Issue #77, opened 2015-10-29
timeout on ibm does nothing


fooelisa:

kderynski:
We can close this one. Issue has been resolved :)

fooelisa:
Nice!! Thanks

PR #89, opened 2015-11-28
timeout implemented in IBM module

kderynski:
I have added timeout parameter to bnclient library, so now it can be used in napalm module for IBM/Lenovo devices. It solves #77 bug :)


dbarrosop:
Awesome! : )

PR #111, opened 2015-12-08
Adding get_bgp_neighbors to eos.py

marcovespignani:
Change-Id: I15d7dd4f7e282ea6f6cadbec2d1a694aec4c6fd0

### DIFF
--- a/napalm/eos.py
+++ b/napalm/eos.py
@@ -237,48 +237,94 @@ class EOSDriver(NetworkDriver):
         return interface_counters
 
 
+    def get_bgp_neighbors(self):
 
-    # def get_bgp_neighbors(self):
-    #     bgp_neighbors = dict()
-    #
-    #     for vrf, vrf_data in self.device.show_ip_bgp_summary_vrf_all()['vrfs'].iteritems():
-    #         bgp_neighbors[vrf] = dict()
-    #         bgp_neighbors[vrf]['router_id'] = vrf_data['routerId']
-    #         bgp_neighbors[vrf]['local_as'] = vrf_data['asn']
-    #         bgp_neighbors[vrf]['peers'] = dict()
-    #
-    #         for n, n_data in vrf_data['peers'].iteritems():
-    #             bgp_neighbors[vrf]['peers'][n] = dict()
-    #
-    #             if n_data['peerState'] == 'Established':
-    #                 bgp_neighbors[vrf]['peers'][n]['is_up'] = True
-    #                 bgp_neighbors[vrf]['peers'][n]['is_enabled'] = True
-    #             else:
-    #                 bgp_neighbors[vrf]['peers'][n]['is_up'] = False
-    #
-    #                 reason = n_data.pop('peerStateIdleReason', None)
-    #                 if reason == 'Admin':
-    #                     bgp_neighbors[vrf]['peers'][n]['is_enabled'] = False
-    #                 else:
-    #                     bgp_neighbors[vrf]['peers'][n]['is_enabled'] = True
-    #
-    #             bgp_neighbors[vrf]['peers'][n]['remote_as'] = n_data['asn']
-    #             bgp_neighbors[vrf]['peers'][n]['uptime'] = n_data['upDownTime']
-    #
-    #             raw_data = eval(
-    #                 'self.device.show_ip_bgp_neighbors_vrf_{}(format="text", pipe="section {}")'.format(vrf, n)
-    #             )['output']
-    #
-    #             n_data_full =  colon_separated_string_to_dict(raw_data)
-    #             sent, rcvd = n_data_full['IPv4 Unicast'].split()
-    #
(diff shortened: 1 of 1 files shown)


napalm/eos.py
```diff
@@ -237,48 +237,94 @@ def get_interfaces_counters(self):
...
+                elif re.search(remote_id_regexp, line):
+                    bgp_neighbors[peer]['remote_id'] = re.search(remote_id_regexp, line).group(1)
+                    bgp_neighbors[peer]['vrf'] = re.search(vrf_regexp, line).group(1)
+                    continue
+                elif re.search(IPv4_sent_regexp, line):
+                    bgp_neighbors[peer]['ipv4'] = re.search(IPv4_sent_regexp, line).group(2)
+                    continue
+                elif re.search(IPv6_sent_regexp, line):
+                    bgp_neighbors[peer]['ipv6'] = re.search(IPv6_sent_regexp, line).group(2)
+                    continue
```
bigzaqui: throw exception if nothing matches

napalm/eos.py
```diff
@@ -237,48 +237,94 @@ def get_interfaces_counters(self):
...
+
```
bigzaqui: the loop below can be removed using next(peer for peer in bgp_neighbors if peer == needed_peer)

napalm/eos.py
```diff
@@ -237,48 +237,94 @@ def get_interfaces_counters(self):
...
+                bgp_counters[vrf]['router_id'] = output[id]['vrfs'][vrf]['routerId']
+                bgp_counters[vrf]['peers'] = dict()
+                for peer in output[id]['vrfs'][vrf]['peers']:
+                    bgp_counters[vrf]['peers'][peer] = dict()
+                    bgp_counters[vrf]['peers'][peer]['local_as'] = int(output[id]['vrfs'][vrf]['asn'])
+                    bgp_counters[vrf]['peers'][peer]['remote_as'] = int(output[id]['vrfs'][vrf]['peers'][peer]['asn'])
+                    peerState = output[id]['vrfs'][vrf]['peers'][peer]['peerState']
+                    bgp_counters[vrf]['peers'][peer]['is_up'] = peerState == 'Established'
+                    try:
+                        output[id]['vrfs'][vrf]['peers'][peer]['peerStateIdleReason']
```
bigzaqui: ? line above is doing nothing

napalm/eos.py
```diff
@@ -237,48 +237,94 @@ def get_interfaces_counters(self):
...
+                        bgp_counters[vrf]['peers'][peer]['is_enabled'] = False
+                    except:
+                        bgp_counters[vrf]['peers'][peer]['is_enabled'] = peerState == 'Established' or peerState == 'Active'
```
bigzaqui: fancy

napalm/eos.py
```diff
@@ -237,48 +237,94 @@ def get_interfaces_counters(self):
...
+                if peer == needed_peer:
+                    if bgp_neighbors[peer]['vrf'] == vrf:
+                        bgp_peer['remote_id'] = bgp_neighbors[peer]['remote_id']
+                        bgp_peer['description'] = bgp_neighbors[peer]['description']
+                        bgp_peer['ipv4'] = bgp_neighbors[peer]['ipv4']
+                        bgp_peer['ipv6'] = bgp_neighbors[peer]['ipv6']
+            return bgp_peer
+
+        bgp_counters = dict()
+        for id in [0,2]:
```
bigzaqui: a bit too hardcoded, will do an assert of output to check that indeed has only 2 keys, otherwise throw exception

napalm/eos.py
```diff
@@ -237,48 +237,94 @@ def get_interfaces_counters(self):
...
+            neighbor_regexp = re.compile('BGP neighbor is (.*?),')
+            description_regexp = re.compile('Description: (.*?)$')
+            remote_id_regexp = re.compile('remote router ID (.*?),')
+            vrf_regexp = re.compile('VRF (.*?)$')
+            IPv4_sent_regexp = re.compile('IPv4 Unicast: ( *)(\d*) ')
+            IPv6_sent_regexp = re.compile('IPv6 Unicast: ( *)(\d*) ')
+
+            for line in output_to_parse.splitlines():
+                if re.search(neighbor_regexp, line):
+                    peer = re.search(neighbor_regexp, line).group(1)
```
bigzaqui: .group() returns None when finds nothing, you should handle that case

napalm/eos.py
marcovespignani: Due to the output from Arista, there will be unmatched lines. If I throw an exception, it will stop the execution

napalm/eos.py
marcovespignani: As discussed, the regex is executed twice, so the if statement checks if it finds something or not

bigzaqui:
+1

PR #132, opened 2016-01-05
Change type for variables to bool

ppieprzycki:
This commit will change variables commit_changes and replace_config to bool.
Also according to the doc string replace_config variable is not required but has default set to false.

After this change module will accept following values for mentioned variables:
yes,on,1,true,1,no,off,0,false,0

After this change module will not be backward compatible if someone used values "True" or "False" as in following example

msg: value of replace_config must be one of: yes,on,1,true,1,no,off,0,false,0, got: False

### DIFF
--- a/ansible/napalm_install_config
+++ b/ansible/napalm_install_config
@@ -64,19 +64,19 @@ options:
     replace_config:
         description: If set to True the entire configuration on the device will be replaced during the commit. If
                      set to False, we will merge the new config with the existing one. Default: false.
-        choices: [yes,on,1,true,1,no,off,0,false,0]
+        choices: [yes,on,1,true,no,off,0,false]
         required: False
     diff_file:
         description: A file where to store the "diff" between the running configuration and the new configuration. If
                      it's not set the diff between configurations is not saved.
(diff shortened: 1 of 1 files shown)


dbarrosop:
Hi!
The docstring mentions True and False for those parameters. Could you update the docstring to reflect what you mention in your PR message?

Thanks!

ppieprzycki:
Just updated doctoring. 
Added all possible combinations as choices.

dbarrosop:
Thanks! Will do some tests soon and approve the PR if I don't see anything wrong (I a sure I wont but better safe than sorry). I am still enjoying some days off so I guess I will take a look early next week : )

dbarrosop:
Hi! As you probably read here:

https://groups.google.com/forum/#!topic/napalm-automation/DI2w4s1vk9I

We have changed the organization and moved the ansible plugins to a dedicated repo. Would you mind closing this PR and sending a new one to the new repo? The reason why I prefer to do this is so your contribution is not lost during the movement.

ppieprzycki:
New pull request in separate repository 
https://github.com/napalm-automation/napalm-ansible/pull/2

PR #134, opened 2016-01-06
Cisco IOS support using SCP file transfers

ktbyers:
This includes unit testing. I will provide some additional details in a comment on the unit testing.


ktbyers:
I created/updated the unit tests (see TestIOSDriver.py).

I am testing the following against a Cisco 800 series router.

Core file operations:
load_replace_candidate
load_merge_candidate
compare_config
commit_config
discard_config
rollback

I also wrote tests for pretty much all of the internal/misc methods that I am using (in order to support the file operations.

Internal methods:
_enable_confirm         
_disable_confirm        
_gen_rollback_cfg       Tested as part of rollback
_check_file_exists      

Misc methods:
open                        
close                                       Skipped
normalize_compare_config    Tested indirectly
scp_file                    
gen_full_path               


One main thing that is still missing is the 'config' option to both the load_replace_candidate() and load_merge_candidate() methods. Currently, these are both using only the 'filename' option.

ktbyers:
Also, I didn't do anything beyond minor cleanup for the 'get_' methods (including no current unit tests).

dbarrosop:
Hello,
good work. I have a few concerns:

1. You are changing the behavior of the driver compared to the rest. Your __init__ method supports extra parameters that if you try to use in any other driver the code will break as that extra parameter is not supported. I am open to discuss if we want to support extra options for different drivers as there might be some use cases (vdoms for fortios, ssh_keys for netconf based devices, etc) but we need consensus and a proper way of doing this before merging this PR.
2. You are adding new tests. Could you prepend ios_only_test to the test names that apply only to IOSDriver? Do you think those tests should be brought to base class so we run them for all the vendors?
3. Could you or @GGabriele help with the unittests for the getters? At least providing the mocking data.
4. You mention that "'configure replace' will accept the config with a typo command, but will just reject the relevant command". IMHO I think it would be better to detect that there was some issue and rollback automatically. That's what FortiOS does and it's one of the premises of napalm, that all intended changes are applied or none. What do you think?

I am also nudging @fooelisa as she proposed long ago adding the capability of allowing extra options for some drivers.

Thanks again for the good work : )

David

ktbyers:
Item #1 -- extra \_\_init\_\_ arguments

Yes, this a bit of a challenge since the behavior of API devices vs SSH devices has some different requirements.

So can I get rid of the extra arguments...I could get rid of these (and just use hard-coded values). This would be fine and not important.
                 candidate_cfg='candidate_config.txt', 
                 rollback_cfg='rollback_config.txt',
                 merge_cfg='merge_config.txt'

These two are a bit more of a problem:
                 dest_file_system='flash:'
                 global_delay_factor=.2

For dest_file_system, this will be an issue (see the comment on this issue https://github.com/spotify/napalm/issues/133).

I could probably have the global_delay_factor be based on the timeout since I am not using the timeout, but this isn't a great practice since then 'timeout' would then have two different meanings. The problem that I will run into without this variable is some people will have really slow devices and won't have a way to work around it (i.e. their code will break and no direct way to fix in NAPALM).

If we want to support extra arguments, then the key question becomes how do you make the Ansible modules support the extra arguments in a transparent way (and a way that is not overly complex). I somewhat view as less important for code that is using NAPALM outside of Ansible. I say this because I there is a pretty easy workaround for code outside of Ansible.

Another parameter that I need is a port for non-standard SSH/API ports, but this might be a general argument that you want to add to the Ansible modules anyway.

Here is one possible solution: in the Ansible module, we could add a single argument 'optional_args'. Optional arguments would be passed as a string to the underlying NAPALM driver. In the underlying NAPALM driver, it would default to the null string. It would then be the responsibility of each underlying driver to process optional_args (probably would want to standardize the way the string should be formatted).

Obviously, this would involve some work on existing drivers. Although it would be pretty quick work to make it work transparently on all of the drivers (i.e. add the optional_args driver to each module and have it do nothing by default).

Thoughts?

ktbyers:
Item #2 -- tests.

Yes, not a problem on appending 'ios_only_test' to IOS specific tests. Most of these tests were specific to the ios code so not applicable to other vendors. 

Item #3 -- Yes, I can probably work on this.

Item #4 -- I think this might be hard (given the way 'configure replace' works). Let me do some additional testing and provide some details.

ktbyers:
I am going to open item 4 as a separate issue (issue #135)

dbarrosop:
Regarding 1, I was actually thinking on having optional_args as a dictionary for the __init__ method and then have keys with names like $driver_$usage for vendor specific stuff and just $usage for things we could reuse in all the drivers. So the __init__ method in the base class would be:

    def __init__(self, hostname, username, password, timeout, optional_args={}):

And your implementation could easily be:

    def __init__(self, hostname, username, password, timeout, optional_args={}):
        # We use the get method of the dictionary with a default value so we can omit them
        self.port = optional_args.get('port', 22) # Optional arg that we can actually reuse everywhere
        self.candidate_cfg = optional_args.get('ios_candidate_cfg', 'candidate_config.txt') # IOS only
        ...

This should work for both ansible and not ansible code. Thoughts?

ktbyers:
@dbarrosop Yes, that is a better solution (having optional_args as a dict)

Let's do a small proof of concept on it...

One change you should have it be like this for the default

<pre>
def __init__(self, hostname, username, password, timeout, optional_args=None):

    if optional_args is None:
        optional_args = {}
</pre>

Otherwise you will run into an issue with mutable default arguments in Python.

dbarrosop:
> Otherwise you will run into an issue with mutable default arguments in Python.

I was expecting nobody would add anything to that parameter during runtime (just reading values) but you are right, you never know what crazy things we might be doing in the future so better safe than sorry.

Thanks for that catch : )

ktbyers:
@dbarrosop Okay, this looks like it would work.

The only pain I ran into was in the playbook. Ansible kept giving me grief on the dictionary variable, eventually I just made it a separate var and that worked:

<pre>
  vars:
    optional_args: {'dest_file_system': 'flash:', 'global_delay_factor': .5, 'candidate_cfg': 'my_test_cfg.txt'}
</pre>

and then reference it:

<pre>
optional_args="{{ optional_args }}"
</pre>

ktbyers:
@dbarrosop 

Let me know if you are okay with me going ahead and doing this and I will update the code to take these optional_args.

ktbyers:
@dbarrosop Okay, I went ahead and made the optional_args changes and pushed it to the head of the cisco_ios_scp branch. It was mostly done anyways.

See https://github.com/ktbyers/napalm/commit/9f346e51b08a72a0fa0c40137acaf79b628f97f2

I have only tested it on Cisco IOS.

PR #143, opened 2016-01-15
Dont poll other that physical interfaces

tehhobbit:


dbarrosop:
As discussed, please, filter out subinterfaces only, don't filter bundles or virtual interfaces.

PR #154, opened 2016-02-02
Extended NAPALM with PluribusDriver & few more methods

mirceaulinic:
In this PR we propose to extend the capabilities of NAPALM with one more driver for Pluribus devices. Although the library pyPluribus is not yet mature, we are still working on it and will add new features in the near future.
Also we have implemented the following methods for JunOS, IOS-XR, NXOS, EOS and Pluribus:
- get_arp_table
- get_mac_table
- get_ntp_peers
- get_lldp_neighbors_detailed (which is a detailed version of the existing get_lldp_neighbors -- probably it would be a good idea to merge them in a future release).

Also, on some devices we have designed few other methods such as show_route, get_interfaces_ip or get_bgp_neighbors_detail.

### DIFF
--- a/napalm/__init__.py
+++ b/napalm/__init__.py
@@ -19,6 +19,7 @@ from fortios import FortiOSDriver
 from nxos import NXOSDriver
 from ibm import IBMDriver
 from ios import IOSDriver
+from pluribus import PluribusDriver
 
 def get_network_driver(vendor):
     driver_mapping = {
@@ -32,6 +33,7 @@ def get_network_driver(vendor):
         'NXOS': NXOSDriver,
         'IBM': IBMDriver,
         'IOS' : IOSDriver,
+        'PLURIBUS': PluribusDriver
     }
     try:
         return driver_mapping[vendor.upper()]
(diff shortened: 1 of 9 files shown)


dbarrosop:
Hi!
This is awesome! A couple of things:

 1. Could you document the new methods in base.py? It's important to make sure people understands what the method is supposed to return.
 1. Does pluribus have a VM or something so we can run unittests? Otherwise it would be interesting if at least we could have some mock data to test the results.
 1. Could you provide mock data for the methods you implemented?

Feel free to ask me any doubts you might have.

Regards.
David

dbarrosop:
Could you also break down the PR into smaller PRs? I would suggest one per new method and another one with the new driver. The reasoning is that it's hard to actually peer review a huge PR and some stuff might be merged straight away while other might prompt some discussion and maybe require some changes.

mirceaulinic:
Hi David,

I will submit soon a couple of PRs as you proposed together with mock inputs. I have not forgot abouth the documentation and when I will make the requests I will include it as well.

Thanks!
Mircea

PR #155, opened 2016-02-03
Method get_mac_address_table() defined for JunOS, NXOS, IOS-XR & EOS

mirceaulinic:
Child of PR #154

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,6 +58,5 @@ docs/_build/
 env
 *.swp
 
-napalm/fix_script.sh
 test/unit/test_devices.py
 
--- a/Manifest.in
+++ b/Manifest.in
@@ -1,2 +1,0 @@
-include requirements.txt
-include napalm/utils/*.yml


dbarrosop:
Hello,
a few comments first:

1. I like the returning structure, simple and effective.
2. Could you add a brief description on base.py of the attributes? I know they are super obvious but sometimes it isn't. For example, due to active being hardcoded on IOSXR to True I am a bit confused of the meaning.
3. There seems to be some conflicts, could you fix them?

Regarding the method, I am not a fan of methods with different keywords that alter the behavior of the method. It usually yields complex code and it's a mess to actually test. For this particular case I have an alternative:

1. We can skip the filter entirely. Computationally it's going to make no difference if we get all entries and given how the returning structure data is it's super easy to build a comprension list to filter stuff out (for people with hundreds of thousands of mac address entries in the same device this might be a problem but I don't think there are that many out there, at that scale L3 is the norm, I hope...):
        ```
        mac_address_table = device.get_mac_address_table()
        my_filtered_table_by_interface = [m for m in mac_address_table if m.interface == 'vlan100' ]
        my_filtered_table_only_dyn = [m for m in mac_address_table if not m.static ]
        ```
2. We build several methods instead of one, a method per filter.

What do you think? As mentioned, it's just to avoid having methods which change the behavior depending on the argument and that requires to keep switching and checking variables to avoid surprises. It's not a big deal so if you think this is going to work better for you I am happy to leave it as you proposed.

Regards.
David

jejenone:
@dbarrosop I agree with keeping things simple - However looking forward, we'll probably need something like this for looking at the routing tables. Also keep in mind that IOS-XR tends to return extremely long XML structures.

jejenone:
@dbarrosop

mirceaulinic:
Thank you very much for your feedback, @dbarrosop!

We will update the documentation to make sure everything will be clear, no problem about that.

Regarding the filters they might be really handy sometimes. I agree with you that it is very easy to rebuild a filtered output based on the whole output using comprehensions. Still there are cases when retrieving a bigger output can cause problems: for example on JunOS when retrieving the whole MAC Address Table  from devices with many entries will cause RpcTimeoutError.

Here we will remove the filters and for ARP getter as well but as @jejenone said we are constrained to keep filters for routes (and perhaps other methods in the future).

jejenone:
Rebased, cleaned, and fixed conflicts.

mirceaulinic:
@dbarrosop Everything should be ready by now.

mirceaulinic:
@ande0581 could you please have a look in this thread? You might need to make few slight changes in #178 and #179. Sorry about that!

ande0581:
I have updated both #178 and #179 to remove the filters.

PR #156, opened 2016-02-03
Method CLI defined for JunOS, NXOS, IOS-XR & EOS

mirceaulinic:
Child of #154

### DIFF
--- a/napalm/base.py
+++ b/napalm/base.py
@@ -405,3 +405,10 @@ class NetworkDriver:
 
         """
         raise NotImplementedError
+
+    def cli(self, command = None):
+
+        """
+        Will execute a valid command on the CLI and returns the output.
+        """
+        raise NotImplementedError
--- a/napalm/base.py
+++ b/napalm/base.py
@@ -406,9 +406,28 @@ class NetworkDriver:
         """
         raise NotImplementedError
 
-    def cli(self, command = None):
+    def cli(self, *commands):
 
         """
-        Will execute a valid command on the CLI and returns the output.
+        Will execute a list of commands and return the output in a dictionary format.
+        Example:
+            {
+                u'show version and haiku':  u'''Hostname: re0.edge01.arn01
+                                                Model: mx480
+                                                Junos: 13.3R6.5
+
+                                                        Help me, Obi-Wan
+                                                        I just saw Episode Two
+                                                        You're my only hope
+                                            ''',
+                u'show chassis fan'     :   u'''Item                      Status   RPM     Measurement
+                                                Top Rear Fan              OK       3840    Spinning at intermediate-speed
+                                                Bottom Rear Fan           OK       3840    Spinning at intermediate-speed
+                                                Top Middle Fan            OK       3900    Spinning at intermediate-speed
+                                                Bottom Middle Fan         OK       3840    Spinning at intermediate-speed
+                                                Top Front Fan             OK       3810    Spinning at intermediate-speed
+                                                Bottom Front Fan          OK       3840    Spinning at intermediate-speed
+                                            '''
+            }
         """
         raise NotImplementedError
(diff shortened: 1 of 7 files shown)


jejenone:
+1 on this - this is super useful when using a framework like Salt.

napalm/base.py
```diff
@@ -364,3 +364,9 @@ def get_interfaces_counters(self):
             }
         """
         raise NotImplementedError
+
+        def cli(self, command = ''):
```
dbarrosop: Could we set that to None and specify in the docstring that it has to be a list of commands?

mirceaulinic:
@dbarrosop I suppose this PR should be merged before any others. Do you think it looks good now?

napalm/eos.py
```diff
@@ -484,3 +484,18 @@ def get_lldp_neighbors_detail(self, interface = ''):
                 )
 
         return lldp_neighbors_out
+
+    def cli(self, command = None):
+
+        if type(command) is not str:
```
dbarrosop: Why a string? Wouldn't be better to allow a list of commands instead of just a single command?

napalm/eos.py
```diff
@@ -484,3 +484,18 @@ def get_lldp_neighbors_detail(self, interface = ''):
...
+            return 'Please enter a valid command!'
+
+        commands = list()
+        commands.append(command)
+
+        try:
+            return self.device.run_commands(commands, encoding = 'text')[0].get('output')
+        except pyeapi.eapilib.CommandError:
+            return 'Unrecognized command: "{0}"'.format(command)
```
dbarrosop: Could we raise an exception instead? I think it's better if we raise an Exception so you can handle them (or ignore them) programatically. By looking at your code I guess we will need at least two:

1. CommandError
2. CommandTimeout

The messages are fine, it's just that I'd rather know that an error occurred without having to parse the return string.

mirceaulinic:
@dbarrosop Is what you thought about?

napalm/iosxr.py
```diff
@@ -630,3 +631,25 @@ def get_lldp_neighbors_detail(self, interface = ''):
                 continue # jump to next neighbor
 
         return lldp_neighbors
+
+    def cli(self, commands = None):
+
+        cli_output = dict()
+
+        if type(commands) is not list:
+            return 'Please enter a valid list of commands!'
```
dbarrosop: Can't we just throw a TypeError exception?

dbarrosop:
Hi, not exactly but almost. What I want is to through an exception. For example:

            try:
                 cli_output[unicode(command)] = self.device.run_commands([command], encoding = 'text')[0].get('output')
                 # not quite fair to not exploit rum_commands
                 # but at least can have better control to point to wrong command in case of failure
             except pyeapi.eapilib.CommandError:
                 # for sure this command failed
                 cli_output[unicode(command)] = 'Invalid command: "{cmd}"'.format(
                     cmd = command
                 )
                 raise napalm.exceptions.CommandError(cli_output)
             except Exception as e:
                 # something bad happened
                 cli_output[unicode(command)] = 'Unable to execute command "{cmd}": {err}'.format(
                     cmd = command,
                     err = e
                 )

We will have to create the exception as that one doesn't exist yet. The same goes for the timeout error. We could create a new exception called CommandTimeout or something like that.

mirceaulinic:
That's easy, but firstly I would like to understand why would we want to do something like that instead of specifying exactly what command caused the error.
Let's take the case of an user that executes something like ```device.cli(["show version", "show something bad"])``` while ```show something bad``` would throw an Exception. Do you think we should raise CommandTimeout even if "show version" run properly? If so, I will refactor the code -- doesn't cost very much :)

dbarrosop:
I think it's worth throwing the exception and returning the whole dictionary, that way you know immediately that some command might have gone wrong and you have to figure out which one. Otherwise you will have to go through the output and try to find if there was any error. It's just an easy way of controlling the flow of your code.

mirceaulinic:
Agreed, will change it like that.

mirceaulinic:
Is it anything better now?

dbarrosop:
Looks great now : )

PR #157, opened 2016-02-03
Method get_arp_table() defined for JunOS, NXOS, EOS & IOS-XR

mirceaulinic:
Child of #154


dbarrosop:
This one has conflicts as well.

jejenone:
rebased and squashed

jejenone:
Actually I don't agree with the return structure.
Interface can't be the key. An ARP table is a list of IP/MAC association - having the interface is a bonus.
This should just return a list of key/values

mirceaulinic:
Changed output, no more conflicts, removed filters.

dbarrosop:
Regarding the output, I don't care that much. I would like it to look like the ``get_mac_address`` method as they provide similar functionality. So as long as we have a similar output in both of them I am fine.

PR #159, opened 2016-02-03
Defined get_lldp_neighbors_detail() method for JunOS, NXOS, EOS & IOS-XR

mirceaulinic:
Child #154 
Although get_lldp_neighbors() does exist, there is a need to retrieve further details of the LLDP neighbors such as capabilities, system description etc.
We preferred to define a different method in order to not overlap with existing work. In the future they might be merged into a single method.


dbarrosop:
More conflicts...

Regarding having two methods. I think it's better to have multiple methods. Otherwise the code becomes entangled and testing becomes harder.

napalm/base.py
```diff
@@ -364,3 +364,41 @@ def get_interfaces_counters(self):
...
         """
         raise NotImplementedError
+
+    def get_lldp_neighbors_detail(self, interface = ''):
+        """
+        Returns a detailed view of the LLDP neighbors as a dictionary
+        containing lists of dictionaries for each interface.
+
+        Inner dictionaries contain fields:
+            * parent_interface (string)
```
dbarrosop: What is this attribute?

jejenone:
rebased, squashed and cleaned

napalm/base.py
mirceaulinic: parent_interface? The name of the aggregated interface

mirceaulinic:
I think this one is ready too

napalm/base.py
jejenone: @dbarrosop 
```
jerome@re0.edge01.sjc01> show lldp neighbors
Local Interface    Parent Interface    Chassis Id          Port info          System Name
xe-0/1/1           ae7                 00:1c:73:ee:c0:46   Ethernet3/29/1     core01.sjc01
xe-0/3/3           ae7                 00:1c:73:ee:c0:46   Ethernet3/29/2     core01.sjc01
```

PR #161, opened 2016-02-03
Defined get_route_to for JunOS, IOS-XR & EOS

mirceaulinic:
Child of #154


napalm/base.py
```diff
@@ -364,3 +364,44 @@ def get_interfaces_counters(self):
             }
         """
         raise NotImplementedError
+
+        def show_route(self, destination = ''):
```
dbarrosop: Could you rename it to: `get_ipv4_route_to`. One to make it more human readable and two to make sure people understands they are going to get ipv4 only. We might get a `get_ipv6_route_to` some day : )

napalm/base.py
```diff
@@ -364,3 +364,44 @@ def get_interfaces_counters(self):
...
+                    "1.0.0.0/24": [
+                        {
+                            "active_route"      : True,
+                            "age"               : 1368451,
+                            "as_path"           : "15169 I",
+                            "local_preference"  : 200,
+                            "next_hop"          : "206.223.116.21",
+                            "outgoing_interface": "ae10.0",
+                            "preference"        : 170,
+                            "protocol"          : "BGP"
```
dbarrosop: What if it comes from OSPF? Do you leave the BGP attributes empty?  Fine by me, just clarifying.

mirceaulinic: Haven't checked, but normally should be OSPF specified instead of BGP.

napalm/base.py
jejenone: How about leaving that name, and return an exception when an IPv6 route is asked ?
I think ultimately we should be able to use the same function for v4 and v6.

dbarrosop: Ok! In that case we could create an exception called IPv6NotImplemented and use it for cases like this. I still would like to change the name though, but to `get_routes_to` if we are going to support both v4 and v6. The reason is that the rest of the methods are named get_something and I would like to keep it consistent.

napalm/base.py
```diff
@@ -364,3 +364,44 @@ def get_interfaces_counters(self):
...
+                * next_hop (str)
+                * outgoing_interface (str)
+                * preference (int)
+                * protocol (int)
+
+            Example:
+                {
+                    "1.0.0.0/24": [
+                        {
+                            "active_route"      : True,
```
dbarrosop: I guess you are doing this because you plan to return all possible routes, are you contemplating ECMP routes as well? Would I get multiple routes with this attribute set to True?

mirceaulinic:
Sure, this attribute should be True for all active routes.

napalm/base.py
jejenone: Agree on both.

mirceaulinic:
Some issues on IOS-XR. Unable to retrieve details of BGP routes...

dbarrosop:
Can you rebase this one? It would be great if you could also add the tests to travis.

Thanks!

mirceaulinic:
@dbarrosop unfortunately this cannot be refactored too soon. Perhaps next week or we can postpone it for release 0.53. What do you think?

dbarrosop:
Next week is fine :)

mirceaulinic:
@dbarrosop From my point of view this one is ready now.

PR #162, opened 2016-02-03
Defined get_bgp_neighbors_detail() for JunOS & IOS-XR

mirceaulinic:
Child of #154

PR #160, opened 2016-02-03
Defined get_interfaces_ip() method for JunOS, NXOS, EOS & IOS-XR

mirceaulinic:
Child of #154

PR #158, opened 2016-02-03
Method get_ntp_peers() defined of JunOS, NXOS, EOS & IOS-XR

mirceaulinic:
Child of #154

PR #173, opened 2016-02-08
add get_interfaces_ip function to IOS

ande0581:


dbarrosop:
Nice one, I will schedule this for the upcoming release.

mirceaulinic:
Hi @ande0581,

There's PR #160 with implementation of the same method get_interfaces_ip for JunOS, IOS-XR, EOS and NXOS. Have you checked that before? The purpose was to retrieve a simple list of IP addresses, while your method returns a much more elaborate output. 
But maybe you are right -- we will refactor our PR to provide a similar output as you proposed.
Meanwhile, could you please implement in IOS similar methods to #162, #155, #156, #157, #158 and #159? Thanks.

ande0581:
Hi @mirceaulinic,

When i first started building this method i didn't pay attention to #160. I did consult the napalm-dev mailing list and we discussed what information we wanted to see and i built what was discussed.

Regarding matching your other PRs, i have submitted pull requests for the following:

#155, #156, #157, #158, #159 

I have passed on #162. I can potentially come back around to this (if someone else doesn't get to it first) but right now i need to focus on some other priorities.

mirceaulinic:
No problem about that, @ande0581! We also need now to change to output of get_interfaces_ip but will be waiting in the deployment queue for a couple of days.

napalm/ios.py
```diff
@@ -520,6 +520,109 @@ def get_interfaces(self):
...
 
+
+    def get_interfaces_ip(self):
+        '''
+        Get interface ip details
+        Returns a dict of dicts
+
+        Example Output:
+
+        {   u'FastEthernet8': {   'ipv4': {   u'10.66.43.169': {   'prefix-length': 22}}},
```
mirceaulinic: Quick question: if there are two logical units underneath FastEthernet8 for example (say 1 and 2), the output will contain two keys named FastEthernet8.1 and FastEthernet8.2?

ande0581:
Yes, on my side the output would look like this.

{   u'Multilink1': {   'ipv4': {   u'192.168.0.2': {   'prefix-length': 30}}},
    u'Port-channel1.14': {   'ipv4': {   u'10.53.128.1': {   'prefix-length': 24}}},
    u'Port-channel1.81': {   'ipv4': {   u'10.53.153.220': {   'prefix-length': 24}}},
    u'Tunnel1': {   'ipv4': {   u'172.16.21.2': {   'prefix-length': 30}}}}

I didnt give this much thought, but this is how it would work today.

mirceaulinic:
That's fine, thanks for clarification!

napalm/ios.py
mirceaulinic: Missed that: prefix-length instead of prefix_length please. And generally I think we should avoid "-".

ande0581: I changed it to underscore

PR #174, opened 2016-02-08
change output from abbreviated interface names to full names for IOS

ande0581:
changed the output for the get_interfaces() method to return the full interface name instead of the abbreviated name to keeps this consistent.

### DIFF
--- a/napalm/ios.py
+++ b/napalm/ios.py
@@ -322,68 +322,40 @@ class IOSDriver(NetworkDriver):
         return False
 
     def get_lldp_neighbors_detail(self):
-        '''
-        Output command format:
-
-            show lldp neighbors detail
-
-            ------------------------------------------------
-            Local Intf: Gi0/1
-            Chassis id: G404-0844
-            Port id: PORT-3
-            Port Description - not advertised
-            System Name: UEC-WDBYMNUEC1-UNI-53055
-
-            System Description:
-            AMN-1000-GT, FW: AMT_7.0.0.1_2517, HO: Dry-contact Input
-
-            Time remaining: 103 seconds
-            System Capabilities - not advertised
-            Enabled Capabilities - not advertised
-            Management Addresses:
-                IP: 172.29.118.206
-            Auto Negotiation - not supported
-            Physical media capabilities - not advertised
-            Media Attachment Unit type - not advertised
-            Vlan ID: - not advertised
-
-        Return data structure is a dictionary, key is local_port:
-
-        {   u'FastEthernet2': [   {   'hostname': u'SEP204C9ED7CB80',
-                                      'port': u'204C9ED7CB80:P1'}],
-            u'FastEthernet3': [   {   'hostname': u'SEP10BD1801A0B4',
-                                      'port': u'10BD1801A0B4:P1'}],
-            u'FastEthernet7': [{   'hostname': u'Meraki Network', 'port': u'1'}]}
-
-
-        Caveats:
-
-        The original way this method was built used the following command:
-
-            customer-router-2901#sh lldp neighbors
-            Capability codes:
-                (R) Router, (B) Bridge, (T) Telephone, (C) DOCSIS Cable Device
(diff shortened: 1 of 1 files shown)


dbarrosop:
I just realized that get_lldp_neighbors, get_facts and others might be returning short names as well. I would suggest to either change them all or just leave them all as short... I don't know if you can confirm, I don't have an IOS device right now.

If that's the case, maybe we can just add a method that maps short with long names and returns the long format. Something like (quick and dirty example, not necessarily the best solution):

    def short_to_long_iface_name(name):
         # Better to do it with regexps to make sure we don't replace a name that is already long
         if name.startswith('Fa'):
             return name.replace('Fa', 'FastEthernet')
         elif name.startswith('Gi'):
             return name.replace('Gi', 'GigabitEthernet')
         else:
             #catch all
             return name

And then, in the code that is already in place:


      interface = short_to_long_iface_name(interface)

ande0581:
I just looked at the methods. For the ones i can tell that return interfaces this is the output.

get_facts():
	Interfaces are full names

get_interfaces():
	Interfaces are full names

get_interfaces_ip():
	Interfaces are full names

get_interfaces_counters():
	Interfaces are full names

get_lldp():
	Interfaces are abbreviated

I will look at fixing the get_lldp() method to return the full names.

ande0581:
I changed how get_lldp() method works. Changes outlined in the comments of the code. People may not agreed to how i approached this. Caveats to each method are outlined in the code. Feel free to critique.

ande0581:
I changed the get_lldp_neighbors_detail() to have the same output as the other methods for IOS-XR, EOS, etc.

I personally dont think there needs to be both a get_lldp_neighbors and get_lldp_neighbors_detail method. They both return the same information but the get_lldp_neighbors currently returns abbreviated names where the more detailed and consistent get_lldp_neighbors_detail returns full interface names, plus more details.

dbarrosop:
@ktbyers any thoughts on this?

ktbyers:
@dbarrosop Yes, I think we should use the full names and not the abbreviated names. 

Only concern with a helper function is to make sure we properly convert all the names.

I agree that we should only have the 'get_lldp_neighbors' and should return the canonical names. By canonical names I mean the full names as returned by 'show interfaces' or 'show ip int brief'.

ande0581:
Just to point it out, the latest version of what i proposed returns full names directly from the device, not via a helper function.

ktbyers:
@ande0581 @dbarrosop It looks like NAPALM is standardizing on having both get_lldp_neighbors() and get_lldp_neighbors_detail(). Correct me if I am wrong @dbarrosop.

Currently this code only implements get_lldp_neighbors_detail(). It also needs get_lldp_neighbors().

The current code will fail the unit test for get_lldp_neighbors() as that method doesn't exist.

ande0581:
@ktbyers @dbarrosop I added a modified get_lldp_neighbors() method to this pull request. Now both methods exist. One thing i want to point out is that certain IOS versions on the equipment that i have been testing with, the 'Local Intf:' value isnt in the output from the router. This caused the key to be changed to unknown. In the original method that used the abbreviated names, the interface was there but in short form 'Fa4'.

dbarrosop:
Great job @ande0581! Three things:

1. Can you rebase and add the mocked data and enable the unittests? Feel free to let us know if you need help with that.
2. Regarding the 'local intf:'. How difficult would it be to fix? If it's difficult we can always report it as an issue and tackle it later so we don't block this PR.
3. Can you update the docs to reflect that the methods are supported?

Thanks!

dbarrosop:
I fixed a few typos and added the tests. If you want to check what I did check the following commits: 71b4ddbea7c7d1c50207ae8a9b86eb75d9d3ada2, f81546c4418c391b90fc2fdb9614891b5d4fc76a and 75d663fa082c9eb080777e958a65b31f75d5b41a.

ande0581:
Thanks. I will attempt to apply these same changes to mac_address_table,
ntp and snmp.

Jeff

PR #176, opened 2016-02-09
Method get_bgp_config() defined for JunOS, IOS-XR and EOS

mirceaulinic:
Returns BGP configuration


mirceaulinic:
@ande0581 Something for IOS perhaps? :+1:

dbarrosop:
I have been doing some tests and a part that there are conflicts again (seems like breaking down the PRs caused some havoc) there are some failed tests:

test_get_bgp_config (TestEOSDriver.TestGetterEOSDriver) ... FAIL
test_get_bgp_config (TestJunOSDriver.TestGetterJunOSDriver) ... ERROR

Could you check?

mirceaulinic:
In the FakeJunOSDevice class we need to add one more method for the configuration getters since in PyE they are treated as a different type of requests. The method receives two params: get_cmd (an etree type object with a structure like <configuration><protocols><bgp><group/></bgp></protocols></configuration>) and options (a dictionary with options).
Since there's one single RPC for all possible configuration-related requests I thought it would be less messy to have one single text file with the config (where to append parts of the config -- and possibly at some point it might look like a whole device configuration). In the same time I am thinking that we don't deal with one single type of device but with a variety with different features (switches that do not know BGP etc.) and it would be more handy to have files named based on the format of the get_cmd tree (e.g.: configuration__protocols__bgp__group___bgp___protocols___configuration.txt). What do you think?

dbarrosop:
> it would be more handy to have files named based on the format of the get_cmd tree (e.g.: configuration__protocols__bgp__group___bgp___protocols___configuration.txt)

I completely agree with this.

test/unit/eos/mock_data/show_running_config_begin_router_bgp.txt
```diff
@@ -0,0 +1,24 @@
...
+   neighbor IPv6-PEERS-GROUP-NAME route-map reject-all in
+   neighbor IPv6-PEERS-GROUP-NAME route-map reject-all out
+   neighbor IPv6-PEERS-GROUP-NAME maximum-routes 100
+   neighbor 192.168.0.1 peer-group IPv4-PEERS-GROUP-NAME
+   neighbor 192.168.0.1 remote-as 32934
+   neighbor 192.168.0.1 maximum-routes 100
+   neighbor 172.17.17.1 peer-group IPv4-PEERS-GROUP-NAME
+   neighbor 172.17.17.1 remote-as 13414
+   neighbor 172.17.17.1 maximum-routes 500
+   neighbor 2001:db8:0:1 peer-group IPv6-PEERS-GROUP-NAME
```
dbarrosop: You are missing a colon between db8 and 0. I guess the correct ip is 2001:db8::0:1.

mirceaulinic: Nice catch :+1: Most of the mock data is taken from production and some information replaced for non-disclosure thus didn't pay too much attention. Will correct that. Thanks!

mirceaulinic:
> I completely agree with this.

Will change it a bit later today.

napalm/eos.py
```diff
@@ -512,3 +513,210 @@ def cli(self, commands = None):
                 raise CommandErrorException(str(cli_output))
 
         return cli_output
+
+    def get_bgp_config(self, group = '', neighbor = ''):
```
dbarrosop: Have you considered checking TextFSM? Haven't used it but looks promising.

https://github.com/google/textfsm
http://jedelman.com/home/programmatic-access-to-cli-devices-with-textfsm/

I am not implying you should refactor the method, I just wanted to mention it : )

dbarrosop:
I realized that the module 'ipaddress' is missing in the requirements file. By the way, any particular reason to choose ipaddress instead of netaddr, which is part of the stdlib?

napalm/eos.py
```diff
@@ -14,6 +14,7 @@
 
 import pyeapi
 import re
+import ipaddress
```
dbarrosop: maybe we can use the netaddr in the stdlib to avoid having an extra dependency?

dbarrosop:
This one looks good now so I am merging : )

PR #177, opened 2016-02-10
add get_ntp_peers() to IOS

ande0581:


ande0581:
replacing with this PR https://github.com/napalm-automation/napalm/pull/217

PR #180, opened 2016-02-10
add cli() to IOS

ande0581:


napalm/ios.py
```diff
@@ -822,3 +822,15 @@ def get_environment(self):
         environment.setdefault('temperature', {})
         environment['temperature']['invalid'] = {'is_alert': False, 'is_critical': False, 'temperature': -1.0,}
         return environment
+
+    def cli(self, command=''):
```
dbarrosop: Could you revisit this PR and fix it so it complies with #156? Some late changes we did were:

1. command was renamed to commands and the default argument is None
1. If commands is not a list it should raise a TypeError exception
1. Because commands is not a list of commands all of them should be executed.

ande0581:
@dbarrosop 

A couple of questions. On the IOS you can execute anything and it will return 'Invalid input detected'.

The list would look like this:

{   'show clock': [u'*18:24:43.122 UTC Thu Feb 18 2016'],
    'show mistake': [   u"                     ^\n% Invalid input detected at '^' marker.\n"]}

Do you want to see 'show mistake' handled differently or just return it in the dict?

I do not understand what you are saying here.

> Because commands is not a list of commands all of them should be executed.

dbarrosop:
If there is a mistake it would be nice to throw an exception like here: https://github.com/napalm-automation/napalm/pull/156/files#diff-606e22a7673c7bf24f18d423054df653R496

> Because commands is not a list of commands all of them should be executed.

Don't mind me there, I was suffering from a severe lack of caffeine. You already did what I was trying to ask for there : )

Thanks!

dbarrosop:
Could you also update the matrix in the docs to indicate that the method is supported on IOS as well?

ande0581:
@dbarrosop 

Let me know if this is what you were looking for.

dbarrosop:
Yes, thanks!

PR #186, opened 2016-02-16
idiomatic switch-like statement in __init__ module

rightlag:
added a `switch`-like statement in the `napalm` `__init__.py` module for determining which class to instantiate based on the `vendor` argument.

Also cleaned up spacing issues to adhere to [pep8](https://www.python.org/dev/peps/pep-0008/) coding standards.


dbarrosop:
I agree with @ktbyers. I usually like doing things in a _pythonic way(TM)_ but given that there is no gain here and there is s slight detriment in readability I think I will close this PR.

Thanks for the PR @rightlag. I hope you understand why we prefer not to merge it.

PR #189, opened 2016-02-20
removing unnecessary print and typo

fooelisa:


fooelisa:
this is for junos def get_bgp_config() and get_environment()

PR #194, opened 2016-02-20
Make config lock optional on iosxr driver

fooelisa:


fooelisa:
Hm, good question, on iosxr you are able to commit without prior locking - do we want to enforce that?

mirceaulinic:
As far as I am concerned it is not mandatory to lock the config to be able to commit but in case you the config is not locked by default during connection time, it might be a good idea to avoid any other concomitant config change and to block the database hence lock before any load and unlock after. Does it make any sense? :)

mirceaulinic:
Also it would be nice to unlock the device before closing connection.

dbarrosop:
Could you also update the documentation to reflect the new option? Check #193

fooelisa:
Since @dbarrosop approved the Junos lock/unlock change, I guess we should stay consistent wrt locking / unlocking during the commit operation. Let me add that the same way @mirceaulinic did

fooelisa:
ready for merge

fooelisa:
Oh gosh, my rebase here added a lot of history.... how do I squash this properly? @mirceaulinic I'm sure you know? :)

mirceaulinic:
Since there are not conflicts, usually I just rebase -i HEAD~NUMBER_OF_COMMITS_BACKWARDS (e.g. git rebase -i HEAD~5). Keep only the first pick and replace the following with "squash" (or "s") save, exit and afterwards type only one single commit message. Sorry I am in a bit of hurry right now. Hope this helps; if not, tonight I will check it out when I have to solve other issues in #160.

fooelisa:
Sure, no rush :)

I am wondering if it is normal for all the previous commits by others to show up in this PR after my rebase?

jejenone:
Not normal - how did you rebase ?
You might just want to merge master into the branch and fix the conflicts in that case.

fooelisa:
No idea! I closed this and opened another PR after some local cleanup ;)

PR #196, opened 2016-02-21
Test Pluribus too

mirceaulinic:


test/unit/TestPluribusDriver.py
```diff
@@ -32,7 +32,7 @@ def setUpClass(cls):
         cls.device.open()
 
 
-class TestGetterPluribusriver(unittest.TestCase, TestGettersNetworkDriver):
```
mirceaulinic: It's a driver not a river tho...

dbarrosop: :)

dbarrosop:
Nice! Looks like some test is failing, could you check?

fooelisa:
+1 :)

PR #200, opened 2016-02-22
Pylama

dbarrosop:
I thought we could use pylama to enforce the style and do some sanity checks (no unused imports, no super complicated code, etc.).

Thoughs?

In the future we could even do some cleaning and integrate pylama with Travis so PRs report code that don't comply with the style.


jejenone:
Anything that helps having the code clean and consistent is a nice to have.
+1

dbarrosop:
I just realized this is PR 200!

ktbyers:
@dbarrosop This sounds reasonable. 

I have used flake8 which should be close to what you have configured (and separately pylint/pep8). The 120 char line length is reasonable.

PR #206, opened 2016-02-24
JunOS 12.x and under vs. JunOS 13.x LLDP XML RPC

mirceaulinic:
```lldp-local-interface``` (specific to JunOS < 13) vs ```lldp-local-port-id``` (JunOS >= 13) tags in LLDP XML RPC reply

### DIFF
--- a/napalm/utils/junos_views.yml
+++ b/napalm/utils/junos_views.yml
@@ -179,13 +179,13 @@ junos_lldp_neighbors_detail_table:
   args:
     interface_name: '[afgx]e*'
   item: lldp-neighbor-information
-  key: lldp-local-port-id
+  key: lldp-local-interface | lldp-local-port-id
   view: junos_lldp_neighbors_detail_view
 
 junos_lldp_neighbors_detail_view:
   fields:
-    interface: {lldp-local-port-id: unicode}
-    interface_description: {lldp-local-port-id: unicode}
+    # interface: {lldp-local-port-id: unicode}
+    interface_description: {lldp-local-interface: unicode}
     parent_interface: {lldp-local-parent-interface-name: unicode}
     remote_port: {lldp-remote-port-id: unicode}
     remote_chassis_id: {lldp-remote-chassis-id: unicode}


fooelisa:
I had an issue with different values returned by Junos on different devices (qfx) the other week. Since I haven't gotten to it yet: it would me most wonderful if you could look into how to provide multiple output files for the unit tests, to be able to provide output and check for all the different cases we encounter.

mirceaulinic:
Yup, you are right. For the moment, I just made this hack: https://github.com/napalm-automation/napalm/blob/master/test/unit/junos/mock_data/get-lldp-neighbors-information.txt to have both tags. We could probably include the firmware version in the mock filename (i.e. get-lldp-neighbors-information_13.3R6.5.txt). What do you think about that?

fooelisa:
In my case it would be the same firmware, but a different platform. So maybe: get-lldp-neighbors-information_13.3R6.5_$platform.txt ?

mirceaulinic:
On different platforms running the same firmware version? That's very interesting. Do you have an example?

fooelisa:
Actually, you are right, it probably wasn't the same firmware (not sure where the current test files are from).

The issue I had was this in get-route-engine-information on qfx10k with 15.1X53-D30.16:

Our test file has:
<memory-dram-size>1024</memory-dram-size>
My output had:
<memory-dram-size>3360 MB</memory-dram-size>

See PR #190

fooelisa:
But I still vote to include the platform in the filename. There was another issue with "environment" and IOSXR, were the VM doesn't really return any info. So probably good to have that ability to differentiate per platform.

mirceaulinic:
Sure, I'd will submit a different PR wth the required changed. 
Anyway the offline tests are based on static data defined in TestJunOSDriver.py ('version': '12.1X47-D20.7', 'fqdn': 'vsrx'). But of course, to test the behaviour on different platforms it's much easier to have different mock data and just need to change the static values in the tester.

fooelisa:
awesome :+1: thanks!

PR #207, opened 2016-02-24
Updated mock data for JunOS

mirceaulinic:
As discussed in #206. 

Still not 100% convinced that it worths doing it. Or does it?

XML RPC data goes into .xml files, unstructured (CLI) goes into .txt files.

Haven't noticed many differences across different platforms / versions.

If we are still paying attention now at this detail, could you please check later the content of the following, as it might be different on vSRX (and I do not have one to test) -- they actually contain data from MX480 running 13.3R6.5:
- get-lldp-interface-neighbors_12.1X47-D20.7_vsrx.xml
- show_ntp_associations_no-resolve_12.1X47-D20.7_vsrx.xml
- get-arp-table-information_12.1X47-D20.7_vsrx.xml
- configuration__protocols__bgp__group____bgp___protocols___configuration__12.1X47-D20.7_vsrx.xml
- _configuration__protocols__bgp__group__neighbor____group___bgp___protocols___configuration__12.1X47-D20.7_vsrx.xml

In most of the cases, where I have not observed inconsistent tags / data format, I left old data from vSRX basically.


dbarrosop:
What do you think about doing this instead?

1. We have different cases in different folders. Which means different mocked data per folder.
1. To test the same method with different mocked data you just have to change the path to where you get the mocked data from.

I am not sure if you can pass arguments to unittests methods but worst case we can pass the folder via the environment. Something like:

```
# Run default test
  - nosetests -v TestJunOSDriver:TestGetterJunOSDriver.test_get_lldp_neighbors
# Run same test with different mocked data
  - MOCKED_DATA_PATH = './junos/mocked_data/case1/' nosetests -v TestJunOSDriver:TestGetterJunOSDriver.test_get_lldp_neighbors
```

I think this solution will scale better as we can support different mocked data per test even for the same version.

mirceaulinic:
<https://www.cloudflare.com/>I have never tried to pass arguments to unittests methods. But having mock data in different folders sounds like a reasonable solution. Can be used in Travis?

dbarrosop:
If you pass it as an environment variable it will work 100% sure. A similar solution and maybe even better could be the following.

1. Mocked data is moved to ``$vendor/mocked_data/$testname_$id/``
2. The test method loops over all the subfolders of  ``$vendor/mocked_data/`` that start with its name.

So, for example, we would have folder:

* ``junos/mocked_data/test_get_lldp_neighbors_00/``
* ``junos/mocked_data/test_get_lldp_neighbors_01/``

So the test method would discover that there are two different subtests it has to run for the same test.

In this case we don't have to pass any argument to the test.

mirceaulinic:
And yeah, we would run all the tests at once for all platforms, firmware versions and other dependencies. Sounds good!

mirceaulinic:
We will need to update all testers to follow the same pattern.

dbarrosop:
I would personally close this PR and add this task to the backlog. I am planning a major refactoring so we can split drivers into different modules. That might be a good time to do this as well.

mirceaulinic:
Perfect!

PR #211, opened 2016-02-29
Defined load_template

mirceaulinic:


mirceaulinic:
@dbarrosop is there any particular reason why you opted to use ```text``` instead of ```set``` to load config on JunOS? https://github.com/napalm-automation/napalm/blob/master/napalm/junos.py#L89
Can we add an optional parameter to allow using ```set```, without changing the current behaviour?

fooelisa:
sounds good to me

mirceaulinic:
It is not a blocking point, just personal curiosity -- I prefer ```set``` over ```text``` but it is not the case to be subjective.
At the end of the day, this change will be reflected in the templates:

```
system {
  ntp {
      delete: peer 17.107.131.11;
  }
}
```
vs
```
delete system ntp peer 17.107.131.11
```

Would this be an improvement?

dbarrosop:
The reason is mainly that 'set' commands look cluttered and redundant, you get a blob of repetitive text that doesn't give you that much while 'text' looks like json and it looks pretty nice and self-describing. Saying that, it should be easy to choose which method you prefer by adding an optional argument.

jejenone:
I tend to agree with @dbarrosop arguments

mirceaulinic:
In that case I think we have an agreement to leave it like that and define the set/delete templates in text format.

fooelisa:
I like the idea of an optional arg to adjust that though. my config templates are in text also, but i use set from time to time for smaller ad-hock scripts. so if you feel like it, go ahead!

dbarrosop:
-1

Let me elaborate:

1. I don't approve the change of the API : ) If we want to do something it has to be only for JunOS and be done using optional_args. Doing it in `load_replace_candidate` and `load_merge_candidate` doesn't make too sense as that's a junos construct.
2. Remember, we can support everything we want. But we still have to support it. And more features means more chances of introducing tech debt, potential bugs, things that we will have to cover when we start doing coverage tests, etc... So let's think before adding features just for the sake of adding them.

dbarrosop:
And with that comment I am not saying we shouldn't do it. Just that I don't approve the current implementation and that we should think carefully why we want this feature and is it really worth it?

mirceaulinic:
Indeed it is not a must. Removed.

napalm/base.py
```diff
@@ -13,6 +13,12 @@
 # the License.
 import napalm.exceptions
 import sys
+import os
+
+from napalm.exceptions import DriverTemplateNotImplemented, TemplateNotImplemented, TemplateRenderException
```
dbarrosop: You already have in line 13 `import napalm.exceptions` so this one is redundant.

napalm/base.py
```diff
@@ -13,6 +13,12 @@
...
+
+from jinja2 import Environment, FileSystemLoader
```
dbarrosop: Can you import jinja2 instead? Why? To avoid local names collisions. I know code gets longer and we could get into an endless debate here about the usage of `from x import y` vs `import x` but I find way more readable:

```
try:
    do_something()
except jinja2.exceptions.Environment as e:
   do_something_else()
```

than:

```
try:
    do_something()
except Environment as e:
   do_something_else()
```

as in the second case I have to go to the imports to check who might be triggering that `Environment` exception. Another benefit is that we might need to catch later on an exception also called named `Environment` coming from another library (`Environment` is a rather generic name).

mirceaulinic: Absolutely, no problem about that :)

> Can you import jinja2 instead?

dbarrosop:
Great! Does this mean this one is ready?

mirceaulinic:
Yes, for me this is prepared.

PR #216, opened 2016-03-03
Fix issue with get_lldp_neighbors returning unknown

ktbyers:


dbarrosop:
If there is any mocked data that is no longer necessary, can you remove it, please? I am thinking of the mocked data for `show lldp neigh detail` command.

napalm/ios.py
```diff
@@ -320,59 +320,44 @@ def _check_file_exists(self, cfg_file):
             return True
         return False
 
+    def _expand_interface_name(self, interface_brief):
```
dbarrosop: Could we cache this? Maybe we could have `self.expand_interface_name_map = {}` initialized when the object is instantiated and this method could first check if the short name is there and if it's not execute the code.

The reason is that I think we could reuse this in some other parts as well. Not sure how much complexity was added in the code when @ande0581 refactored the code to return long interface names instead.

ktbyers:
FYI, I didn't correct this for show_lldp_neighbors_detail yet (I didn't get to it last night).

I was going to reuse this function _expand_interface_name() for show_lldp_neighbors_detail. I will search the ios.py code and see if there are other places the interface expansion happens.

Yes, we could easily cache this (i.e. setup an interface_map that gets created when this is first used).

ktbyers:
@dbarrosop FYI, added interface_map cache.

ktbyers:
Fixed get_lldp_neighbors_detail as well.

This should be good to go now.

PR #218, opened 2016-03-03
get_snmp_information for IOS

ande0581:


mirceaulinic:
One this PR is merged I will start working on the getters for other vendors as promised. Should not take very long.

PR #220, opened 2016-03-04
Added get_route_to in support doc & rearranged lines

mirceaulinic:
Just realised that ```get_route_to``` is missing. I rearranged the lines in such order that makes more sense to me -- but not necessarily to everyone. @dbarrosop it is just an idea, just close the PR if does not look good to you, but please add ```get_route_to```. Thanks!


dbarrosop:
No big deal, I will merge it. At some point it might be interesting to just sort it alphabetically just for simplicity : )

PR #225, opened 2016-03-09
adds a dynamic network driver loader concept

privateip:
This adds a dynamic loader to replace the currently static loading of
network device modules.  This allows NAPALM to load drivers dynamically
based on namespace

```
>>> import napalm
>>> driver = napalm.get_network_driver('napalm.eos')
>>> device = driver('veos01', 'admin', 'admin')
>>> device.open()
>>> device
<napalm.eos.EOSDriver instance at 0x10629d638>
>>> device.get_facts()
{'os_version': u'4.15.0F-2387143.4150F', 'uptime': 1309109, 'interface_list': [u'Ethernet1', u'Ethernet2', u'Ethernet3', u'Ethernet4', u'Ethernet5', u'Ethernet6', u'Ethernet7', u'Management1'], 'vendor': u'Arista', 'serial_number': u'', 'model': u'vEOS', 'hostname': u'veos01', 'fqdn': u'veos01.eng.ansible.com'}
```

Drivers do not have to be in the napalm namespace although for this to work.

```
>>> driver = napalm.get_network_driver('my.namespace.eos')
>>> device = driver('veos01', 'admin', 'admin')
>>> device.open()
>>> device
<my.namespace.eos.EOSDriver instance at 0x102193b68>
>>> device.get_facts()
```

This commit is just a proof of concept.  It includes on module (eos.py) that
has been modfied for testing purposes.

### DIFF
--- a/napalm/__init__.py
+++ b/napalm/__init__.py
@@ -12,31 +12,142 @@
 # License for the specific language governing permissions and limitations under
 # the License.
 
-from eos import EOSDriver
-from iosxr import IOSXRDriver
-from junos import JunOSDriver
-from fortios import FortiOSDriver
-from nxos import NXOSDriver
-from ibm import IBMDriver
-from ios import IOSDriver
-from pluribus import PluribusDriver
+import sys
+import imp
 
-def get_network_driver(vendor):
-    driver_mapping = {
-        'EOS': EOSDriver,
-        'ARISTA': EOSDriver,
-        'IOS-XR': IOSXRDriver,
-        'IOSXR': IOSXRDriver,
-        'JUNOS': JunOSDriver,
-        'JUNIPER': JunOSDriver,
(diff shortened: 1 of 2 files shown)


dbarrosop:
I like the idea as it solves some problems I found while breaking down napalm into different modules. However, I would like some minor adjustments. For example, I would still like to be able to call by OS name:

```
get_network_driver('eos')
```

I was thinking that if the parameter has no dots we could try to resolve the module name ourselves:

```
module = 'napalm_{v}.{v}'.format(v=vendor) if '.' not in vendor else vendor  
```

That should import automatically `napalm_eos.eos`.

Could you propose something like that here in napalm-base? https://github.com/napalm-automation/napalm

privateip:
This change is now being tracked here --> https://github.com/napalm-automation/napalm-base/pull/7

Issue #226, opened 2016-04-04
ios.py get_facts bgp neighbours producing incorrect prefix counts


lukerussell:
Hi,

get_facts for bgp_neighbors is producing incorrect results for the prefix counts with my version of IOS/XE. Images are provided by VIRL.

I strongly suspect this is related to the regex match patterns, but I don't know enough to offer a solution.

Sample output from IOSv 15.6(1)T:

`    Prefixes Current:               4         12 (Consumes 960 bytes)`
 ```
 Local Policy Denied Prefixes:    --------    -------
    AS_PATH loop:                       n/a          4
    Bestpath from this peer:             12        n/a
    Total:                               12          4
```

get_facts produces this:
```
ok: [2A-R1] => {
    "result": {
        "ansible_facts": {
            "bgp_neighbors": {
                "global": {
                    "peers": {
                        "10.130.128.2": {
                            "address_family": {
                                "ipv4": {
                                    "accepted_prefixes": 4, 
                                    "received_prefixes": 8, 
                                    "sent_prefixes": 4
                                }
                            }, 
                            "description": "", 
                            "is_enabled": true, 
                            "is_up": true, 
                            "local_as": 65200, 
                            "remote_as": 65530, 
                            "remote_id": "192.168.0.13", 
                            "uptime": 3807
                        }, 

```

Sample output from IOS-XEv 03.17.00.S (15.6(1)S):

`    Prefixes Current:               2         14 (Consumes 1680 bytes)`
````
  Local Policy Denied Prefixes:    --------    -------
    AS_PATH loop:                       n/a          2
    Bestpath from this peer:             14        n/a
    Total:                               14          2

```

get_facts produces this:
```
ok: [1A-R1] => {
    "result": {
        "ansible_facts": {
            "bgp_neighbors": {
                "global": {
                    "peers": {
                        "10.130.0.14": {
                            "address_family": {
                                "ipv4": {
                                    "accepted_prefixes": 2, 
                                    "received_prefixes": 4, 
                                    "sent_prefixes": 2
                                }
                            }, 
                            "description": "", 
                            "is_enabled": true, 
                            "is_up": true, 
                            "local_as": 65100, 
                            "remote_as": 65530, 
                            "remote_id": "192.168.0.13", 
                            "uptime": 5381
                        }
                    }, 
```

dbarrosop:
Could you share the output of the following commands?

- `show ip bgp summary`
- `show ip bgp summary | begin Neighbor `
[...]

PR #227, opened 2016-04-06
Fix get_bgp_neighbors bug with received prefixes

ktbyers:
Fix for issue #226. Verify with end-user that it fixes his issue on both Cisco IOS and Cisco IOS-XE before merging.

### DIFF
--- a/napalm/ios.py
+++ b/napalm/ios.py
@@ -44,7 +44,9 @@ class IOSDriver(NetworkDriver):
         self.candidate_cfg = optional_args.get('candidate_cfg', 'candidate_config.txt')
         self.merge_cfg = optional_args.get('merge_cfg', 'merge_config.txt')
         self.rollback_cfg = optional_args.get('rollback_cfg', 'rollback_config.txt')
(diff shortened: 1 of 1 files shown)


lukerussell:
Confirmed working with IOSv 15.6(1)T and IOS-XEv 03.17.00.S.

Thanks Kirk.

dbarrosop:
@ktbyers can you send the same PR to https://github.com/napalm-automation/napalm-ios? Main IOS code will be there. I will probably start working into turning this into a metapackage this weekend so next release of napalm will contain no code.

ktbyers:
@dbarrosop Can you merge this? It should be good to go. I am pretty sure it is right. Bug in my regex processing that unit testing didn't catch.

dbarrosop:
Please, send the PR to `napalm-ios`. I don't want to merge new code here as I am going to turn this into a metapackage for the next release. BTW, I thought the discovery code was going to live in `netmiko`, which I think completely makes sense.

ktbyers:
This was committed to napalm-ios.

PR #229, opened 2016-04-08
PROPOSAL, do not merge: add test discovery

jzohrab:
_This PR is for discussion.  If we agree that it's a good direction, I'll put more work into it._

Python unittest has a nice feature of test discovery.  This PR lets me run tests from the project root with `python -m unittest discover`.  This is nice for a few reasons:

- I _believe_ it's a standard for python projects
- good onboarding experience for new developers - tests are up and running with `git clone ...; pip install -r requirements; python -m unittest discover`

A couple of requirements for this to work:

- test directories need `__init__.py`
- test files must be named `test_*.py`

With the changes in this PR (two separate commits), running `python -m unittest discover` from the project root gave me the following:

```
(venv) ~/Documents/Projects/napalm$ python -m unittest discover
EE.E..........E..

ERROR: setUpClass (test.unit.test_ios_driver.TestConfigIOSDriver) ... [snip] ...
NetMikoTimeoutException: Connection to device timed-out: cisco_ios 127.0.0.1:12204
ERROR: test_bgp_neighbors_detail (test.unit.test_ios_driver.TestGetterIOSDriver) ... [snip] ...
NotImplementedError
ERROR: test_get_bgp_config (test.unit.test_ios_driver.TestGetterIOSDriver) ... [snip] ...
NotImplementedError
ERROR: test_get_route_to (test.unit.test_ios_driver.TestGetterIOSDriver) ... [snip] ...
NotImplementedError
```

I'm not concerned about the `NotImplementedError` at this point, though it would be something I'd want to address with a full implementation of this idea, after discussion.

I'd want to do the following for this PR to be considered complete:

- change all of the test files to `test_*.py`
- fix/deal with all of the NotImplementedErrors.  They could either be fixed or swallowed, or the tests could somehow handle them to say that there are things pending implementation.  I don't like hiding errors like that, it's good to keep them public ... but that information could be output to a report and actually committed into the main repo as part of the product roadmap.
- somehow handle the "Connection to device timed-out" - presumably these tests are actually trying to connect to a real thing.  If we're mocking the underlying classes, maybe the tests can be categorized somehow.  When I changed this test to mocking like the other tests, I got the error "No such file or directory: 'ios/initial.conf'".  Not a big deal, just fyi.
- add notes to the README or to a developer README explaining how to run the tests
- maybe add a Makefile to the project root to handle running the tests and dealing with the NotImplementedExceptions report.  It could, for example, run the tests with `--buffer` and then process the completed report.  It could also specify command-line args that would let us toggle between mock devices during tests vs real devices.

I hope to get a chance to discuss this on Slack (I'm on networktocode as jzohrab), and if we agree that this is a good direction I'd be happy to work on getting this fleshed out for all of the classes in `test/unit` and getting the tests to run cleanly on a fresh install.

Thanks very much!


jzohrab:
The travis test failures are unrelated to this change. :-)

dbarrosop:
I am going to close this PR for the simple reason that some people is already working on standarizing this. We decided to follow the same guidelines as openstack. The main reason because we can reuse all the tools they already wrote and a lot of projects seem to be doing the same so it's starting to feel like the 'de facto standard'.

Let me comment a few parts though:

> good onboarding experience for new developers - tests are up and running with `git clone ...; pip install -r requirements; python -m unittest discover`

Main problem with that is that tests that involve configuration changes are not mocked so either you start all needed vagrant boxes to run all the tests or you have access to real equipment. So the result is just going to be that most of the tests are just going to fail.

> fix/deal with all of the NotImplementedErrors. They could either be fixed or swallowed

No, I don't really want to do that. Why? Because you don't want to hide it in the code and swallowing it in the tests might lead to untested code. In addition, it is very useful to see the current state of drivers as with one command you can see which methods are implemented and which ones are not.

> presumably these tests are actually trying to connect to a real thing. If we're mocking the underlying classes, maybe the tests can be categorized somehow.

Tests are already categorized. You have two:

* Tests that inherit from `TestConfigNetworkDriver`. These perform tests that involve config management methods. They require either a VM or a real device. This is the one that is probably failing.
* Tests that inherit from `TestGettersNetworkDriver`. These test `get_*` methods. These are mocked so you can just run them.

This is the reason why I mentioned in the first place that the autodiscover is not a good idea for this project.

> When I changed this test to mocking like the other tests, I got the error "No such file or directory: 'ios/initial.conf'"

That shouldn't happen. Which tests did you change? Tests from `TestConfigNetworkDriver` can't be mocked and tests from `TestGettersNetworkDriver` are mocked already so you shouldn't have any issue at all. Could you provide more info? Create an issue and instructions to reproduce it, please.

> add notes to the README or to a developer README explaining how to run the tests

I agree with this. I was planning to write a CONTRIBUTE.md file with some guidelines : ) 

> maybe add a Makefile to the project root to handle running the tests and dealing with the 

Yes, I agree with this as well. This is where the openstack stuff will come in handy : )

P.S: I am closing this PR for the moment because there is another effort in parallel that conflicts with this. However, feel free to keep the discussion here.

Issue #230, opened 2016-04-08
Request: intro materials


jzohrab:
I'm looking to try out Napalm, but I have ​_no idea_​ where to start.  I've read the docs on github.  I have a coding background and CCNA-level networking understanding, which should be sufficient to try out the basics.  Familiar with Python, have used Chef and am happy to try out Ansible.  I'm okay to try out Ansible.  My ideal tutorial would include the following:

- setting up a vagrant box or similar for local testing
- setting up Ansible (seems like path of least resistance)
- creating a simple config
- using Ansible to push it out to the device

I'd be interested in creating an intro tutorial like the above, if I can get some guidance.  I have some networking colleagues who would likely benefit from such material as well.

ande0581:
@jzohrab You could check out some of kirks byers articles.

https://pynet.twb-tech.com/

He also has a formal course that covers a lot of what you are asking about.

https://pynet.twb-tech.com/class.html

ktbyers:
@jzohrab I wrote this article here regarding NAPALM and Cisco ios (which might help):

https://pynet.twb-tech.com/blog/automation/napalm-ios.html

jzohrab:
Thanks @ande0581 and @ktbyers for your input, appreciated.  @ktbyers I'd read that already, thanks for posting it.

I think what I'd like to see - for myself, and perhaps as a quickstart for the project - would be a simple set of instructions to set up a lab environment locally, including Ansible etc, so that I have something running that I can immediately hack around with.  It's not a place for me to learn about Ansible or other things, but rather just for me to get up and running.  I often find that setting up tools etc is an exercise in frustration, and having a low-friction entry is valuable.

There's already a Vagrantfile in the code here.  I submitted a PR for review and discussion that I hope would also simplify devs getting involved with the project.

I think all the pieces are in place, and it's just a matter of putting together a simple tutorial for people like me to follow to get something up and running very quickly, so I can get an end-to-end understanding of what Ansible can do on my own box.  The vagrantfile stuff might be tough with some vendors due to licensing, but maybe one of the other vendors has a free/student/whatever version that would be enough.

dbarrosop:
Hello @jzohrab,
there are a lot of resources out there to set up a lab. For example, I wrote:

https://www.dravetech.com/blog/2016/01/08/vagrant-for-network-engineers.html
https://www.dravetech.com/blog/2016/01/14/vagrant_box_ios_xr.html

[...]

PR #235, opened 2016-04-14
Add quickstart tutorial.

jzohrab:
Compilation of information from several sources into a single canonical document for new users.

I added a separate Vagrantfile, rather than reusing the one at test/unit/Vagrant/Vagrantfile, as I wanted to keep the vagrantfile short focussed, and to reduce the version numbering issues (the new Vagrantfile hardcodes the box name as "vEOS-lab-quickstart" per the tutorial instructions).

Addresses https://github.com/napalm-automation/napalm/issues/230

### DIFF
--- a/docs/tutorials/quickstart.rst
+++ b/docs/tutorials/quickstart.rst
@@ -1,0 +1,109 @@
+Quickstart
+==========
+
+This tutorial gets you up-and-running quickly with NAPALM in a local virtual environment so you can see it in action.
+
+.. note::  This tutorial does not cover fully automated configuration management (e.g., using NAPALM in conjunction with Ansible, Chef, Salt, etc.).  We hope that tutorials for these tools will be contributed soon so that you can evaluate the options for your particular environment.
+
+Requirements
+------------
+
+You'll need a few tools:
+
+* Python
+* `pip <https://pip.pypa.io/en/stable/installing/>`_: The PyPA recommended tool for installing Python packages
+* `virtualenv <http://docs.python-guide.org/en/latest/dev/virtualenvs/>`_: a tool to create isolated Python environments
+* `VirtualBox <https://www.virtualbox.org/>`_: a software virtualization tool
+* `Vagrant <https://www.vagrantup.com/downloads.html>`_: a command line utility for managing the lifecycle of virtual machines
+
+As the focus of this tutorial is NAPALM, we don't even scratch the surface of these tools.  If you're not familiar with them, please do some research [#f1]_ as they will be an important part of your development/ops toolkit.
+
+Installation
+------------
+
+If you haven't already installed NAPALM, you can run the code from a virtual environment in the root directory::
+
+    $ virtualenv venv                        # create environment
+    $ source venv/bin/activate               # start environment
+    $ sudo pip install -r requirements.txt   # install dependencies
+    $ python setup.py install                # install NAPALM
+
+Arista vEOS
+-----------
+
+We'll use Arista, which can be downloaded for free from the Arista site.
+
+Create an account at https://www.arista.com/en/user-registration, and go to https://www.arista.com/en/support/software-download.
+
+Download the latest "vEOS-lab-<version>-virtualbox.box" listed in the vEOS folder at the bottom of the page.
+
+Add it to your vagrant box list as "vEOS-lab-quickstart"::
+
+    $ vagrant box add --name vEOS-lab-quickstart ~/Downloads/vEOS-lab-<version>-virtualbox.box
+    $ vagrant box list
+    vEOS-lab-quickstart (virtualbox, 0)
+
+Notes:
+
+* we're setting the ``--name`` parameter here so that the tutorial's Vagrantfile will work out-of-the-box.  You may wish to also ``vagrant box add`` this file with the correct version
+* ``vagrant box add`` copies downloaded file to a designated directory (e.g., for Mac OS X and Linux: ``~/.vagrant.d/boxes``, Windows: ``C:/Users/USERNAME/.vagrant.d/boxes``).
+
+Starting Vagrant
+----------------
+
+The Vagrantfile (in this directory) creates a base box and a vEOS box when you call "vagrant up"::
+
+    $ cd docs/tutorials
+    $ vagrant up
+    ... [omitted] ...
+
+    $ vagrant status
+    Current machine states:
+    base                      running (virtualbox)
+    eos                       running (virtualbox)
+
+You may see some errors when the eos box is getting created [#f2]_.
+
+
+Using the NAPALM command-line client
+------------------------------------
+
+You can now try the example commands at http://napalm.readthedocs.org/en/latest/cli.html.  Cd to the ``test/unit/eos`` directory which contains the .conf files::
+
+    $ cd PROJECT_ROOT/test/unit/eos
+
+    # dry run.
+    # (When prompted, the password is "vagrant")
+    $ cl_napalm_configure \
+      --user vagrant \
+      --vendor eos \
+      --strategy replace \
+      --optional_args 'port=12443' \
+      --dry-run \
+      new_good.conf localhost
+
+
+Try the other examples and commands given at http://napalm.readthedocs.org/en/latest/cli.html.
+
+Shutting down
+-------------
+
+    $ cd PROJECT_ROOT/docs/tutorials
+    $ vagrant destroy -f
+    $ deactivate           # exit the virtualenv environment
+
+Next Steps
+----------
+
+There are many possible steps you could take next:
+
+* create Vagrant boxes for other devices
+* explore using configuration management tools (Ansible, Chef, Salt, etc.)
+
+Thanks for trying NAPALM!  Please contribute to this documentation and help grow the NAPALM community!
+
+
+.. [#f1] Vagrant's `getting started guide <https://www.vagrantup.com/docs/getting-started/>`_ is worth reading and working through.
+
+.. [#f2] Currently, ``vagrant up`` with the eos box prints some warnings: "No guest additions were detected on the base box for this VM! Guest additions are required for forwarded ports, shared folders, host only networking, and more. If SSH fails on this machine, please install the guest additions and repackage the box to continue. This is not an error message; everything may continue to work properly, in which case you may ignore this message."  This is not a reassuring message, but everything still seems to work correctly.
+
--- a/docs/tutorials/quickstart.rst
+++ b/docs/tutorials/quickstart.rst
@@ -12,7 +12,6 @@ You'll need a few tools:
 
 * Python
 * `pip <https://pip.pypa.io/en/stable/installing/>`_: The PyPA recommended tool for installing Python packages
-* `virtualenv <http://docs.python-guide.org/en/latest/dev/virtualenvs/>`_: a tool to create isolated Python environments
 * `VirtualBox <https://www.virtualbox.org/>`_: a software virtualization tool
 * `Vagrant <https://www.vagrantup.com/downloads.html>`_: a command line utility for managing the lifecycle of virtual machines
 
@@ -21,12 +20,9 @@ As the focus of this tutorial is NAPALM, we don't even scratch the surface of th
 Installation
 ------------
 
-If you haven't already installed NAPALM, you can run the code from a virtual environment in the root directory::
+Install NAPALM with pip::
 
-    $ virtualenv venv                        # create environment
-    $ source venv/bin/activate               # start environment
-    $ sudo pip install -r requirements.txt   # install dependencies
-    $ python setup.py install                # install NAPALM
+    pip install napalm
 
 Arista vEOS
 -----------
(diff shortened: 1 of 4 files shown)


docs/tutorials/quickstart.rst
```diff
@@ -0,0 +1,109 @@
...
+
+We'll use Arista, which can be downloaded for free from the Arista site.
+
+Create an account at https://www.arista.com/en/user-registration, and go to https://www.arista.com/en/support/software-download.
+
+Download the latest "vEOS-lab-<version>-virtualbox.box" listed in the vEOS folder at the bottom of the page.
+
+Add it to your vagrant box list as "vEOS-lab-quickstart"::
+
+    $ vagrant box add --name vEOS-lab-quickstart ~/Downloads/vEOS-lab-<version>-virtualbox.box
```
dbarrosop: I would use their generic name `vEOS-lab-<version>-virtualbox` instead of a specific one. It's supposed to be portable... : )

dbarrosop:
I really like it, I think a few things might be interesting as well:

1. Change the current tutorials to use the same vagrantfile you have. With that I mean: http://napalm.readthedocs.org/en/latest/tutorials/index.html and http://napalm.readthedocs.org/en/latest/cli.html
1. Reference all the tutorials in your document. I think it's interesting that after seeing how simple the cli is they can jump into some python code and see that's it's almost as simple and not that different from using just the CLIs they are used to... Just a quick introduction to python code and then a link to the longer tutorial in the same way you did the `cli` introduction.

Does it make sense?

docs/tutorials/quickstart.rst
```diff
@@ -0,0 +1,109 @@
...
+
+Installation
+------------
+
+If you haven't already installed NAPALM, you can run the code from a virtual environment in the root directory::
+
+    $ virtualenv venv                        # create environment
+    $ source venv/bin/activate               # start environment
+    $ sudo pip install -r requirements.txt   # install dependencies
+    $ python setup.py install                # install NAPALM
```
dbarrosop: Don't use this method. Use pip instead, please. There are some caveats when installing the new modular code with this mechanism.

docs/tutorials/quickstart.rst
```diff
@@ -0,0 +1,109 @@
...
+
+Notes:
+
+* we're setting the ``--name`` parameter here so that the tutorial's Vagrantfile will work out-of-the-box.  You may wish to also ``vagrant box add`` this file with the correct version
+* ``vagrant box add`` copies downloaded file to a designated directory (e.g., for Mac OS X and Linux: ``~/.vagrant.d/boxes``, Windows: ``C:/Users/USERNAME/.vagrant.d/boxes``).
+
+Starting Vagrant
+----------------
+
+The Vagrantfile (in this directory) creates a base box and a vEOS box when you call "vagrant up"::
```
dbarrosop: As pip is the preferred method to install napalm I guess that you should mention first that either they download the Vagrantfile or they just create the file and copy & paste the content. I guess downloading is better to avoid issues.

docs/tutorials/quickstart.rst
jzohrab: Yes, the reason for setting the `--name` like this is so that the vagrantfile can find the box, without them changing the vagrantfile.  The file name is still `...<version>..`.  I specifically didn't want to say "update the Vagrantfile to use the same version you downloaded earlier".

docs/tutorials/quickstart.rst
jzohrab: Ah right, I was thinking that they'd clone this repo, which is likely too much for some people.  Got it, thanks.

docs/tutorials/quickstart.rst
jzohrab: I see.  I was thinking about this from the viewpoint of a dev who'd checked everything out.  So should this be `sudo pip install napalm`?

jzohrab:
@dbarrosop - re "I think a few things might be interesting as well" - yes, I agree.  How about we work on that as separate issues once we work out the points you're raised, and this gets into master?  This keeps this PR's scope focused.  I agree with what you say, it does provide a nice starting point for those tutorials and would provide something concrete to work with.  Presumably, "set up your lab" would be split out into a separate section entirely, and referred to from those tutorials.

docs/tutorials/quickstart.rst
dbarrosop: What I meant is that I think it's better if we stick with the default name on the Vagrantfile. Maybe some people already have it or might start using it for other purposes.

docs/tutorials/quickstart.rst
dbarrosop: From a developer's point of view is a bit more complex than that. Wait until the refactoring is merged because when it's done setting up the dev environment will involve cloning a few repos and tweaking the pythonpath. Don't worry, is not as hard as it sounds and it can be easily automated ;)

BTW, you can always do `pip install -e .` if you are inside the folder to install the cloned repository with pip.

docs/tutorials/quickstart.rst
jzohrab: Right, I see.  Since we're talking about people installing with pip and copying/downloading the vagrantfile, rather than doing a `git checkout` as I'd originally thought, then using the name as-is makes sense.  If it was strict version control, I'd say use the name as suggested - but that implies that devs would be using this, and they should know enough anyway.  Sounds good, will change.

docs/tutorials/quickstart.rst
jzohrab: ok, np.  I see the instructions on the README, will change.  Thanks.

jzohrab:
Will make the changes suggested, repush the branch, and then will ping you to review again.  I won't rebase and squash the branch unless you'd prefer that for git commit history.  Thanks for all the feedback, appreciated.

jzohrab:
Making some changes and repushing.

jzohrab:
@dbarrosop - I committed some changes.  the docs look OK to me when generated with sphinx locally.  Once this gets checked and  into master, we can start to iterate on the other suggestions you had.

docs/tutorials/quickstart.rst
```diff
@@ -0,0 +1,109 @@
...
+    $ cl_napalm_configure \
+      --user vagrant \
+      --vendor eos \
+      --strategy replace \
+      --optional_args 'port=12443' \
+      --dry-run \
+      new_good.conf localhost
+
+
+Try the other examples and commands given at http://napalm.readthedocs.org/en/latest/cli.html.
```
dbarrosop: Maybe we should add instructions on the cli page to download the rest of the files as you did here. Otherwise, going from here to the link and keep following it's going to result in failures and frustrations : )

docs/tutorials/quickstart.rst
```diff
@@ -0,0 +1,109 @@
...
+    $ cd PROJECT_ROOT/docs/tutorials
+    $ vagrant destroy -f
+    $ deactivate           # exit the virtualenv environment
+
+Next Steps
+----------
+
+There are many possible steps you could take next:
+
+* create Vagrant boxes for other devices
```
dbarrosop: Maybe you can link to instructions/examples on what you mean with that? For example, I wrote this post some time ago https://www.dravetech.com/blog/2016/01/14/vagrant_box_ios_xr.html. However, you can link something else if you know of some other resource.

docs/tutorials/quickstart.rst
```diff
@@ -0,0 +1,109 @@
...
+* explore using configuration management tools (Ansible, Chef, Salt, etc.)
```
dbarrosop: Same here, some links might be interesting.

docs/tutorials/quickstart.rst
```diff
@@ -0,0 +1,109 @@
...
+      --dry-run \
+      new_good.conf localhost
+
+
+Try the other examples and commands given at http://napalm.readthedocs.org/en/latest/cli.html.
+
+Shutting down
+-------------
+
+    $ cd PROJECT_ROOT/docs/tutorials
```
dbarrosop: This is the first time this path appears, I guess it's some leftover. By the way, this gave me an idea. Do you want me to create a `napalm-tutorials` repo where you can put the files needed to run this and other tutorials? No documentation there, just files so you can start with:

```
git clone https://github.com/napalm-automation/napalm-tutorials
cd napalm-tutorials/quickstart```

and take it from there.

docs/tutorials/quickstart.rst
jzohrab: Reflecting further on test files: I think it would be best to actually copy the test files from the eos repo into this repo, and to use them here.  The files are small, and can be made even smaller.  Then all tutorials could reference the same files, and they could all be updated.  Let's do that in a separate PR.

docs/tutorials/quickstart.rst
jzohrab: Right, I needed a way for people to get back to the folder with the Vagrantfile, typing "vagrant destroy" in any other directory doesn't work.  This was the cleanest way I could think of to write it.

I would recommend keeping the tutorial in alongside the rest of the code, rather than as a separate repo.  Ideally, the tutorial instructions should be tested alongside the code, to ensure that the instructions stay current.  By "testing" here, I mean actual automated testing, and not relying on people to remember to test things when they change code.  This isn't actually done here yet, we could backlog that idea.

@dbarrosop - can we merge this into master?  I have two reasons for the request:

1) It looks like we're talking about improvements that we can make iteratively.  "Release early, release often" is a good thing to follow, sometimes.  Also, once it's in, everyone can make contributions to it and it could go faster.
2) we're talking about changing some of the other tutorials, and I don't want the scope to creep.  Small PRs are good PRs!
3) A colleague was asking for this quickstart material, if this gets into master and onto "readthedocs" it will help our team. :-)

If you would rather keep this separate, I can continue to iterate on this, but I'd prefer to merge it and add issues that I'll deal with quickly in separate smaller PRs.

dbarrosop: WALL OF TEXT ALERT!!! : )

Honestly, I would rather not merge yet. I see the value of what we are trying to do but I don't think we are there yet. I also think there is some confusion here. But that might be due to the latest changes in NAPALM or due to the lack of "common goals" for this PR.

First, some notes and clarifications regarding napalm.

> I would recommend keeping the tutorial in alongside the rest of the code, rather than as a separate repo

The repo `napalm` contains no code. It's only a placeholder for documentation and a wrapper to install all the drivers. The common code of napalm lives now in `napalm-base` and each driver has a dedicated repo, ie, `napalm-eos`, `napalm-ios`, etc... There was some discussion in the mailing list about this a month or two ago and the latest release notes has also some explanations around this and why we did it. In any case, what I meant by having a dedicated repo where the support files, not the content. However, I don't mind having the support files in the main `napalm` repo either, at the end of the day, this is one of the main reasons of this repo.

> the tutorial instructions should be tested alongside the code, to ensure that the instructions stay current. By "testing" here, I mean actual automated testing, and not relying on people to remember to test things when they change code. This isn't actually done here yet, we could backlog that idea.

That is not necessary. Two reasons:

1. The API is stable, that means that code written following napalm API should work, always, no matter what.
2. Each driver has it's own set of integration tests that are run automatically every time someone pushes new code. So we should find bugs and issues there without the need of testing the quickstart. I am not claiming the tests are perfect so feel free to comment if you think something could be improved. Bottom line here is that we already have automated testing covering this.

> "Release early, release often" is a good thing to follow, sometimes.

Agree, that was mostly the reason behind the latest refactoring. That and smaller PRs. However, features should be complete, releasing something that doesn't work might just frustrate people and make them lose confidence on the project so I am all for releasing quick and fast, as long as is complete. Specially if it's documentation. A common concern on OSS is that documentation is confusing or that you can't just follow it because you always find surprises.

> A colleague was asking for this quickstart material, if this gets into master and onto "readthedocs" it will help our team. :-)

But this is not about one particular's use case. It's about the community. Being a contributor has the advantage that you can focus on your particular problem but we should still aim to build something as generic as possible so everybody could get the benefit of it. I know that's probably not what you meant but I felt obliged to clarify anyway ; )

Back to the PR itself. Right now, I think that most of the content of this PR mostly duplicates work already done in the tutorials. So let me try to explain what I envision for this PR:

1. The target audience should be developers/network_engineers/users_in_general that heard about napalm and what to see it in action and maybe try it by themselves. The key word here is **users**. Anyone that might be considering napalm as their main interface to the network. They could be people looking into ansible, writing their own scripts or something else. But definitively not people trying to contribute to napalm. For that people we will have some other document.
2. The goal of the document should be to be able to copy & paste a few lines and see different things you can do with napalm:
   1. Change configuration
   1. Get diffs.
   1. Get LLDP neighbors/BGP/interfaces/whatever
   1. Test the CLI tool. 
3. Provide a list of resources to explore a bit further.

Right now, most of the stuff is covered on the tutorials, what is missing is:

1. How to quickly build a lab so you don't have to test on your own network.
2. Fixing the tutorials so you can actually copy & paste them.

In summary, my problem with the PR right now is that overlaps a lot with the tutorials. I would rather have this PR just telling you how to start a lab to move forward onto the rest of the tutorials and adapt the tutorials so they can be reproduced with the lab.

I hope it makes sense, otherwise, feel free to send me back a wall of text :P

jzohrab:
Closing in favor of PR #238 , we'll iterate on that branch.  Thanks @dbarrosop for the feedback.

PR #239, opened 2016-04-20
Quickstart outline

jzohrab:
This PR is a follow-up of https://github.com/napalm-automation/napalm/pull/238.

I haven't tried the tutorial end-to-end yet.  I believe there are some small differences that we should look into, but before fixing those I wanted to ensure that this work was on the right track.

If this is ok, let's merge into the feature branch and then create any additional issues as needed so they can be addressed.  Thanks!


dbarrosop:
Sorry it's taking so long for me to look into this one. I will try to find some time to look into this tomorrow. I have been mostly stuck on planes and meetings for the last week and it's about to get even worse.

jzohrab:
Hey David, np.  I think this is a good start, so when you get around to it merge it in.  I may add some commits to it if I get the time.  Right now the python pages aren't clear enough, would like to actually commit some python code to the repo as part of the exercise for people to run and work with.

PR #241, opened 2016-04-26
fixing format issue in documentation

fooelisa:


mirceaulinic:
👍

docs/base.rst
```diff
@@ -1,7 +1,7 @@
 NetworkDriver
 -------------
 
-.. autoclass:: napalm.base.NetworkDriver
+.. autoclass:: napalm_base.base.NetworkDriver
```
mirceaulinic: If imported like ``` from napalm_base import NetworkDriver``` here probably should be just ```napalm.NetworkDriver```

PR #244, opened 2016-04-30
Automate generation of the documentation

dbarrosop:
Maintaining the support of the getters is a complete nightmare. To avoid the issue I would like to automate generating the `Getters support matrix` as much as possible.
### Proposed Solution

The idea is as follows:
1. Modules will start using the gitflow model. Or at least partially. We will have a `master` (production code) and a `devel` branch.
2. Before merging `devel` into `branch` we will run `nosetests -v TestEOSDriver.py:TestGetterEOSDriver --with-json-extended` (or the equivalent for the given driver) and push the results to `master`.
3. `napalm` will automatically download the output of the tests from each repo and use it to generate the support matrix. For example: https://raw.githubusercontent.com/napalm-automation/napalm-eos/master/test/unit/nosetests.json (it doesn't exist yet)

We can probably extend this to automate other pieces of the documentation.
### POC

This PR contains a POC with static data. The final solution would download the content from each repo.


dbarrosop:
@mirceaulinic @fooelisa @ktbyers thoughts?

mirceaulinic:
Another super idea of @dbarrosop! I like it!

I would just like to propose two minor details:

- insignificant detail, but it would make more sense to me: can we use the name ```develop``` instead of ```devel```? 
- can we start using tags? That way we can get rid of the ```master``` branch and we'd have ```devel```/```develop``` and one for each release, having a better history overview.

dbarrosop:
The milestone helped but I still have to  copy paste a few times and fix the formatting. An alternative would be to have a `changelog` message on the PR so at least I have something to copy paste that I don't have to reformat (the title of the PR is not that friendly).

dbarrosop:
> insignificant detail, but it would make more sense to me: can we use the name develop instead of devel?

Of course!

> can we start using tags? That way we can get rid of the master branch and we'd have devel/develop and one for each release, having a better history overview.

I want to accommodate to the gitflow model. I am not entirely familiar with it butt I feel that it's better to reuse tools/workflows that are already in place. People might know already how to use it and it's easier to point out to already written documentation/tutorials on how to use it. Would that work for you? Or does that gitflow model miss something you think that might be useful?

dbarrosop:
Another thing I noticed is that we already have tags. Because I have been using the release mechanism of github a tag per release has been created automatically. So the only effective change that we have to do (because we use forks instead of branches) is that people will have to send PRs against `develop` instead of `master` and I will have to merge `develop` into `master` for releasing. Other than that nothing really has to change.

mirceaulinic:
@dbarrosop Awesome! I have not noticed that Github creates tags automatically! Usually I was adding tags from the command line.
I think for that you'd need to make ```develop``` as the main branch, to be clear for all developers. Moreover, this will be the first suggestion when opening the pull request. What do you think?

dbarrosop:
I already did that : )

dbarrosop:
Obsoleted by #324

Issue #249, opened 2016-05-11
Change 'cl_napalm_configure' reference in setup.py, remove redundant setup


jzohrab:
`cl_napalm_configure` currently fails:

```
Traceback (most recent call last):
  File "/usr/local/bin/cl_napalm_configure", line 9, in <module>
    load_entry_point('napalm==1.0.1', 'console_scripts', 'cl_napalm_configure')()
  File "/Library/Python/2.7/site-packages/pkg_resources/__init__.py", line 542, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/Library/Python/2.7/site-packages/pkg_resources/__init__.py", line 2569, in load_entry_point
    return ep.load()
  File "/Library/Python/2.7/site-packages/pkg_resources/__init__.py", line 2229, in load
    return self.resolve()
  File "/Library/Python/2.7/site-packages/pkg_resources/__init__.py", line 2235, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
ImportError: No module named clitools.cl_napalm_configure
```

The `cl_napalm_configure` alias is currently set in both this repo and napalm-base:

https://github.com/napalm-automation/napalm/blob/develop/setup.py#L31:
`'cl_napalm_configure=napalm.clitools.cl_napalm_configure:main',`

https://github.com/napalm-automation/napalm-base/blob/develop/setup.py#L33:
`'cl_napalm_configure=napalm.clitools.cl_napalm_configure:main',`

The clitools exist in [napalm_base](https://github.com/napalm-automation/napalm-base/tree/develop/napalm_base/clitools), so I believe it should be removed from the napalm repo, and fixed in the napalm-base repo.

PR #250, opened 2016-05-11
Remove bad clitools entry point.

jzohrab:
clitools has moved to napalm_base module.

Proposed fix for part of https://github.com/napalm-automation/napalm/issues/249

I didn't bump the version, I assume that the maintainers do that.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,10 +25,5 @@ setup(
     ],
     url="https://github.com/napalm-automation/napalm",
     include_package_data=True,
-    install_requires=reqs,
-    entry_points={
-        'console_scripts': [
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -25,10 +25,5 @@
     ],
     url="https://github.com/napalm-automation/napalm",
     include_package_data=True,
-    install_requires=reqs,
```
dbarrosop: We need this line : )

jzohrab: @dbarrosop: It's below, I just removed the comma.

jzohrab: Unless you mean the line needs to have the comma after it - a style issue?

dbarrosop: No, I just missed it : )

PR #263, opened 2016-05-25
Add note to clarify NXOS minumum supported version

DiogoAndre:
Adding a small note to clarify the minimum supported version for Cisco Nexus devices.


jejenone:
on Nexus 5000 and 6000 you need 7.2 minimum.

DiogoAndre:
@jejenone Yes, exactly.  I added the note because I didn't want to add complexity to the table, and 6.2 is the absolute initial version for the overall NXOS. 

Should we just use the minimum version that covers the most devices?t

PR #265, opened 2016-06-13
Add load_replace tutorial example.

jzohrab:
An initial example for discussion with @dbarrosop and anyone else interested.

This sample code builds on the virtual lab, so people can try out a running script immediately.

We could merge this into the quickstart branch and iterate on it, or I can fix up minor things here beforehand.

Future todos:

- add other samples.  Slack chat w/ DB: "would be cool to organize scripts by major feature: `context manager`, `load_merge`, `load_replace`, `load_template` ... then showing the `compare`, `discard`, `rollback`, etc... you would actually use the functional tests defined here as general idea"
- figure out how the docs should be arranged
- _somehow_ add unit tests to ensure that these things continue to work in the future.  I've manually tried it out and it works fine, but that's really not good enough for an ongoing project.

### DIFF
--- a/docs/tutorials/index.rst
+++ b/docs/tutorials/index.rst
@@ -7,6 +7,7 @@ Tutorials
    outline
    installation
    lab
+   samples
    ../cli.rst
    first_steps_config
    context_manager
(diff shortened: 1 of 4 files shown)


jzohrab:
@dbarrosop, since this is going into different branch and not master, can this be merged?

dbarrosop:
Yeah, no problem. At some point, probably by the end of summer, I want to revisit this branch and probably merge it into master after a few additions I have in mind regarding the testing framework and other contributors tips.

PR #268, opened 2016-06-18
ISSUE TEMPLATE

dbarrosop:
It's based on the ansible one because I thought it made sense to use it as a base to start discussing it.

### DIFF
--- a/.github/ISSUE_TEMPLATE
+++ b/.github/ISSUE_TEMPLATE
@@ -16,16 +16,3 @@ Please, open the issue in the correct repo to help us keep things organized : )
 -->
 
 ### Description of Issue/Question
-
-### Setup
-
-### NAPALM version
-(Paste verbatim output from “pip freeze | grep napalm” between quotes below)
-
-### Network OS version
-(Paste verbatim output from “pip freeze | grep napalm” between quotes below)
-
-### Steps to Reproduce the Issue
-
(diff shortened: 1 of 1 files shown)


mirceaulinic:
We must design the issue templates keeping in mind that they must be present also here, in the main repo, but also in each sub-driver (e.g. a JunOS issue would be preferably open in napalm-junos etc.).

Although their purpose is quite obvious and helpful, I kinda find ```<!-- -->``` blocks slightly harder to digest. It would be also a big step from nothing to a pretty big form that must be filled in.

What do you think about a simpler model like:

```

### Setup

### NAPALM version
(Paste verbatim output from “pip freeze | grep napalm” between quotes below)

### Network OS version

### Error Traceback
(Paste the complete traceback of the exception)
```

dbarrosop:
We don't have to use the same exact template for every model. For example, for IOS-XR we could have some reminders like 'make sure you are running at least version X and xml tty is enabled` and probably have the exact command for the `show version`. Actually, in this template we should probably have instructions to go to the correct repo.

dbarrosop:
I added links to other repos, instructions on when to go to which one and I also used your template. However, I am thinking that your template makes more sense on the rest of the repos but not here. This one should probably be an open form.

mirceaulinic:
Let's merge this one @dbarrosop

PR #267, opened 2016-06-17
Some guidelines for submitting issues

mirceaulinic:
From time to time there are folks encountering issues using NAPALM and they simply paste a traceback of the error(s) they encounter, without providing any other piece of information, making debugging almost impossible till we get back and ask pretty much the same questions each and every time.
I suspect we will see this pattern more and more often and we should provide some guidelines to speed up this process.

@dbarrosop, @jzohrab Can you please give a helping hand?

PR #269, opened 2016-06-21
Raise RuntimeError when NAPALM is not installed in Python 2.7 environment

mirceaulinic:
Comes together with https://github.com/napalm-automation/napalm/pull/268 and https://github.com/napalm-automation/napalm/pull/267 to make sure that NAPALM is used in the right environment.

### DIFF
--- a/napalm/__init__.py
+++ b/napalm/__init__.py
@@ -12,4 +12,9 @@
 # License for the specific language governing permissions and limitations under
 # the License.
 
+import sys
+
 from napalm_base import get_network_driver
+
+if not(sys.version_info.major == 2 and sys.version_info.minor == 7):
+    raise RuntimeError('You need Python 2.7 for NAPALM.')


dbarrosop:
Good idea! : )

dbarrosop:
However, this is probably more beneficial if we have it in `napalm_base/base.py`. That file is **always** imported no matter if we are importing a driver directly or using `napalm` or `napalm-base` as entry points.

mirceaulinic:
You are right, I will move it to napalm-base then

jejenone:
That will block deployment of Napalm in Nexus Open Agent Container

see http://www.cisco.com/c/en/us/td/docs/switches/datacenter/nexus5000/sw/programmability/guide/b_Cisco_Nexus_5K6K_Series_NX-OS_Programmability_Guide/open_agent_container.html

It is running a CentOS 6.7 that has python 2.6.6 only. I think if we can avoid python 2.7 specific for the moment it would be preferable.

dbarrosop:
@jejenone problem is that the code is not guaranteed to work on python <2.7. I know the installer doesn't work properly on older versions of python.

However, being an LXC it shouldn't be easy to install python2.7 inside it. Or is Cisco limiting what you can do there somehow?

PR #283, opened 2016-08-12
add dependency installation commands to README

bordeltabernacle:
Using Napalm with Ansible, it was not finding `napalm-ios` even though it was installed via pip.
```
fatal: [192.168.0.2]: FAILED! => {"changed": false, "failed": true, "msg": "cannot connect to device: Cannot import \"napalm_ios\". Is the library installed?"}
```
After some digging it turned out the [cryptography](https://cryptography.io/en/latest/) package had not installed properly. This was fixed by installing `libssl` & `libffi`.

I took the liberty of adding this to the readme, rather than raising an issue, I hope you don't mind, if you'd rather it was mentioned elsewhere, let me know, or of course feel free to not merge, obviously. :smile: 
I can also add it to the docs?

Thanks to [this SO answer](https://stackoverflow.com/questions/22073516/failed-to-install-python-cryptography-package-with-pip-and-setup-py#22210069)

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,13 @@ pip install napalm
 
 That will install all the drivers currently available.
 
+When installing on Linux, to ensure all dependencies are met, use the following commands:
+
+**Debian/Ubuntu**:
+`sudo apt-get install build-essential libssl-dev libffi-dev python-dev`
+
+**Fedora and RHEL-derivatives**:
+`sudo yum install gcc libffi-devel python-devel openssl-devel`
 
 Partial Installation
 --------------------


mirceaulinic:
Hi @bordeltabernacle thanks for sharing your experience.

We are aware of this drawback when loading NAPALM drivers dynamically  - there's already an issue raised in napalm-base https://github.com/napalm-automation/napalm-base/issues/57 and will be addressed most probably during the upcoming [hackaton](http://napalm.readthedocs.io/en/latest/hackathons/hackathon2016/index.html).

Unfortunately I don't like the idea of adding the details you proposed in the description of the main lib README since they are specific to napalm-ios (probably for others too, but not for all drivers). 

```cryptograpy``` is not a direct dependency of NAPALM and the libraries and the underneath libraries should specify a proper installation as junos-enzc does, for example: https://github.com/Juniper/py-junos-eznc/blob/master/INSTALL-FEDORA.md or https://github.com/Juniper/py-junos-eznc/blob/master/INSTALL-FREEBSD.md etc. Unfortunately many of them don't have a proper documentation.
I think a fair tradeoff would be including an ```installation``` section under [tutorials](http://napalm.readthedocs.io/en/latest/tutorials/index.html) or even under [supported devices](http://napalm.readthedocs.io/en/latest/support/index.html).

Feel free to begin working on a draft - for reference, I opened this issue: https://github.com/napalm-automation/napalm/issues/284

bordeltabernacle:
Hey @mirceaulinic that all sounds fair enough, I probably should've dug a little deeper into the issues.  I wasn't sure if it was a NAPALM wide issue or an IOS specific one.
I'll see if there's a more suitable place for this info in the docs.
Thanks for getting back to me on this so promptly.

PR #311, opened 2016-10-11
adding get_config to getters support matrix

ebeahan:
Adding entry in getters support matrix for new get_config method https://github.com/napalm-automation/napalm-base/issues/65


dbarrosop:
Thx!

PR #321, opened 2016-11-21
docs: fix cl_napalm_configure example

vincentbernat:
First positional argument is the host, second positional is the configuration file. Not the reverse.


dbarrosop:
Thx!

PR #323, opened 2016-11-26
Added ROS to the supported methods matrix.

inetuid:


dbarrosop:
There is no official release of the driver yet so I will hold onto this one. Thx!

PR #328, opened 2016-12-10
updated imports to support new method network_device

jedelman8:
Adding new method in napalm-base project and allowing to access it by updating this file.
Will cross reference once that PR is open.


ktbyers:
Since this get all put on the back-burner we should probably close it...any objections?

dbarrosop:
agreed

PR #331, opened 2016-12-15
Update index.rst

sjtarik:
just updating the sample code after napalm_base refactoring.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -27,7 +27,7 @@ You can select the driver you need by doing the following:
 
 .. code-block:: python
 
-   >>> from napalm import get_network_driver
+   >>> from napalm_base import get_network_driver
    >>> get_network_driver('eos')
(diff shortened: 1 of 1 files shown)


dbarrosop:
Thx! The existing example should work but I agree it's better to update it.

sjtarik:
Thanks @dbarrosop. With the following list, I was able to get a connection working with devices but example script needed that small change from napalm to napalm_base. Is there a way to get it working with napalm import, without installing all drivers? 

napalm-base (0.20.4)
napalm-ios (0.4.0)
napalm-nxos (0.4.1)

dbarrosop:
No, `from napalm import...` only works if you install all the drivers with `pip install napalm`. However, `from napalm_base import...` *always* works. That's why I think your PR makes complete sense. It's more generic : )

PR #346, opened 2017-01-30
Add explanation of how the _name keyword works

dbarrosop:
Related to napalm-automation/napalm-base#180 and napalm-automation/napalm-base#191.


docs/tags
```diff
@@ -0,0 +1,521 @@
+!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
```
mirceaulinic: Can we remove this file?

I'm not sure where does this come from. I guess it's from your env and you might need to add it in the global .gitignore

dbarrosop: goddammit, yes, that's `ctags`. Sorry about that.

PR #356, opened 2017-03-06
Better install docs

mirceaulinic:
The current installation docs don't provide enough details when it comes to platform dependencies (http://napalm.readthedocs.io/en/latest/installation.html#dependencies)
Also, they depend on the driver the user needs to install. So I thought to separate them by the driver so the user can go straight and see what they need to install for each driver.


dbarrosop:
+2^128

PR #360, opened 2017-03-22
Cleaning up documentation for validator files

fach:
The examples in the documentation for writing validator files is incorrect since landing the following, as yaml dicts were being converted to lists:

https://github.com/napalm-automation/napalm-base/pull/203/commits/b59144b0c92f127230034185e5175c267d92affe

When writing validators now, you must explicitly add each section of the validator as a list entry.


dbarrosop:
Oh, good catch! Thanks!

PR #378, opened 2017-05-31
Doc improvements for opt args

mirceaulinic:


mirceaulinic:
Both napalm-iosxr 0.5  and napalm-junos 0.10 have been released.

PR #393, opened 2017-07-25
Fixed NXOS Version

secahtah:
The version of Cisco NX-OS required to add the feature "nxapi" is actually 7.2 or greater.


dbarrosop:
I think it actually depends on the platform but I am not an expert on cisco equipment so I will defer this to others better suited to review this change in the requirements.

@jedelman8 @mirceaulinic ^

secahtah:
Thanks. The Nexus 5696Q and 7010 platforms both required a minimum of 7.2 in order to add the nxapi feature. I'm not sure about any of the others. Would it be helpful to include a platform matrix?

mirceaulinic:
For 9k series, I think the NX-API is available starting with 7.x, or even a very late 6.x IIRC.
Yeah, a platform matrix sounds sane.

dbarrosop:
Should we just phrase it as "Your device must be running an NXOS version with the API service available which can vary from platform to platform" or something like that instead?

ktbyers:
It also isn't just NX-API support as NX-API originally XML only. 

I believe we require NX-API with JSON-RPC support as that is what the pynxos library is using.

dbarrosop:
Yay! What about:

```
"Your device must be running an NXOS version with the API/JSON-RPC service available which can vary from platform to platform. <sarcasm>We love you cisco</sarcasm>"
```

secahtah:
I modified the readme as suggested above. Does this work? Thanks,

dbarrosop:
@secahtah sorry about this but with the whole reunification process looks like we forgot about this PR and now it's broken :( Would you mind resubmitting? Thanks and sorry again.

PR #423, opened 2017-10-21
Cleanup versioning from drivers and move Python warning

ogenstad:
As mentioned on Slack yesterday, this removes the versioning from the drivers and only uses the core Napalm package.


dbarrosop:
Thanks a lot!

PR #424, opened 2017-10-21
Bring get_network_base up to package level

ktbyers:
Also minor cleanup so pylama passes.

### DIFF
--- a/napalm/__init__.py
+++ b/napalm/__init__.py
@@ -1,7 +1,7 @@
 import pkg_resources
 import sys
-import json
 from napalm.base import get_network_driver
+from napalm._SUPPORTED_DRIVERS import SUPPORTED_DRIVERS
 
 # Verify Python Version that is running
 try:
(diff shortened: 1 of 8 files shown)


napalm/__init__.py
```diff
@@ -1,5 +1,6 @@
 import pkg_resources
 import sys
+from napalm.base import get_network_driver
```
dbarrosop: shouldn't we just move the code?

requirements/dev
```diff
@@ -1,3 +1,4 @@
+future
```
dbarrosop: this is in the `base` requirements already, do you think we need it here too?

ktbyers: Will remove...missed it there.

napalm/__init__.py
ktbyers: It shouldn't matter...napalm.base.__init__ will be processed regardless (as get_network_driver will use code from napalm.base).

PR #476, opened 2017-10-25
Model parsing and hostname on nxos_ssh

ktbyers:
This is a minor fix to improve some issues I saw with NXOS-SSH while testing on the reunification branch.

It doesn't matter if you implement it now or not. I can re-submit it later, if need be.

### DIFF
--- a/napalm/nxos_ssh/nxos_ssh.py
+++ b/napalm/nxos_ssh/nxos_ssh.py
@@ -777,9 +777,9 @@ class NXOSSSHDriver(NetworkDriver):
                 os_version = line.split()[2]
                 os_version = os_version.strip()
 
-            if 'cisco' in line and 'Chassis' in line:
-                _, model = line.split()[:2]
-                model = model.strip()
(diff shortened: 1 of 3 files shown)


napalm/nxos_ssh/nxos_ssh.py
```diff
@@ -791,6 +791,7 @@ def get_facts(self):
                 break
         if hostname.count(".") >= 2:
             fqdn = hostname
+            hostname = hostname.split(".")[0]
```
mschedrin: Hostname itself can contain dots, this code will then fail splitting domain name from hostname. For example fqdn: "my.nexus.spam.com" will give hostname my, while expected result should be "my.nexus". More intelligent parsing should be applied involving getting domain name from the box.

ktbyers: Good point. Will fix.

ktbyers:
Very unimportant...just wait until post release of unified driver.

ktbyers:
Closing as this has been superceded by more recent  PR.

PR #482, opened 2017-10-27
Fix linter issues with iosxr and fix py.test not linting source code

ktbyers:


ktbyers:
I had a big PIA trying to get py.test and pylama to execute together...here is what I did:

In tox.ini:
```
-    py.test test_base
 +    py.test
```

In setup.cfg, I added this to norecursedirs:
```
    napalm/base/test
```

PR #492, opened 2017-11-04
ios: Add get_users implementation and test

nickethier:
This PR adds support for the `get_users` method for cisco ios. Ios doesn't store user public keys, just the fingerprint so that is all that is returned in the `sshkeys` list.


nickethier:
Closing in favor if #495 which is a PR for develop branch

PR #494, opened 2017-11-06
Add Multicast / MDT / VPN IPv4 / VPN IPv6 / L2VPN VPLS support for get_bgp_neighbors

afourmy:
Right now, only IPv4 and IPv6 are listed as supported AFI.
In case the output contains the IPv4 MDT AFI, the function fals because the AFI regex catches 'IPv4 MDT' as 'IPv4' and it is considered a supported AFI (=> duplicated IP)
Fix: change the regex so that it catches the full name of the AFI. 
Add support for more AFI:

        supported_afi = ('ipv4 unicast', 
                         'ipv6 unicast',
                         'ipv4 multicast',
                         'ipv6 multicast',
                         'ipv4 mdt',
                         'l2vpn epn',
                         'l2vpn vpls',
                         'vpnv4 unicast',
                         'vpnv6 unicast')

The strange thing is that the "show bgp ipv4 unicast neighbors" seem to contain all AFI, which means we don't actually need to do "show bgp {AFI} neighbors" for all AFI.

### DIFF
--- a/test/ios/mocked_data/test_get_bgp_neighbors/multicast/expected_result.json
+++ b/test/ios/mocked_data/test_get_bgp_neighbors/multicast/expected_result.json
@@ -1,27 +1,137 @@
-{'global': {'router_id': '192.168.242.162', 'peers': {'192.168.242.152': 
-{'local_as': 65421, 'remote_as': 65421, 'remote_id': '192.168.242.152', 'is_up': True, 
-'is_enabled': True, 'description': '# primary RR1 #', 'uptime': 1, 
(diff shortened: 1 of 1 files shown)


afourmy:
I close this pull request because with the reunification, I had troubles resyncing my pre-reunification fork to the post-reunification napalm with git, so I deleted my fork and this request is now considered from "unknown repository". I think the only way to modify it is to recreate it. sorry 😰

dbarrosop:
No problem, thanks for your patience :)

PR #495, opened 2017-11-06
ios: Add get_users implementation and test

nickethier:
This PR adds support for the get_users method for cisco ios. Ios doesn't store user public keys, just the fingerprint so that is all that is returned in the sshkeys list.


napalm/ios/ios.py
```diff
@@ -1955,6 +1955,54 @@ def get_snmp_information(self):
...
+        The level is an integer between 0 and 15, where 0 is the
+        lowest access and 15 represents full access to the device.
+        """
+        users = {}
+        command = "show run | section username"
+        output = self._send_command(command)
+        curuser = ""
+        for line in output.splitlines():
+            if line.startswith("username"):
+                tokens = line.split(" ")
```
ktbyers: Why not just `line.split()` here it will split on consecutive white space...so probably marginally less fragile.

napalm/ios/ios.py
```diff
@@ -1955,6 +1955,54 @@ def get_snmp_information(self):
...
+                users[tokens[1]] = {
+                    'level': 1,
+                    'password': "",
+                    'sshkeys': []
+                }
+                i = 2
+                while i < len(tokens):
```
ktbyers: I think this parsing will probably work better with regular expressions...I am happy to help write this, but basically cover the patterns of `privilege` and `secret` using regex.

This shifting and incrementing I think is harder to read and easier to make a mistake on.

napalm/ios/ios.py
```diff
@@ -1955,6 +1955,54 @@ def get_snmp_information(self):
...
+                    if tokens[i] == "privilege":
+                        users[tokens[1]]["level"] = int(tokens[i+1])
+                        i += 2
+                    elif tokens[i] == "secret":
+                        users[tokens[1]]["password"] = tokens[i+2]
+                        i += 3
+            elif line.startswith(" "):
```
ktbyers: I think we should do regex here as well and have a newline in the regex pattern (once again I can help with this).

napalm/ios/ios.py
nickethier: I agree. Do you have a good example to point me at?

ktbyers: Here is a first pass at it:

Where first username is:

```
username pyclass privilege 15 secret 5 $1$CQCe$poonxxxzzzttabct/3eXub41
```

```
>>> re.search(r"^username\s+(?P<username>\S+)\s+privilege\s+(?P<priv_level>\S+)\s+secret \d+\s+(?P<pwd_hash>\S+)$", output, re.M).groupdict()
{'username': 'pyclass', 'priv_level': '15', 'pwd_hash': '$1$CQCe$poonxxxzzzttabct/3eXub41'}
```

ktbyers: We would need to account for a few scenarios i.e. when privilege is there or not (if it is password and not secret, et cetera).

napalm/ios/ios.py
```diff
@@ -1955,6 +1955,37 @@ def get_snmp_information(self):
...
+        users = {}
+        command = "show run | section username"
+        output = self._send_command(command)
+        for match in re.finditer(r"^username\s+(?P<username>\S+)\s+(?:privilege\s+(?P<priv_level>\S+)\s+)?(?:secret \d+\s+(?P<pwd_hash>\S+))?$", output, re.M):
+            users[match.groupdict()["username"]] = {
+                'level': int(match.groupdict()["priv_level"]) if match.groupdict()["priv_level"] else 1,
+                'password': match.groupdict()["pwd_hash"] if match.groupdict()["pwd_hash"] else "",
+                'sshkeys': []
+            }
+        for match in re.finditer(r"^\s+username\s+(?P<username>\S+)(?P<keys>(?:\n\s+key-hash\s+(?P<hash_type>\S+)\s+(?P<hash>\S+)(?:\s+\S+)?)+)$", output, re.M):
```
nickethier: @ktbyers I couldn't figure out a way to match the key-hash lines multiple times so this is the best I could come up with.

napalm/ios/ios.py
```diff
@@ -1955,6 +1955,37 @@ def get_snmp_information(self):
...
+
+        *Note: sshkeys on ios is the ssh key fingerprint
+
+        The level is an integer between 0 and 15, where 0 is the
+        lowest access and 15 represents full access to the device.
+        """
+        users = {}
+        command = "show run | section username"
+        output = self._send_command(command)
+        for match in re.finditer(r"^username\s+(?P<username>\S+)\s+(?:privilege\s+(?P<priv_level>\S+)\s+)?(?:secret \d+\s+(?P<pwd_hash>\S+))?$", output, re.M):
```
nickethier: This seems to catch all cases I could think of

nickethier:
I took a stab at cleaning things up with the help of regex. Parsing out the key-hash lines is still a little messy. Let me know if you can think of a better way.

nickethier:
@ktbyers Ready for another look when you get the chance. Thanks!

nickethier:
@mirceaulinic yeah IOS-XE actually only stores the fingerprint which is also strange to me.

Ping @ktbyers for another look.

dbarrosop:
@ktbyers can you review again?

ktbyers:
Will review...it is on my todo list.

PR #496, opened 2017-11-07
IOS ipv6 neighbors table

afourmy:
get_ipv6_neighbors_table for IOS following the format defined in NAPALM doc

3 tests:
- empty table
- normal 
- long


afourmy:
Empty test: fails because of the `assert len(get_ipv6_neighbors_table) > 0` condition in the base getter. Shouldn't this condition be removed, as the IPv6 neighbor table can be empty ?
Normal / long test: not sure why they fail.

afourmy:
OK, I will remove it. For the other tests, I'm less sure what to do to fix them. They work locally for me.

dbarrosop:
Looks like one of the tests the expected result is not what the device is getting? The other test is returning mac address as -1 so the data type check is failing. @mirceaulinic do you know what we should return if the mac address is not available? Should we return `-1` or just something like `00:00:00:00:00`?

afourmy:
Empty test: condtion removed, it's fixed.
Long test: I checked and it's definitely an ordering issue. the result is basically the same, but the elements of the list are not in the same order. When I run the function locally, I get the order which is in the expected_result.json, I don't understand why the CI test would find a different order.

napalm/ios/ios.py
```diff
@@ -2173,6 +2173,52 @@ def get_config(self, retrieve='all'):
...
+                }
+            ]
+        """
+
+        ipv6_neighbors_table = []
+        command = 'show ipv6 neighbors'
+        output = self._send_command(command)
+
+        for entry in output.split('\n')[4:]:
+            # typical format of an entry in the IOS IPv6 neighbors table:
```
ktbyers: Don't split on a fixed location in the file. Do a split() or an re.split() and remove all the header information before the for loop:

Something like:
```
_, neighbor_table = re.split(r^IPv6 Address.*nterface$', output, flags=re.M)
```

Note, the information like:

```
Load for five secs: 9%/5%; one minute: 9%; five minutes: 9%
Time source is NTP, 08:55:30.881 DST Tue Nov 7 2017
```

is not guaranteed to be there (i.e. it depends on your config) and will be stripped off by this method:

```
def _send_command_postprocess(output):
```

This is causing at least part of the issue with the unit tests.

napalm/ios/ios.py
```diff
@@ -2173,6 +2173,52 @@ def get_config(self, retrieve='all'):
...
+            # 2002:FFFF:233::1 0 2894.0fed.be30  REACH Fa3/1/2.233
+            ip, age, mac, state, interface = entry.split()
+            mac = -1 if mac == '-' else napalm.base.helpers.mac(mac)
```
ktbyers: mac will have to be a string for the unknown value and not an integer (or else it will fail the type check for the value). So this should be set to a null string.

afourmy:
Thank you ! I didn't know the first part was stripped by send_command, tests are fixed

PR #497, opened 2017-11-07
IOS get network instances

afourmy:
This is a first shot for get_network_instances for IOS. The expected output doesn't make any sense to me so maybe we can clarify a few things here:

Expected output:
```
    u’MGMT’: {
        u’name’: u’MGMT’, u’type’: u’L3VRF’, u’state’: {
            u’route_distinguisher’: u‘123:456’,
        }, u’interfaces’: {
            u’interface’: {
                u’Management1’: {}
            }
        }
    } u’default’: {
        u’name’: u’default’, u’type’: u’DEFAULT_INSTANCE’, u’state’: {
            u’route_distinguisher’: None,
        }, u’interfaces: {
            u’interface’: {
                u’Ethernet1’: {} u’Ethernet2’: {} u’Ethernet3’: {} u’Ethernet4’: {}
            }
        }
    }
}
```

A few questions:
- the doc says `Return a dictionary of network instances (VRFs) configured, including default/global`: I don't think this can apply to Cisco.
- the name of the VRF is the key of the main dictionnary, why do we need to repeat it in the inner dictionnary ?
- what is expected for the type ? I just put L3VRF everywhere, I don't think this makes sense for Cisco.
- 'state' is a dictionnary that contains a single parameter (route distinguisher)
- interface is a dictionnary containing each interface as a key and a empty dictionnary as a value, why not just a list of interfaces ?

Overall, this function is hard to code for Cisco as-is. I guess we can change it, or maybe have a different function for Cisco VRF (get_vrf) as Juniper and Cisco handle VRF very differently. For Cisco, we would need to know which route-targets are imported/exported, and which route-maps as well, that information is missing right now. This route-target information would be per-AFI, i.e ipv4 unicast, ipv6 unicast, ipv4 multicast, ipv6 multicast.


bewing:
> A few questions:
> - the doc says `Return a dictionary of network instances (VRFs) configured, including default/global`: I don't think this can apply to Cisco.

The main table (no VRF) is the default table.
> - the name of the VRF is the key of the main dictionnary, why do we need to repeat it in the inner dictionnary ?

The return of this function tries to mirror the openconfig network-instances model as closely as possible.
> - what is expected for the type ? I just put L3VRF everywhere, I don't think this makes sense for Cisco.

If you're only dealing with `show ip vrf`, etc, and not pseudowires, VPLS, etc, then L3VRF would be correct
> - 'state' is a dictionnary that contains a single parameter (route distinguisher)
> - interface is a dictionnary containing each interface as a key and a empty dictionnary as a value, why not just a list of interfaces ?

Same as above.

afourmy:
OK Bewing. If the purpose of this function is to mirror the openconfig network-instances YANG, maybe there should be a specific Cisco get_vrf function with more information, and formatted in a pythonic way. (I will do such a function anyway because I need it for my work, but I'm not sure it would make sense to have it in napalm)

I just noticed that I messed up the "interface" part, I forgot an intermediate dictionnary: I'll fix it tomorrow.

bewing:
>  maybe there should be a specific Cisco get_vrf function with more information, and formatted in a pythonic way

That is similar to how it is handled in EOS.  Please note that any driver-specific functions must start with an underscore to avoid failing the method signatures test.

https://github.com/napalm-automation/napalm/blob/develop/napalm/eos/eos.py#L1572-L1630

afourmy:
OK thanks. I will try to fix this one tomorrow with the default VRF and the 'interfaces/interface' dict, and I will make another PR for a _get_vrf where we can discuss about the format of the ouput and the information it should contain.

dbarrosop:
> interface is a dictionary containing each interface as a key and a empty dictionnary as a value, why not just a list of interfaces ?

Because a dictionary can be further extended if we have to.

> maybe there should be a specific Cisco get_vrf function with more information

No, that's not going to be accepted, if there is something that is missing in `get_network_instances` it should be fixed but we are not going to add a vendor specific method unless there is a compelling reason for that. Specially for something as generic as VRFs.

> and formatted in a pythonic way

Not sure what that means. Data is data, it's not tied to a particular programming language.

afourmy:
OK tests are passing, it's ready for reviewing now.

> No, that's not going to be accepted, if there is something that is missing in get_network_instances it should be fixed but we are not going to add a vendor specific method unless there is a compelling reason for that. Specially for something as generic as VRFs.

I think that RT are missing. It could be added to the "state" dictionnary along with the RD, but RT are per AFI, so it could be like 
```'state': {RT: {AFI1: {'import': ..., 'export': ...}, AFI2: .....}```

dbarrosop:
A route target is a BGP community. Cisco and juniper have "macros" to deal with them in very weird and different ways. I'd like @mirceaulinic to bring this up to the openconfig group and see if they have any idea of how this import/export targets are supposed to be represented. I'd prefer to follow their models whenever possible so other tools that talk openconfig can benefit from this.

@mirceaulinic is this something you know or can ask?

dbarrosop:
@afourmy If you are ok with it I'd like to review and merge this independently of the route target discussion (would be nice if you could create an issue to discuss it). I just don't want to block your PR because of it.

afourmy:
Yes sure, you can review it and see if it's working 😊
The fact that Juniper and Cisco manage route-target in very different ways in the reason why I was speaking about a specifc Cisco get_vrf function to deal with RT, but if you can come up with a vendor-agnostic format for RT in get_network_instances, it's even better and I would update the function accordingly.

PR #507, opened 2017-11-10
Adding optional Arguments for EOSDriver function load_replace_candidate

bharath-ravindranath:
Adding optional Arguments for EOSDriver function load_replace_candidate and load_merge_candidate


ktbyers:
@bharath-ravindranath I assume this is just an error? It doesn't look like there are any changes from you?

You need to rebase and submit PR against `develop` branch.

PR #511, opened 2017-11-11
Restore test textfsm_templates symlinks

### LINKED ISSUE #510: textfsm symlinks lost in migration

bewing:
Fixes #510


ktbyers:
I vaguely recall us having a conversation on this issue not too long ago?

I see I did this in the old napalm-nxos...we should just decide how we want it and make it consistent.

dbarrosop:
What are we trying to solve here?

bewing:
In some drivers, `test/<driver>/utils/textfsm_templates` was a symlink to `<module>/<driver>/utils/textfsm_templates`:  https://github.com/napalm-automation/napalm-eos/tree/develop/test/unit/utils

After the unification, it appears this was lost.  This requires one to remember to edit TextFSM templates in both the module and the unit test.  

This PR restores the symlinks.

ktbyers:
The issue now is if the TextFSM files ever get updated (in the ./napalm code)...people will not remember to also update them in the ./test directories.

And you need the TextFSM files in the test directories for the unit tests. I don't really like either solution (copying the files or symlinks, but of the two symlinks is probably better). I wonder if we could just fix the unit tests to point to the real files in ./napalm some
how.

dbarrosop:
Ok, +1 :)

PR #512, opened 2017-11-11
Add vrf paramter to get_arp_table()

bewing:
Update all drivers `get_arp_table` signature to accept a VRF name, and create test.

- [x] Base/testcase
- [x] EOS
- [x] EOS mocked data
- [x] IOS 
- [ ] IOS mocked data
- [ ] IOS-XR 
- [ ] IOS-XR mocked data
- [x] JunOS
- [ ] JunOS mocked data
- [ ] NX-OS
- [ ] NX-OS mocked data
- [ ] NX-OS (SSH) 
- [ ] NX-OS (SSH) mocked data


afourmy:
in the possible implementation for IOS, the vrf keyword is missing before the VRF name. I don't see any differences in the normal and VRF format on IOS, so if you remove the NotImplementedError exception, it should work (haven't tried though)

napalm/nxos/nxos.py
```diff
@@ -595,7 +595,9 @@ def cli(self, commands):
             cli_output[py23_compat.text_type(command)] = command_output
         return cli_output
 
-    def get_arp_table(self):
+    def get_arp_table(self, vrf=c.ARP_VRF):
+        if vrf:
```
dbarrosop: We should probably use a helper method for this. Something like:

```
# in something something helpers.py
def raise_if_not_supported(supported, args):
    "if an argument in args is not in supported raise NotImplementedError" 
   for kw in args:
       if kw not in supported:
           raise NotImplementedError("{} not implemented".format(kw))

# in drivers
def get_arp_table(self, **kwargs):
    helpers.raise_if_not_supported(['vrf'], kwargs.keys())
```

With this pattern we can start migrating all the existent methods to it so when we implement a new arg we don't have to care about it. We could also add an optional arg "ignore_not_implemented_kwargs" or something like that avoid raising a NotImplementedError if the user passes a kwargs that is not supported.

I don't like it, btw, but would simplify implementing new kwargs and even vendor specific options like commit messages.

dbarrosop:
We should discuss the behavior for this. If `vrf` is not passed, are we returning everything or just the default/global vrf?

ktbyers:
I think the default behavior should remain unchanged so it should default to vrf='global/default'. Probably should be some way to return all VRFs, however.

bewing:
"Implementation is left up to the reader" ?

If they want all ARP in all VRF, they can loop get_network_instance.keys() through get_arp_table()

mirceaulinic:
Yes, I agree with @ktbyers - the behaviour should remain unchanged.

dbarrosop:
Fine by me. I was mostly asking because the default behavior is not documented anywhere.

bewing:
Would appreciate some assistance during the hackathon with implementations for remaining core drivers, and example output for test cases!

dbarrosop:
We should probably use a branch for things like this so people can submit independent PRs against it and merge the branch against develop only when everything has been fixed.

bewing:
Well, it is a branch off develop, so do we want to close this PR and create an issue instructing PRs against it?

dbarrosop:
yes, let's start use feature branches when we need to synch multiple PRs as it's this case.

dbarrosop:
what's the status of this one?

mirceaulinic:
@ckishimo implemented for Junos too.

tardoe:
What's left to do on this one? Are we still pending the mocked data for all but EOS and NX-OS/XXR implementations?

ktbyers:
I think I want to rip out that constant (default argument) which is just defined as a null string. I think that is a bit confusing/unnecessary (i.e. we should just move the default value to the method definition and eliminate the constant with a default value of either null string or None).

Someone mentioned on the get_interfaces_ip (adding VRF support) that we might need to review napalm-validate to ensure we don't break something there. So that needs looked into (i.e. that adding this vrf argument doesn't cause an issue with napalm validate in some way).

Looks like there is also an issue on the Travis tests with the auto-build of the getters table that needs resolved.

Besides that--as long as we can do it in a backwards compatible way, I think we should just release it (even if it is only supported on some of the core platforms).

In other words, core platforms that don't support it will just raise NotImplementedError (if someone tries to pass in a VRF and the getter for that platform hasn't added support for it).

We also need a way to update the getter support matrix to indicate VRF support in some way.

ktbyers:
Structure was added here:

https://github.com/napalm-automation/napalm/pull/900

I am going to close this as moving forward it will be easier to just incrementally add this support platform-by-platform. Also there isn't really any active work that is happening directly on this PR.

PR #515, opened 2017-11-13
Adding base in _SUPPORTED_DRIVERS and renaming functions

GGabriele:
The reason to add `base` to `_SUPPORTED_DRIVERS` is to be able to install napalm without installing dependencies. The changes in the validate methods are to better reflect which one is supposed to be "public" and which one not.


dbarrosop:
+1

Issue #490, opened 2017-11-02
Add MTU to interfaces coming from get_interfaces


Ragsboss:
### Description of Issue/Question
MTU is becoming an important piece of information on interfaces especially with abundance of overlays these days, which is putting some specific requirements on MTU on the underlay. I see that this is not part of the interface model documented to be output from `get_interfaces` method. This issue is an enhancement request to extend the model to include MTU and all the drivers to start fetching MTU. In most cases, the commands drivers are using already have MTU in the output.

### Did you follow the steps from https://github.com/napalm-automation/napalm#faq
[X] Yes
[ ] No

### napalm version
(Paste verbatim output from `pip freeze | grep napalm` between quotes below)

```
You are using pip version 8.1.2, however version 9.0.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
napalm==1.2.0
napalm-base==0.25.0
napalm-eos==0.6.1
napalm-fortios==0.4.0
napalm-ios==0.8.1
napalm-iosxr==0.5.6
napalm-junos==0.12.1
napalm-nxos==0.7.1
napalm-panos==0.4.0
napalm-pluribus==0.5.1
napalm-ros==0.2.2
napalm-vyos==0.1.3
```

### Network device software version
N/A

### Steps to Reproduce the Issue
N/A

### Error Traceback
N/A

### Trace captured
N/A

dbarrosop:
Yeah, I think it makes sense. We should probably add it to the output of `get_interfaces`. @mirceaulinic @ktbyers thoughts?

@Ragsboss is this something you'd want to work on?

targuan:
I'm working on the implementation for junos driver. I don't have any junos device, I'm working with the test output.
Some logical interfaces don't report the mtu (for instance ge-0/0/0.0 in [this test cast](https://github.com/napalm-automation/napalm/blob/develop/test/junos/mocked_data/test_get_interfaces/normal/get-interface-information.xml#L209).
 Anybody know how this should be reported by the driver ? With a special value ? Inherit the value from the physical interface ? Inherit the value from the physical interface with a delta to report the encapsulation ?

afourmy:
@targuan it is already done for junos, see https://github.com/napalm-automation/napalm/compare/get_interfaces_mtu3

targuan:
@afourmy the test of get_interfaces has not been updated with the mtu values and some interfaces where added inbetween in the test set and are not handled.

ktbyers:
[...]

PR #516, opened 2017-11-14
Junos: add mtu in get_interfaces()

ckishimo:
Support for mtu in Junos `get_interfaces()`. Issue #490 
The mocked data needs to be fixed, but I'd like to know first if it's fine to keep mtu as `str` as there are interfaces in `Junos` that return `Unlimited` or if it needs to be `int` and then change the `Unlimited` to `-1` ?
```
    "lo0": {
        "description": "",
        "is_enabled": true,
        "is_up": true,
        "last_flapped": -1.0,
        "mac_address": "Unspecified",
        "mtu": "Unlimited",
        "speed": -1
    },
```
Also which is the default value in case `mtu` is not available `0` or `-1` ? Thanks
```
    "lc-0/0/0": {
        "description": "",
        "is_enabled": true,
        "is_up": true,
        "last_flapped": -1.0,
        "mac_address": "Unspecified",
        "mtu": "0",
        "speed": 800
    },
```


dbarrosop:
Unfortunately we have to update the rest of the core drivers and base. Is this something you can do? Existing tests/mocked data should be enough so you don't even need access to real devices.

ktbyers:
I committed this to a feature branch.

PR #522, opened 2017-11-15
base: helpers: Added jinja_filters arg to load_template

nickethier:
This is a bit more of a discussion with working code, but happy to have it merged if it checks out.

In the ios driver I want to implement the set_users template, but I would need to either compute each ssh key's fingerprint, or wrap the text since ios doesn't allow any lines longer than 254. Adding a filter to the base class seemed like the wrong solution, so I came up with this. This way the ios driver could override `base.load_template` and inject the filters specific to the driver.


napalm/base/helpers.py
```diff
@@ -61,8 +61,9 @@ def load_template(cls, template_name, template_source=None, template_path=None,
             loader = jinja2.FileSystemLoader(search_path)
             environment = jinja2.Environment(loader=loader)
 
-            for filter_name, filter_function in CustomJinjaFilters.filters().items():
-                environment.filters[filter_name] = filter_function
+            for d in [CustomJinjaFilters.filters(), jinja_filters]:
```
bewing: Maybe `for filter_name, filter_function in itertools.chain(CustonJinjaFilters.filters().items(), jinja_filters.items()):` ?

nickethier: Ah I see you've uncovered my lack of python experience!

nickethier:
Anything else you folks need from me to get this merged? @ktbyers @mirceaulinic 
I've about got a PR ready to make use of this if you'd like to see that first.

dbarrosop:
LGTM as well :)

PR #526, opened 2017-11-16
allow custom napalm methods

itdependsnetworks:
This intent based software defined class discovery will rock your socks off! It slices it dices, it uses 40% less energy then previous generations!!!

As discussed before, my preference to this method is it will allow me to integrate custom methods easier as I go from customer to customer. This is unobtrusive method, that I think should allow for easy custom methods to those in the know, and those not, will not be affected.


itdependsnetworks:
I assume you mean: 

```
driver = get_network_driver('ios_custom')
```

Which should be natively supported, and thus not needed? I hope not, but trying to understand what you are suggesting.

dbarrosop:
I understand that what you are trying to do is `get_network_driver('ios')` and get your custom driver if available or the official one if not. I think that's fine as per previous discussions but whatever fix we implement should modify *only* the `get_network_driver` function, nothing else. Your current implementation is doing weird stuff with imports that might have unexpected side effects.

itdependsnetworks:
What would be the non-weird way of doing this?

dbarrosop:
Actually, checking at your changes, you would just have to remove the imports which I am not entirely sure why you need them.

napalm/ios/__init__.py
```diff
@@ -13,7 +13,10 @@
 # the License.
 
 """napalm.ios package."""
-from napalm.ios.ios import IOSDriver
+try:
```
dbarrosop: Why is this needed?

napalm/ios/ios.py
```diff
@@ -25,7 +25,10 @@
 import copy
 
 from netmiko import ConnectHandler, FileTransfer, InLineTransfer
-from napalm.base.base import NetworkDriver
+try:
```
dbarrosop: Why is this needed?

itdependsnetworks:
As always, thanks to @dbarrosop for his help. We agreed that the by product of not having a `NotImplementedError` raised was reasonable. You color outside the lines, you have to manage.

PR #528, opened 2017-11-17
OS version on Cisco IOS

### LINKED ISSUE #469: get_facts returns the whole string from "show version" not just the software version
When using the generic NAPALM function `get_facts` against an `ios` device, the whole string from the command "show version" is returned, not just the software version (like a `junos` device does).

- [X] Yes
- [ ] No

### napalm-ios version
(Paste verbatim output from `pip freeze | grep napalm-ios` between quotes below)

```
$ pip freeze | grep napalm-ios
napalm-ios==0.7.0
```

### IOS version
(Paste verbatim output from `show version` between quotes below)

```
#sh ver
Cisco IOS Software, s72033_rp Software (s72033_rp-ADVIPSERVICESK9-M), Version 15.1(2)SY9, RELEASE SOFTWARE (fc1)
Technical Support: http://www.cisco.com/techsupport
Copyright (c) 1986-2016 by Cisco Systems, Inc.
Compiled Thu 29-Sep-16 04:07 by prod_rel_team

ROM: System Bootstrap, Version 12.2(17r)SX5, RELEASE SOFTWARE (fc1)
BOOTLDR: Cisco IOS Software, s72033_rp Software (s72033_rp-ADVIPSERVICESK9-M), Version 15.1(2)SY9, RELEASE SOFTWARE (fc1)

 bld-vsc-a uptime is 20 weeks, 3 days, 6 hours, 43 minutes
Uptime for this control processor is 20 weeks, 3 days, 6 hours, 43 minutes
System returned to ROM by reload at 15:17:00 UTC Fri Mar 17 2017 (SP by reload)
System restarted at 15:20:16 UTC Fri Mar 17 2017
System image file is "sup-bootdisk:s72033-advipservicesk9-mz.151-2.SY9.bin"
Last reload reason: Reload Command



This product contains cryptographic features and is subject to United
States and local country laws governing import, export, transfer and
use. Delivery of Cisco cryptographic products does not imply
third-party authority to import, export, distribute or use encryption.
Importers, exporters, distributors and users are responsible for
compliance with U.S. and local country laws. By using this product you
agree to comply with applicable laws and regulations. If you are unable
to comply with U.S. and local laws, return this product immediately.

A summary of U.S. laws governing Cisco cryptographic products may be found at:
http://www.cisco.com/wwl/export/crypto/tool/stqrg.html

If you require further assistance please contact us by sending email to
export@cisco.com.

cisco WS-C6509-E (R7000) processor (revision 1.2) with 983008K/65536K bytes of memory.
Processor board ID SMG1027NQ4V
SR71000 CPU at 600Mhz, Implementation 0x504, Rev 1.2, 512KB L2 Cache
Last reset from s/w reset
27 Virtual Ethernet interfaces
51 Gigabit Ethernet interfaces
18 Ten Gigabit Ethernet interfaces
1917K bytes of non-volatile configuration memory.

65536K bytes of Flash internal SIMM (Sector size 512K).
Configuration register is 0x2102
```

When executing the `get_facts` function against the device above, the `os_version` variable that is returned contains the string `s72033_rp Software (s72033_rp-ADVIPSERVICESK9-M), Version 15.1(2)SY9, RELEASE SOFTWARE (fc1)"` and not just `15.1(2)SY9` (similar to the way a `junos` device returns something like `12.3R7.7`).

### Error Traceback
N/A

FYI, @ktbyers requested this bug to be opened.

---

Issue moved from https://github.com/napalm-automation/napalm-ios/issues/184

dmcarmo:
In order to fix #469 I added a regex pattern that matches the IOS naming convention. First pull request so I hope everything is ok with the process.


ktbyers:
I am not sure we should change this. See more details here:

https://github.com/napalm-automation/napalm-ios/issues/184

@dbarrosop proposed using the file name, but the file name can be any arbitrary name so it doesn't necessarily tell you the OS version.

I think we need to retain these two pieces of information: ADVIPSERVICESK9-M), Version 15.1(2)SY9

I think we should just leave it to what it currently is. Someone can then post process it for what they want.

dmcarmo:
I think the problem is the consistency of possible values for this (as @dbarrosop mentioned before). Should we discuss/develop this issue or just leave it as it is and close it?

mirceaulinic:
We have discussed previously regarding the possibility of having per-platform methods. What would you think about a little helper such as `ios_version` inside the IOS class, or simply inside the `napalm.base.helper` that users can call to parse the version further?

ktbyers:
I think we should just leave it as-is. 

It is good enough and people can then further parse the string to extract the information they want. 

The problem is that there is other useful information in the string like `advipservicesk9`.

```
Cisco IOS Software, s72033_rp Software (s72033_rp-ADVIPSERVICESK9-M), Version 15.1(2)SY9, RELEASE SOFTWARE (fc1)
```

There are other nuances in IOS-XE. 

Also Cisco IOS changed their licensing at a certain point so there will likely be screen-scraping issues handling that also.

I think it is going to be problematic to figure out what exactly to parse and retain and to present.

In other words by trying to further parse the string we are going to create a lot of screen-scraping problems...so I think we should not do this.

dbarrosop:
I agree with ktbyers.

ktbyers:
Closer...napalm, library users can parse further if they want more delimited os version information.

PR #529, opened 2017-11-17
add mtu to get_interfaces for IOS

afourmy:
add mtu to get_interfaces


afourmy:
fixing the tests now

afourmy:
OK so I just understood that the tests for this PR cannot pass. It was failing because of `assert helpers.test_model(models.interface, interface_data)`, so I updated the keys in base/test/models.py.
Now all the other drivers are failing for the same reason.

ktbyers:
@afourmy If get_interfaces is going to return the MTU, then it has to return the MTU for all of the core platforms (IOS, NX-OS, EOS, IOS-XR, Junos).

That is a requirement of napalm (i.e. the data structure returned is the same across all platforms).

afourmy:
Yes, it was already done for Junos so I did it for IOS, but I understand that it will need to be updated for all drivers. Can it be done in separate PR that will be merged all at once, or does everything need to be in one PR ?

ktbyers:
@afourmy How about we do this, I will create a new branch `get_interfaces_mtu`. You submit your PR against this branch.

That way we can incorporate the PRs into this branch platform-by-platform. Once all of the platforms are done, we can submit a PR from this branch to `develop`.

Does that sound reasonable?

afourmy:
Yes sounds good, thank you ! @ckishimo I guess you'll have to do the same things for Junos i.e resubmit https://github.com/napalm-automation/napalm/pull/516.

ktbyers:
@afourmy @ckishimo I will see if I can do them both (potentially).

ktbyers:
Unified into a new branch.

PR #530, opened 2017-11-18
Get interfaces mtu2

ktbyers:
Placeholder PR for MTU update includes Junos and IOS.


ktbyers:
IOS PR is from @afourmy reference here: https://github.com/napalm-automation/napalm/pull/529

ktbyers:
Made an error...need to redo it.

dbarrosop:
We should probably use feature branches for things like this.

PR #518, opened 2017-11-14
Junos: Support interface units in get_interfaces()

ckishimo:
Support for interface units in Junos `get_interfaces()` only. Issue #431 
Based on the Arista example, we have to include the subinterfaces which will inherit most of the information from the physical interface except the `description` and the `is_enabled` fields. Can somebody clarify if that is enough or is there any other thing that needs to be done ? Thanks

```
    "ge-0/0/8": {
        "description": "", 
        "is_enabled": true, 
        "is_up": true, 
        "last_flapped": 12106.0, 
        "mac_address": "00:05:86:71:F9:08", 
        "speed": 1000
    }, 
    "ge-0/0/8.32767": {
        "description": "", 
        "is_enabled": true, 
        "is_up": true, 
        "last_flapped": 12106.0, 
        "mac_address": "00:05:86:71:F9:08", 
        "speed": 1000
    }, 
    "ge-0/0/9": {
        "description": "", 
        "is_enabled": true, 
        "is_up": true, 
        "last_flapped": 12106.0, 
        "mac_address": "00:05:86:71:F9:09", 
        "speed": 1000
    }, 
    "ge-0/0/9.0": {
        "description": "test test", 
        "is_enabled": true, 
        "is_up": true, 
        "last_flapped": 12106.0, 
        "mac_address": "00:05:86:71:F9:09", 
        "speed": 1000
    }, 
```

PR #531, opened 2017-11-18
Placeholder for MTU change (in separate branch)

ktbyers:
Contains Junos and IOS change.


ktbyers:
IOS change from @afourmy https://github.com/napalm-automation/napalm/pull/529 was integrated in.

ktbyers:
The IOS part of this change is good.

Junos part needs reviewed.

Still need nxos, nxos_ssh, eos, and iosxr.

dbarrosop:
#539 contains the nxos one

dbarrosop:
We should probably use feature branches for things like this so everybody can work independently and only merge the feature branch into develop when it's fully implemented.

afourmy:
@dbarrosop There is a branch called "get_interface_mtu3" (see https://github.com/napalm-automation/napalm/pull/529). That's where I pushed https://github.com/napalm-automation/napalm/pull/539

dbarrosop:
Ah, my bad, thanks for the clarification :)

mirceaulinic:
FYI, I only resolved the conflict here. We might need to update the mock data, due to recent changes in the output structure of `get_interfaces` (which contains the logical interfaces too): https://github.com/napalm-automation/napalm/pull/518

mirceaulinic:
Just checking in: besides IOS-XR and NXOS is there anything else to be sorted here?

ktbyers:
Someone probably needs to own coordinating the whole task i.e. making sure all the code is good, that we have proper unit tests for it, finishing any missing drivers, and releasing it.

napalm/junos/junos.py
```diff
@@ -293,9 +293,12 @@ def _convert_to_dict(interfaces):
                         napalm.base.helpers.mac,
                         interfaces[iface]['mac_address'],
                         py23_compat.text_type(interfaces[iface]['mac_address'])),
-                    'speed': -1
+                    'speed': -1,
+                    'mtu': -1
                 }
                 # result[iface]['last_flapped'] = float(result[iface]['last_flapped'])
+            if interfaces[iface]['mtu'] != 'Unlimited':
```
dbarrosop: I think this is missing indentation as it seems to be outside the previous for loop.

ktbyers:
Closing...don't think it makes sense to keep this open any longer.

PR #533, opened 2017-11-18
Fix dictionary update behavior for canonical name / abbreviated name

ktbyers:
Also change behavior of abbreviated_interface so that name lookup attempt occurs regardless of failure of canonical name lookup.

Some minor name changes.


itdependsnetworks:
sounds good.

PR #534, opened 2017-11-18
Fixing NXOS file write issue Python3

ktbyers:
Using Python3 nxos API will fail on an replace config operation due to a failure writing a temporary file.


ktbyers:
Tested on both Py2.7 and Py3.6.

dbarrosop (approved):

PR #535, opened 2017-11-18
Add code to build documentation for napalm-ansible

ogenstad:
This will add documentation for napalm-ansible modules.

It's related to this issue for the Hackathon - https://github.com/napalm-automation/hackathon-2017/issues/10 and depends on this PR getting merged https://github.com/napalm-automation/napalm-ansible/pull/106

Currently there's no script to git clone the napalm-ansible modules and will currently fail.

Up for discussion is where in the documentation tree to put these things.

Current problems
* The tables in the output aren't working and needs formatting.
* Subheaders for each of the modules are included in the top tables of contents.


ktbyers:
I don't have that much preference here (for where to put the documents). 

What about making a napalm-ansible section and then do something similar to `napalm-yang` or `napalm-logs` -- will this be the structure?

How about put it after `napalm-logs`.

dbarrosop:
We can probably just add an "integrations" sections where we can document things like ansible, salt, st2, etc...

ogenstad:
Ok så in that case I'm thinking

/integrations/ansible/modules/napalm_cli
/integrations/ansible/modules/napalm_get_facts
etc.

That way we can also have /integrations/ansible/installation, and I saw some PR in napalm-ansible about adding action plugins. The above structure would make it easy to add that without having to change the urls. The downside is that it might feel like the modules are a bit deep down in the tree before other content is added.

dbarrosop:
nice! :D

ogenstad:
Thanks @dbarrosop!

This one is ready for review now. Later it wouldn't hurt if the actual text in the documentation was improved a bit and more verifications can be done in the napalm-ansible repo. But this should build correctly now.

Issue #456, opened 2017-10-24
device.get_lldp_neighbors() does not support spaces in Device ID


mirceaulinic:
### Description of Issue/Question
device.get_lldp_neighbors() does not support spaces in Device ID 

### Did you follow the steps from https://github.com/napalm-automation/napalm#faq
- [x] Yes
- [ ] No

### napalm-ios version
(Paste verbatim output from `pip freeze | grep napalm-ios` between quotes below)

```
napalm-ios==0.6.1
```

### IOS version
(Paste verbatim output from `show version` between quotes below)

```
Version 12.2(53r)SE
```

### Steps to Reproduce the Issue
device.get_lldp_neighbors() does not support spaces in Device ID 
`device_id, local_int_brief, hold_time, capability, remote_port = lldp_entry.split()`
will result in 
`except ValueError:`
with a value of 
`len(lldp_entry.split()) > 5`
but this is not expected so it will pass 
`if len(lldp_entry.split()) == 4:`
move onto 
`local_port = self._expand_interface_name(local_int_brief)`
which local_int_brief is never defined and result in
`Traceback (most recent call last):
  File "hello.py", line 57, in <module>
    print (local_int_brief)
NameError: name 'local_int_brief' is not defined`

Example data:
```
Device ID           Local Intf     Hold-time  Capability      Port ID
Polycom VVX 500     Gi2/0/36       120        T               0004.f263.5b3e
```

Example out of lldp_entry.split():
`['Polycom', 'VVX', '500', 'Gi2/0/36', '120', 'T', '0004.f263.5b3e']`

### Error Traceback
(Paste the complete traceback of the exception between quotes below)

```
Traceback (most recent call last):
  File "hello.py", line 57, in <module>
    print (local_int_brief)
NameError: name 'local_int_brief' is not defined
```

---

Issue moved from https://github.com/napalm-automation/napalm-ios/issues/141

bewing:
Resolved in #538

PR #538, opened 2017-11-18
Handle multiple spaces in device_id from show lldp neighbors

chadell:
Fixing issue #456


dbarrosop:
Would you mind adding a test case for this?

https://github.com/napalm-automation/napalm/tree/develop/test/ios/mocked_data/test_get_lldp_neighbors

chadell:
Ok, do you prefer a new alternate case, or just adapt this https://github.com/napalm-automation/napalm/blob/136c335fbf7089393e6438002dfc84862cf930fe/test/ios/mocked_data/test_get_lldp_neighbors_detail/alternate2/show_lldp_neighbors.txt#L12 to make it consistent with the changes?

dbarrosop:
always a new test case, old ones should be kept as they are to ensure we have no regressions :)

PR #539, opened 2017-11-18
NXOS add mtu get interfaces

afourmy:
Add MTU to get_interface for the NXOS-SSH driver.


afourmy:
```
test/nxos_ssh/test_getters.py::TestGetter::test_get_interfaces[5548_7_0_8_N1_1] PASSED
test/nxos_ssh/test_getters.py::TestGetter::test_get_interfaces[alternate1] PASSED
test/nxos_ssh/test_getters.py::TestGetter::test_get_interfaces[normal] PASSED
```
=> Tests are OK

The overall build fails because I updated the data model in the IOS PR.
However, I saw that for junos, it fails because of the type

```
test/junos/test_getters.py::TestGetter::test_get_interfaces[normal] <- napalm/base/test/getters.py key: mtu
model_class: <class 'int'>
data_class: <class 'str'>
FAILED
```

I think the MTU type should be updated to int in the junos PR

PR #540, opened 2017-11-19
Pre decorator commit

itdependsnetworks:
ok, this is a mess, but it's a start. 

I had troubles making it a decorator, so I wanted to get something on paper to agree on, and perhaps some pointers.

### DIFF
--- a/napalm/base/base.py
+++ b/napalm/base/base.py
@@ -1590,3 +1590,23 @@ class NetworkDriver(object):
             return napalm.base.helpers.canonical_interface_name(interface, update_os_mapping=None)
         else:
             return interface
+
+#    def log_command_output(self, func):
+#        """Generic method to log commands"""
+#        def wrap_function(self, *args, **kwargs):
+#            output = func(*args, **kwargs)
+#
+#            if optional_args.get('save_commands'):
+#                self.save_commands_directory = save_command_directory()
+#
+#            file_path = self.write_log_file
+#            file_path = file_path.format(commands="sho_run")
+#            folder_path = os.path.dirname(os.path.abspath(file_path))
+#            if not os.path.exists(folder_path):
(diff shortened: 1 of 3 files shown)


dbarrosop:
What is it you are trying to achieve here?

itdependsnetworks:
https://github.com/napalm-automation/hackathon-2017/issues/4

dbarrosop:
Ok, so if we want to go with the recorder/replayer I think we should continue the work in napalm-automation/napalm-base/pull/296.

If you just want to log for posterity what's going on do instead:

```
import logging

log_backend = logging.getLogger("napalm-backend")

...

    def __send_command(self, command):
    ...
        output = __send_command(self, command)
        log_backend.debug(command)    
```

Way simpler and cleaner and then you can just enable it and configure it by just configuring the logger.

itdependsnetworks:
Logging in that matter does not accomplish my goals. See my comment in PR 296.

Reading the prior implementation PR, it is unclear to the reader (or at least me) what was discussed on slack and what the context of that convo was. Picking up this work seems like a much larger proposition, when in reality I was only selfishly concerned with the recording feature of it.

dbarrosop:
well, if you want the recording capability we should continue the work previously started. This solution here is very adhoc and will probably lead to mistakes and inconsistencies.

itdependsnetworks:
@dbarrosop 

I don't know what was decided, so there is nothing for me to continue.

dbarrosop:
I think we should resend the recording/replayer PR to `napalm` and then take it from there. I will try to do that this week and ping you so you can comment and provide feedback/ideas to make sure it meets your needs.

dbarrosop:
Closing, as usual feel free to comment/reopen if you don't agree with my decision or if you have something else to discuss.

PR #541, opened 2017-11-19
Int map tests

itdependsnetworks:
This builds on @ktbyers PR, add tests, and fixes a few things. 

please note the change to abbreviated.


napalm/base/helpers.py
```diff
@@ -258,52 +258,57 @@ def as_number(as_number_val):
...
-    '''
-    Function to retun interface canonical name
+def split_interface(intf_name):
+    """Split an interface name based on first digit, slash, or space match."""
+    head = intf_name.rstrip(r'/\0123456789. ')
+    tail = intf_name[len(head):].lstrip()
+    return (head, tail)
+
+
+def canonical_interface_name(interface, addl_name_map=None):
```
dbarrosop: What is `addl_name_map`?

itdependsnetworks: A map to add or override in case there is vendor specific info

napalm/base/helpers.py
```diff
@@ -258,52 +258,57 @@ def as_number(as_number_val):
...
+    """Function to return an interface's canonical name (fully expanded name).
+
     This puposely does not use regex, or first X characters, to ensure
```
dbarrosop: You do understand this statement is completely false, don't you? Both using regex and `iface_name[0:2]` works perfectly fine in the example you are giving.

napalm/base/helpers.py
```diff
@@ -258,52 +258,57 @@ def as_number(as_number_val):
...
 
 
-def abbreviated_interface_name(interface, update_os_mapping=None):
-    '''
-    Function to retun interface canonical name
-    This puposely does not use regex, or first X characters, to ensure
-    there is no chance for false positives. As an example, Po = PortChannel, and
-    PO = POS. With either character or regex, that would produce a false positive.
-    '''
+def abbreviated_interface_name(interface, addl_name_map=None, addl_reverse_map=None):
```
dbarrosop: Same here, what are `addl_name_map` and `addl_reverse_map`? Format docstring properly and add it to the documentation with autodoc, please. We probably need a new section for helpers.

napalm/base/helpers.py
dbarrosop: My points with the previous comment are:

1. I want to make sure we are doing the right thing for the right reason. And that comment only makes me think the reason is not correct (maybe we are still doing the right thing though, not arguing that).
2. If the statement is false we should remove it.

napalm/base/helpers.py
dbarrosop: Ok, I added a comment on the other method but it'd be great to format the docstring properly explaining the arguments and adding it to the documentation with autodoc.

napalm/base/helpers.py
itdependsnetworks: It will be a short time until you add case insensitivity to that statement which is what I was hinting at, but not explicitly stating. 

I cannot change your mind, but again, I have been normalizing interfaces for a while, and my regex attempts failed, and the cpan found &harmonizeIntsnhave not.

napalm/base/helpers.py
itdependsnetworks: Will do

napalm/base/helpers.py
dbarrosop: > It will be a short time until you add case insensitivity to that statement which is what I was hinting at, but not explicitly stating.

What does this mean? regexes are case sensitive unless you specifically disable it

> I cannot change your mind, but again, I have been normalizing interfaces for a while, and my regex attempts failed, and the cpan found &harmonizeIntsnhave not.

It's not about changing my mind, I am not arguing about the implementation, I personally don't mind and I am not asking you to do it anyway. I just want to make sure we do not confusing people making false claims.

dbarrosop: Again, not trying to change the implementation, I just think we should remove that statement from the docstring: https://regex101.com/r/gAmu1N/1

itdependsnetworks: it’s in the doc string since it will come up again, over and over and it usually is easier to explain to future folks why not to do it to beat them to it with an explanation.

itdependsnetworks: I will make a more complete example

ktbyers:
@itdependsnetworks A bunch of pylama/lint issues that need fixed. The additional argument on `abbreviated` is fine.

I think we should merge this after the lint issues are fixed (as I think the only thing missing was docstrings) and we can add those in a separate PR.

itdependsnetworks:
Saw your note after I updated doc string, hopefully good to go.

napalm/base/helpers.py
```diff
@@ -258,52 +258,74 @@ def as_number(as_number_val):
...
-        return reverse_mapping[long_int] + str(interface_number)
-    # if nothing matched, at least return the original
+    name_map = {}
+    name_map.update(base_interfaces)
+    interface_type, interface_number = split_interface(interface)
+
+    if isinstance(addl_name_map, dict):
+        name_map.update(addl_name_map)
+    if isinstance(addl_reverse_map, dict):
+        reverse_mapping.update(addl_reverse_map)
```
ktbyers: @itdependsnetworks You can't update `reverse_mapping` as that is modifying the original dictionary and will potentially have unintended behavior if multiple calls are made.

You will need to do something like:

```
name_map_reverse = {}
name_map_reverse.update(reverse_mapping)

# Then...
    if isinstance(addl_reverse_map, dict):
         name_map_reverse.update(addl_reverse_map)

# And update additional references to use name_map_reverse
```

napalm/base/helpers.py
```diff
@@ -258,52 +258,79 @@ def as_number(as_number_val):
...
+    interface_type, interface_number = split_interface(interface)
+
+    if isinstance(addl_name_map, dict):
+        name_map.update(addl_name_map)
     # check in dict for mapping
-    if base_interfaces.get(interface_type):
-        long_int = base_interfaces.get(interface_type)
+    if name_map.get(interface_type):
+        long_int = name_map.get(interface_type)
         return long_int + str(interface_number)
```
ktbyers: Can't type cast as a `str` (due to PY2 versus PY3 differences).

Needs to be:

```
return long_int + py23_compat.text_type(interface_number)
```

napalm/base/helpers.py
```diff
@@ -258,52 +258,79 @@ def as_number(as_number_val):
...
+
+    # Try to ensure canonical type.
+    if name_map.get(interface_type):
+        canonical_type = name_map.get(interface_type)
     else:
-        return interface
+        canonical_type = interface_type
+
+    try:
+        abbreviated_name = rev_name_map[canonical_type] + str(interface_number)
```
ktbyers: Same issue can't type cast as `str`.

```
py23_compat.text_type(interface_number)
```

test/base/test_helpers.py
```diff
@@ -388,6 +388,46 @@ def test_convert_uptime_string_seconds(self):
...
+                         "TenGigabitEthernet1/1/1.5")
+        self.assertEqual(napalm.base.helpers.canonical_interface_name('Gi1/2'),
+                         "GigabitEthernet1/2")
+        self.assertEqual(napalm.base.helpers.canonical_interface_name('HundredGigE105/1/1'),
+                         "HundredGigabitEthernet105/1/1")
+        self.assertEqual(napalm.base.helpers.canonical_interface_name('Lo0'), "Loopback0")
+        self.assertEqual(napalm.base.helpers.canonical_interface_name('lo0'), "Loopback0")
+        self.assertEqual(napalm.base.helpers.canonical_interface_name('no_match0/1'),
+                         "no_match0/1")
+        self.assertEqual(napalm.base.helpers.canonical_interface_name('lo',
```
ktbyers: I would probably make this test be `lo1/0/1` and have it return `something_custom1/0/1`...wouldn't that be more consistent with the actual usage pattern?

test/base/test_helpers.py
```diff
@@ -388,6 +388,46 @@ def test_convert_uptime_string_seconds(self):
...
+                         addl_name_map={"lo": "something_custom"}), "something_custom")
+        self.assertEqual(napalm.base.helpers.canonical_interface_name('uniq',
```
ktbyers: Same should be `uniq0` and return `something_custom0`. I think this is more consistent with usage pattern.

PR #544, opened 2017-11-19
DON'T MERGE: Add framework to support commit confirmed

ktbyers:


ktbyers:
Reimplementation of https://github.com/napalm-automation/napalm-base/pull/213

ktbyers:
Originally created by @kderynski

ktbyers:
I want to test this some more to make sure we have the right abstractions.

ktbyers:
```#commit ?
  timer  commit session with a timeout. If not committed within this time, config will be reverted.
  <cr>

```
Software image version: 4.18.3.1F-2GB

ktbyers:
This includes a proof-of-concept on implementing in Cisco IOS.

Note, I need to fix one of the methods as the name is wrong. Also I am not sure I want the `confirm_abort` method or whether we should just rely on the standard rollback method.

napalm/eos/eos.py
```diff
@@ -187,8 +187,11 @@ def compare_config(self):
 
             return result.strip()
 
-    def commit_config(self):
+    def commit_config(self, confirmed=None):
         """Implementation of NAPALM method commit_config."""
+        if confirmed is not None:
```
dbarrosop: https://github.com/napalm-automation/napalm/pull/512#discussion_r150777252

PR #545, opened 2017-11-20
ipv6More int tests

itdependsnetworks:
I disabled the old tests, since I could not figure out how to send parameters down.


dbarrosop:
> I disabled the old tests, since I could not figure out how to send parameters down.

What does this mean? I was trying to follow the PR but is kind of large so I am failing to see what the changes do.

itdependsnetworks:
Trying to figure out how to setup multiple tests for expected_results, so test`expected_results.json` and then `expected_results_canonical.json`. I should be able to configure some additional device parameters, but I cannot figure out how.

dbarrosop:
Yeah, that's going to be tricky... I'd suggest creating a few tests adhoc (as we have for other reasons/drivers) just for the sake of having some testing. Adapting the testing framework for this is quite some work and doesn't give us a huge advantage :(

ktbyers:
@itdependsnetworks Is this still active or should it be closed?

ktbyers:
I am going to close this as it looks like it probably isn't active any longer. If it is still active, then just re-open it.

@itdependsnetworks

itdependsnetworks:
This should be added to the tests, but I need some help on the testing framework.

dbarrosop:
Ping me on slack and I can look at it with you.

PR #548, opened 2017-11-20
Fix nxos_ssh mac table processing bug

ktbyers:


dbarrosop:
Should we add a test case for this or is it not worth it?

PR #550, opened 2017-11-22
DO NOT Merge: Add commit confirmed plus commit message support

ktbyers:


ktbyers:
Resubmitted due to some merge conflicts...see additional notes here: https://github.com/napalm-automation/napalm/pull/544

napalm/junos/junos.py
```diff
@@ -234,8 +234,11 @@ def compare_config(self):
         else:
             return diff.strip()
 
-    def commit_config(self):
+    def commit_config(self, confirmed=None, message=None):
         """Commit configuration."""
+        if confirmed is not None:
```
mirceaulinic: Junos supports commit confirmed (and message).

napalm/iosxr/iosxr.py
```diff
@@ -145,7 +145,10 @@ def compare_config(self):
         else:
             return self.device.compare_config().strip()
 
-    def commit_config(self):
+    def commit_config(self, confirmed=None, message=None):
+        if confirmed is not None:
```
mirceaulinic: IOS-XR too (also message).

ktbyers:
@mirceaulinic Cisco IOS should be ready. I tested that when I implemented it.

ktbyers:
@mirceaulinic Let me know when you add this into other platforms and I will regression test it in my lab environment.

We should also regression test it when it gets merged into develop branch (which I should be able to do also).

napalm/base/base.py
```diff
@@ -178,12 +178,16 @@ def compare_config(self):
         """
         raise NotImplementedError
 
-    def commit_config(self):
+    def commit_config(self, confirmed=None, message=None):
```
dbarrosop: We should document what "confirmed" is supposed to be.

dbarrosop:
I think this whole functionality requires two extra methods:

1. `has_pending_commit()` that returns whether if the system has a commit pending confirmation.
2. `revert_commit()` basically to shortcircuit the timer and revert immediately.

napalm/base/base.py
dbarrosop: I also really think we shouldn't name this variable `confirmed`. It is a terrible name; `revert_in` or `timer` are way better names.

ktbyers: `confirmed` is logical and is the most widely adopted name for this function. Basically saying you have to `confirm` this commit in how many minutes.

I think we will probably just add more confusion by naming it something else.

But agreed...it needs documented.

ktbyers:
Added two methods per @dbarrosop comment (although I changed the name of one of them). Implemented and tested them on Cisco IOS.

napalm/base/base.py
sdanelli: +1 on `revert_in`, describes the intent better. `confirmed` nomenclature maybe better known by network engineers, but I wouldn't cater to them exclusively with this library, you may have engineers without that background.

dbarrosop:
Awesome! :)

ktbyers:
@dbarrosop @mirceaulinic Did some more research and review (which also helped me remember past research):

1. IOS - Should be able to do reasonable system-wide solution.

2. NX-OS - Does not have any reasonable solution (AFAIK). The only thing I see is a kludge that we shouldn't do. This is what I found in past research. We can ask in Cisco channel or NAPALM channel and see if anyone know of a reasonable solution.

3. EOS - I think there is support, but a specific version is required. I believe I had this discussion with @bewing previously.

4. Junos - should be fine

5. IOS-XR / can you commit-confirm in a different SSH session on IOS-XR? I wasn't seeing a good way of doing this in quick testing. @mirceaulinic

dbarrosop:
We need to confirm 5. @mirceaulinic is that something you can look into?

ktbyers:
Closing...don't think it makes sense to keep this PR open any longer.

PR #552, opened 2017-11-23
(junos) fix power entries in EX9208 get_environment()

ckishimo:
While testing an EX9208 running 16.1R5.7 I had the following error:
```
  File "client_junos_lldp.py", line 30, in <module>
    res = dev.get_environment()
  File "/home/ckishimo/devel/napalm/junos/junos.py", line 414, in get_environment
    environment_data['power'][pem_name]['capacity'] = pem_table[pem_name]['capacity']
KeyError: u'power'
```
The 'power' entries are not populated from the first table `junos_enviroment_table`
It seems the command `show chassis environment | display xml` returns the PEM but
with `<class>Temp</class>` instead of `<class>Power</class>`
```
ckishimo@EX9208> show chassis environment | display xml
<rpc-reply xmlns:junos="http://xml.juniper.net/junos/16.1R5/junos">
    <environment-information xmlns="http://xml.juniper.net/junos/16.1R5/junos-chassis">
        <environment-item>
            <name>PEM 0</name>
            <class>Temp</class>
            <status>OK</status>
            <temperature junos:celsius="30">30 degrees C / 86 degrees F</temperature>
        </environment-item>
```
From the mocked data, the PSUs are reported with class `Power`:
```
<environment-information>
  <environment-item>
    <name>FPC 0 Power Supply 0</name>
    <class>Power</class>
    <status>OK</status>
  </environment-item>
```
Attached the output from  `show chassis environment | display xml`
[get-environment-information.log](https://github.com/napalm-automation/napalm/files/1498977/get-environment-information.log)


dbarrosop:
would you mind adding a test case?

ckishimo:
@dbarrosop test case for EX9208 was added!

PR #557, opened 2017-11-23
Fixed device_id and local_int_brief if device_id is long and with spaces

dainok:
If LLDP neighbors are long and with spaces, the last token of the device name is joined to the interface name. This commit fixes this wrong behavior.


ktbyers:
This should be a duplicate of this PR:

https://github.com/napalm-automation/napalm/pull/549

PR #558, opened 2017-11-23
(eos) Add mtu to get_interfaces()

bewing:
Update return values for `EOSDriver.get_interfaces() to return MTU.

Logical (loopback) interfaces are 65535, not sure if we want to change that?


bewing:
Implements #490

Issue #524, opened 2017-11-16
NXOS-SSH get_mac_address_table() error


Rian83:
Hello Guys,

When I try to get the mac add table from a Nexus 5020 (version 5.2(1)N1(9b)).
Do I miss something ?

```
from napalm import get_network_driver
driver = get_network_driver('nxos_ssh')


device=driver('AAAA', 'censured', 'censured')
device.open()
out=device.get_mac_address_table()
```

> Traceback (most recent call last):
>   File "script.py", line 8, in <module>
>     out=device.get_mac_address_table()
>   File "/usr/local/lib/python2.7/dist-packages/napalm/nxos_ssh/nxos_ssh.py", line 1318, in get_mac_address_table
>     raise ValueError("Unexpected output from: {}".format(repr(line)))
> ValueError: Unexpected output from: u'+ 96       000e.1e49.bc22    dynamic   0          F    F  Po1'
>

bewing:
Can you paste us a sanitized output of `show mac address-table` from the device in question?

Rian83:
> HOSTNAME_CENSURED# sh mac address-table
> Legend:
>         * - primary entry, G - Gateway MAC, (R) - Routed MAC, O - Overlay MAC
>         age - seconds since last seen,+ - primary entry using vPC Peer-Link
>    VLAN     MAC Address      Type      age     Secure NTFY   Ports/SWID.SSID.LID
> ---------+-----------------+--------+---------+------+----+------------------
> * 2        0000.0c9f.f002    dynamic   14400      F    F  Po2
> * 2        0026.51c7.45c1    dynamic   110        F    F  Po2
> * 2        0026.9802.8cc1    dynamic   0          F    F  Po2
> * 2        0026.9802.ad41    dynamic   0          F    F  Po2
> * 2        0026.9802.b941    dynamic   60         F    F  Po2
> * 2        0050.5692.05da    dynamic   13920      F    F  Po2
> * 2        0050.5692.0ef6    dynamic   5170       F    F  Po2
> * 2        0050.5692.1056    dynamic   0          F    F  Po2
> * 2        0050.5692.214f    dynamic   3240       F    F  Po2
> * 2        0050.5692.2318    dynamic   2520       F    F  Po2
> * 2        0050.5692.37b9    dynamic   730        F    F  Po2
> * 2        0050.5692.40e9    dynamic   10         F    F  Po2
> * 2        0050.5692.55f9    dynamic   200        F    F  Po2
[...]

PR #563, opened 2017-11-24
Fix bug #524: support multiple interfaces on mac format

Rian83:
I got an issue (https://github.com/napalm-automation/napalm/issues/524)
```
* 3        f60c.be90.7a07    dynamic   300        F    F  Po2
  39       0100.5e00.4b4b    igmp      0          F    F  Po1 Po2 Po22
  516      0100.5e00.4b4b    igmp      0          F    F  Po1 Po2 Po14
  110      0100.5e00.0118    igmp      0          F    F  Po1 Po2 Po14
                                                          Po18 Po24
  110      0100.5e01.0203    igmp      0          F    F  Po1 Po2 Po24
  110      0100.5e01.0207    igmp      0          F    F  Po1 Po2 Po24
```
The 4th and 5th line will crash because : 
4th line : there's multiple interface at the end
5th line : the format of the line is not the same and is the suite from previous line

**What I do :** 
For support multiple interfaces for an unique MAC address: I assume to create a new dict key other_interfaces for keeping backward compatibility
FIX: mulitple interfaces on the same line
FIX: multiple interfaces on multiple lines

Previous output example : 
 `{u'vlan': 20, u'last_move': -1.0, u'active': False, u'mac': u'01:00:5E:7F:FF:FD', u'static': False, u'interface': u'Po1', u'moves': -1}`
New output example : 
`{u'vlan': 20, u'last_move': -1.0, u'active': False, u'mac': u'01:00:5E:7F:FF:FD', u'static': False, u'interface': u'Po1', u'moves': -1, u'other_interfaces': [u'Po2', u'Eth142/1/2']}`

**EDIT 1 :** 
no change of the model


Rian83:
I've changed my previous output thanks to @itdependsnetworks 
Now I return the same output model but multiple entries for the same mac address if multiples interfaces exists.

ktbyers:
@Rian83 Per comments on Slack I assume you are going to add some unit tests in for this?

napalm/nxos_ssh/nxos_ssh.py
```diff
@@ -1243,13 +1243,20 @@ def get_mac_address_table(self):
...
+
         """
 
         #  The '*' is stripped out later
         RE_MACTABLE_FORMAT1 = r"^\s+{}\s+{}\s+\S+\s+\S+\s+\S+\s+\S+\s+\S+".format(VLAN_REGEX,
                                                                                   MAC_REGEX)
         RE_MACTABLE_FORMAT2 = r"^\s+{}\s+{}\s+\S+\s+\S+\s+\S+\s+\S+\s+\S+".format('-',
                                                                                   MAC_REGEX)
+        # REGEX dedicated for lines with only interfaces (suite of the previous MAC address)
+        RE_MACTABLE_FORMAT3 = r"^\s+"
```
ktbyers: I think you would need to expand this to at least:

RE_MACTABLE_FORMAT3 = r"\s+\S+"

Rian83:
@ktbyers Yes I'll add a test for : 
- multiple interfaces on same line
- multiple interfaces on "new line"
- `+` at the begging of the line (because I don't see any test about it)

napalm/nxos_ssh/nxos_ssh.py
Rian83: Yes you're right, I'll change it

dbarrosop:
Thanks for this one, everything seems in order so I will proceed to merge :)

Rian83:
You're welcome @dbarrosop, good news! First contribution

PR #564, opened 2017-11-28
Convert doublebang comments in config to eAPI HEREDOC "comment" commands

### LINKED ISSUE #517: EOSDriver._load_config deletes ACL doublebang comments
```python
test_doublebang = ["ip access-list standard doublebang", "    !! This is a doublebang comment", "10 remark this is a remark", "20 permit 192.0.1.2/32"]
with EOSDriver(s, u, p) as d:
    d.load_merge_candidate(config=test_doublebang)
    print(d.compare_config())

@@ -1280,6 +1279,10 @@
<omitted>
+!
+ip access-list standard doublebang
+   10 remark this is a remark
+   20 permit host 192.0.1.2
 !
<omitted>
@@ -1800,7 +1803,6 @@
<omitted>
 !
 ip access-list standard existing_acl_with_comment
-   !! Testing
 !
```

bewing:
Fixes #517

### DIFF
--- a/napalm/eos/eos.py
+++ b/napalm/eos/eos.py
@@ -57,6 +57,7 @@ class EOSDriver(NetworkDriver):
     HEREDOC_COMMANDS = [
         ("banner login", 1),
         ("banner motd", 1),
+        ("comment", 1),
         ("protocol https certificate", 2)
     ]
 
--- a/napalm/eos/eos.py
+++ b/napalm/eos/eos.py
@@ -162,13 +162,15 @@ class EOSDriver(NetworkDriver):
     @staticmethod
     def _mode_comment_convert(config):
         ret = list(config)
-        d = 1
-        s, _ = next((idx, l) for idx, l in enumerate(ret) if isinstance(l, py23_compat.string_types)
-                    and re.match(r"^(\s+)?!!", l))
-        while re.match(r"^(\s+)?!!", config[s + d]):
-            d = d + 1
-        ret[s] = {'cmd': 'comment', 'input': "\n".join(map(lambda s: s.lstrip("! "), ret[s:s + d]))}
-        del ret[s + 1:s + d]
+        depth = 1
+        start, _ = next(
+            (idx, l) for idx, l in enumerate(ret) if isinstance(l, py23_compat.string_types)
+            and re.match(r"^(\s+)?!!", l))
+        while re.match(r"^(\s+)?!!", config[start + depth]):
+            depth = depth + 1
+        ret[start] = {'cmd': 'comment', 'input': "\n".join(
+            map(lambda s: s.lstrip("! "), ret[start:start + depth]))}
+        del ret[start + 1:start + depth]
 
         return ret
 
--- a/napalm/eos/eos.py
+++ b/napalm/eos/eos.py
@@ -162,7 +162,7 @@ class EOSDriver(NetworkDriver):
     def _mode_comment_convert(config):
         ret = list(config)
         d = 1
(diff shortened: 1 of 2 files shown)


bewing:
This does result in deleting the old comment.  I *COULD* look at the running config and try to append comments, but am not sure if it's worth the effort.

bewing:
Think we're good, just need some review lovin'

napalm/eos/eos.py
```diff
@@ -158,6 +159,19 @@ def _multiline_convert(config, start="banner login", end="EOF", depth=1):
...
+    @staticmethod
+    def _mode_comment_convert(config):
+        ret = list(config)
+        d = 1
+        s, _ = next((idx, l) for idx, l in enumerate(ret) if isinstance(l, py23_compat.string_types)
+                    and re.match(r"^(\s+)?!!", l))
+        while re.match(r"^(\s+)?!!", config[s + d]):
+            d = d + 1
+        ret[s] = {'cmd': 'comment', 'input': "\n".join(map(lambda s: s.lstrip("! "), ret[s:s + d]))}
+        del ret[s + 1:s + d]
```
ktbyers: I always find using very abbreviated names makes the code much harder to read. I have to reverse engineer what the variable is from code. As opposed to having a pretty good idea what the variable is just from its name (and then having it make sense in the context of the code).

napalm/eos/eos.py
```diff
@@ -57,6 +57,7 @@ class EOSDriver(NetworkDriver):
     HEREDOC_COMMANDS = [
         ("banner login", 1),
         ("banner motd", 1),
+        ("comment", 1),
```
ktbyers: I didn't follow how this was used relative to _mode_comment_convert?

bewing: It wasn't, but it does reflect that the following is valid EOS config/CLI commands:

```
ip access-list standard example
    comment
    This is a multi-line comment
    HEREDOC command
    EOF
    permit host 192.0.2.1
```

 It gets displayed in the running-config as 

```
ip access-list standard example
    !! This is a multi-line comment
    !! HEREDOC command
    permit host 192.0.2.1
```

napalm/eos/eos.py
```diff
@@ -179,10 +195,14 @@ def _load_config(self, filename=None, config=None, replace=True):
...
                 continue
-            if line.startswith('!'):
+            if line.startswith('!') and not line.startswith('!!'):
                 continue
             commands.append(line)
 
+        while next((l for l in commands if isinstance(l, py23_compat.string_types)
+                    and re.match(r"(\s+)?!!", l)), None):
+            commands = self._mode_comment_convert(commands)
+
```
ktbyers: Strange...it looks like I lost some earlier review comments. @bewing Can you explain what you are doing here? 

I didn't understand why you could just do an re.search and look for any `!!` lines. Also I didn't see why you needed to check if it py23_compat.string_types?

bewing: I changed the lines you commented on, so you'd have to click the "show outdated" expander in the discussion.

Need to check for string type because I'm actively changing strings to dictionaries.

So while there continue to be strings starting with "!!" in the array, convert the next one.

bewing: Regarding exactly what is happening:
`next()` searches for the first entry in the `commands` list that
* is a string (because re.match will die if we hand it a non-string type)
* has the regex match for `r"(\s)?!!"`

If none are found, `next()` returns None instead of raising `StopIteration`, breaking out of the while loop

ktbyers:
+1

PR #565, opened 2017-11-29
Resolve deprecation warnings in tests.

bewing:
Update test framework, CLI warnings, and regex raw strings to resolve warnings in Tox runs.

### DIFF
--- a/napalm/base/mock.py
+++ b/napalm/base/mock.py
@@ -28,6 +28,9 @@ import re
 from pydoc import locate
 
 
+inspect_getargspec = getattr(inspect, "getfullargspec", inspect.getargspec)
+
+
 def raise_exception(result):
     exc = locate(result["exception"])
     if exc:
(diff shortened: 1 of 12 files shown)


napalm/base/mock.py
```diff
@@ -28,6 +28,9 @@
 from pydoc import locate
 
 
+inspect_getargspec = getattr(inspect, "getfullargspec", inspect.getargspec)
```
ktbyers: Can you just add a comment on why we need this? So we remember it in the future.

napalm/base/test/getters.py
```diff
@@ -24,6 +24,9 @@
 from napalm.base.test import conftest
 
 
+inspect_getargspec = getattr(inspect, "getfullargspec", inspect.getargspec)
```
ktbyers: Can you just add a comment on why we need this? So we remember it in the future.

ktbyers:
Looks good...maybe just add that one comment and then should be good to go.

PR #569, opened 2017-11-30
NXOS XML Parser Support POC

mattschwen:
So far this is working, I still need to verify that all values are returned correctly across all platforms. But take a look, run some tests, and I can work more on it next week.


ktbyers:
This was tested against:

```
Tested this code on the following NXOS platforms with success: 7k, 5k, 2k, 9k, 3k.
```

Rian83:
Did you defined `self.use_xml` ?

PR #570, opened 2017-12-01
don't remove file prompt quiet unnecessarily

benmaddison:
If `file prompt quiet` is already present in the config at the start of an operation, napalm shouldn't remove it.
This is fixed by checking the running-config during `_disable_confirm()`, and setting a flag (`IOSDriver._do_enable_confirm`) to indicate whether the configuration line ought to be removed during later calls to `_enable_confirm()`.
A `RuntimeError` is raised if the attribute does not exist, since this indicates that the methods were called in  the incorrect order.
The configuration line is also removed from `ignore_strings` within `_normalize_compare_config()` since this line will only be present in the diff when it is included in either the candidate and/or (existing) running configs. In that case, a change in that command should show up in `coompare_config()`.


ktbyers:
@benmaddison I am somewhat against tracking state of this as it is prone to problems.

Your solution would also need need to check the current status of the flag (`self._do_enable_confirm`) otherwise multiple calls to _disable_confirm() to get into the wrong state.

I would rather just have an `__init__` optional_args of `enable_confirm=True` and if you passed this in as False, then it would never toggle enable/disable confirm.

itdependsnetworks:
I don't think it is obvious to the end user that you are adjusting the config on their behalf. I'm inclined to agree with @benmaddison

ktbyers:
We have to enable this automatically (otherwise the napalm-ios code will just break and then there will be a lot of issues created and it will be very difficult to manage).

So the solutions I see:

1. Track the state and try to restore it to original state (which I am against as it is problematic to do this reliably).**
2. Give people a way to disable this behavior via an optional argument.
3. Do nothing and let people add a call to restore it in their code.

I think item 2 is a reasonable solution...but feel free to chime in.

There are certain prerequisites to automation using screen-scraping and one of them is to turn off the things that make automation incredibly hard (like prompting for additional things).

**Tracking state is very hard as you can't ensure you restore the state properly. Basically abnormal closures of the script will cause the state not to be restored properly.

dbarrosop:
what about having this as a requirement as we have for the archive functionality?

ktbyers:
@dbarrosop 

I am concerned about debugging/troubleshooting failures associated with this (so I think we should automatically do it). I think the side-effects are quite a bit more benign than automatically enabling the `archive` functionality (which is why I don't think we should do that automatically).

I think we should give people a way to turn this behavior off via an optional_args and that should be sufficient.

Note, if you are not using inline transfers, NAPALM-IOS also automatically enables `ip scp server enable` and does not disable it (this statement only applies for NAPALM configuration operations). This is a more significant change than `file prompt quiet`.

I guess we could do something like check this setting immediately at `open()` and terminate NAPALM if it is not set right with a message and do a similar check for any config operations (for both archive and SCP configuration).

Although I think I would probably rather have config side effects (with the option of disabling them) and make the library easier to use and maintain (than not having config side effects).

Anyways we shouldn't do what is proposed in the PR which is trying to track the state of this setting.

ktbyers:
I would say we open up a separate issue on `napalm-ios` configuration side-effects which includes both `ip scp server enable` and 'file prompt quiet` and decide what to do about them in that issue.

ktbyers:
I think we should probably change the default on that `ip scp server enable` and not automatically enable it (i.e. treat it the same way as archive).

benmaddison:
@ktbyers your point regarding state tracking is fair. This can be made safer using a method decorator that implements the `enable; do a file operation; enable` logic rather than having separate methods that could be called out of order. There is no way to track that state across runs if we encounter (for example) transport errors during `do a file operation` though.
Providing a kwarg to `__init__` is a non-starter for me, as one would need to call the constructor differently depending on the desired final config of the managed device, which rather defeats the whole purpose of using Napalm in the first place.
I have no problem with making `file prompt quiet` a pre-requisite, and agree that this is probably the simplest option. The downside to this is that people may have existing tooling (expect scripts, etc) that assumes file operations *will* be prompted for, and this would break things for those people.
In general config pre-reqs should not be enabled auto-magically, nor removed silently - i.e. they should all be handled like the `archive` requirement.
It would be great if we could agree on a solution to this as soon as possible though, as I am currently having to run my fork in production :-(

ktbyers:
tracking state is too problematic...so I think we can scratch that option off the list. And I think it probably isn't a pattern NAPALM should do (in general).


I didn't understand this below statement?

```
Providing a kwarg to __init__ is a non-starter for me, as one would need to call 
the constructor differently depending on the desired final config of the managed 
device, which rather defeats the whole purpose of using Napalm in the first place.
```

Maybe you are misunderstanding what I was proposing for the optional argument? The optional argument would just cause NAPALM to not modify 'file prompt' at all. NAPALM would neither enable it nor disable it (i.e. you would have to have your device configured already for 'file prompt quiet'). So you would pass this same optional argument in for all your devices. If this argument wasn't used, it would do what it currently does.

```
The downside to this is that people may have existing tooling (expect scripts, etc) 
that assumes file operations will be prompted for, and this would break things for 
those people.
```

I think the above is acceptable. There is only so much you can do.

```
In general config pre-reqs should not be enabled auto-magically, nor removed 
silently - i.e. they should all be handled like the archive requirement.
```

I think I disagree with this to a certain extent (though I understand the sentiment). There is a trade-off in how reliable easy to use the library is versus changing things. For some devices it is a config change to disable output paging. This prompting for additional information is really a terminal characteristic (unfortunately it is one that is retained across time). Or more exactly it should be a VTY characteristic and Cisco chose to not make it one.

Anyways we will need to decide on what to do on this and also on 'ip scp server enable'.

dbarrosop:
> In general config pre-reqs should not be enabled auto-magically, nor removed silently - i.e. they should all be handled like the archive requirement.

I tend to agree with this statement and hence my proposal before but I also agree/understand what ktbyers is saying and sometimes you have to balance usability with cleanness.

> The optional argument would just cause NAPALM to not modify 'file prompt' at all

This is probably a good compromise

benmaddison:
I'm not at all convinced that we should be touching configs to facilitate operations under the hood unless we have absolutely no choice (`terminal length 0` can be run at exec to modify the current vty, just BTW). However, the compromise solution fixes my immediate issue, so I can live with it.
Happy to write the changes if that would be useful?

ktbyers:
@benmaddison On IOS you can change `terminal length 0` just for the VTY...on other platforms that is not necessarily true. So this is more a comment for Netmiko, but it applies to the conversation in general.

Yes, if you want to do a PR...sounds good.

I think we agreed to:
1. Change disable_confirm / enable_confirm so that it can be controlled via an optional argument. It will default to enabled (i.e. to the current behavior).
2. Change 'ip scp server enable' so that it doesn't get automatically changed (i.e. it will behave the same as 'archive'...you will have to manually configure it in advance).
3. Update documentation to reflect new optional argument and to make it clearer on the IOS caveats page that we do this config change.

ktbyers:
@bewing @mirceaulinic I will let you guys chime in...I totally understand the sentiment of not wanting any configuration changes to happen behind the scenes, but am also a bit concerned at additional troubleshooting this might involve (not automatically doing this).

So let us know if you have an opinion on it.

benmaddison:
@ktbyers I've added the `auto_file_prompt` argument to control auto-toggling of `file prompt quiet`.
The `_disable_confirm()` and `_enable_confirm()` are refactored into a method decorator (`_file_operation()`) that can be used to wrap methods that perform file ops.
The docs have been updated, except that someone needs to change `XXX` to the correct value in line 119 of docs/ios.rst before a release gets tagged.
I'll do a separate PR for the SCP change.

napalm/ios/ios.py
```diff
@@ -377,6 +382,33 @@ def compare_config(self):
 
         return diff.strip()
 
+    def _file_operation(f):
```
ktbyers: Let's rename this to something like `_file_prompt_quiet`. `_file_operation` is too generic.

docs/support/ios.rst
```diff
@@ -104,20 +104,30 @@ IOS requires that the banner use the EXT character (ASCII 3). This looks like a
...
 
 Configure replace operations will reject a file with a banner unless it uses the ASCII character. Note, this likely also implies you cannot just copy-and-paste what you see on the screen.
 
 In vim insert, you can also type <ctrl>+V, release only the V, then type C
 
 
+File Operation Prompts
+_____
+
+By default IOS will prompt for confirmation on file operations. These prompts need to be disabled before the NAPALM-ios driver requires these to be disabled.
```
ktbyers: Second sentence doesn't make sense?

docs/support/ios.rst
```diff
@@ -104,20 +104,30 @@ IOS requires that the banner use the EXT character (ASCII 3). This looks like a
...
+This can be controlled using the `auto_file_prompt` optional arguement from version XXX:
```
ktbyers: Drop the version XXX as we will probably not remember to update this at the release time (you can add this post release if you want i.e. when you know the version that this change happened in).

napalm/ios/ios.py
```diff
@@ -436,13 +466,16 @@ def commit_config(self):
         output += self.device.send_command_expect("write mem")
 
     def discard_config(self):
+        """Discard loaded candidate configurations."""
+        self._discard_config()
+
+    @_file_operation
+    def _discard_config(self):
```
ktbyers: Why did you need to make a shadow method here (as opposed to just decorating `discard_config`?

docs/support/ios.rst
benmaddison: Hahaha, no, it really doesn't. I'll fix it...

docs/support/ios.rst
benmaddison: Ack

napalm/ios/ios.py
benmaddison: Ack

napalm/ios/ios.py
benmaddison: This was an interesting py2 vs py3 issue that I hadn't seen before:
In 2.7, `functools.wraps()` doesn't fix the signature of `wrapper()` to match the original function.
As a result, `test.ios.test_getters.TestGetter.test_method_signatures()` fails. Since that test skips private methods the easiest way to work-around was to decorate a private method and call it from the public one.
If you're aware of a more elegant fix then I'm happy to incorporate that?

ktbyers: Okay, makes sense. No I am not aware of a better solution...so we should just use your solution.

PR #574, opened 2017-12-04
Improve get_lldp_neighbors_detail to reduce number of send_command calls to device.

ktbyers:


ktbyers:
Also removed this:

def _expand_interface_name

As it is now deprecated in favor of the canonical_name code.

ktbyers:
FYI, @ogenstad

dainok (changes requested):
I suggest to use the current get_lldp_neighbor_detail only because it's faster.
So: set get_lldp_neighbor and get_lldp_neighbor_detail calling the new function _get_lldp_neighbor_detail.

Also I suggest to add a test for long named devices with spaces:
#show lldp neighbors
[...]
Cisco IP Phone SPA50Gi1/0/22       180        B,T             30e4.db01.2345

ktbyers:
@dainok Updated to have `get_lldp_neighbors` call `get_lldp_neighbors_detail`.

dbarrosop (approved):
Nice!

Issue #572, opened 2017-12-04
Error with parsing BGP summary ipv6 on napalm_ios 


natieitan:
Running get_bgp_neighbors() give me error when we have full ipv6 neighbor 

The error:
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Library/Python/2.7/site-packages/napalm_ios/ios.py", line 1330, in get_bgp_neighbors
    remote_addr = napalm_base.helpers.ip(entry['remote_addr'])
KeyError: u'remote_addr'

```
These are the lines of the show command output:
Neighbor        V           AS MsgRcvd MsgSent   TblVer  InQ OutQ Up/Down  State/PfxRcd
FD11:54::10:48:7:2
                4        11054  291258  291230    47481    0    0 8w5d           13
FD11:54:10:48:98:153:0:10
                4        65048   89045   98921    47481    0    0 8w5d           12
FD11:54:10:48:99:53:0:2
                4        65048   89045   98912    47481    0    0 8w5d           12
```

That's when I printed the entries in summary_data:
{u'router_id': u'178.249.100.124', u'uptime': u'8w5d', u'remote_as': u'11054', u'afi': u'IPv6', u'remote_addr': u'FD11:54::10:48:7:2', u'local_as': u'11054', u'accepted_prefixes': u'13'}
{u'router_id': u'178.249.100.124', u'uptime': u'8w5d', u'remote_as': u'65048', u'afi': u'IPv6', u'local_as': u'11054', u'accepted_prefixes': u'12'}

Seems like the regex can't parse the IPv6 on our table and leaves the remote_addr empty.
It seems to happen only on the IPv6 neighbors without any '::'

targuan:
Hi, it seems there is no tests for ipv6 bgp neighbors for this driver.
Could you provide the output of ```show bgp all summary``` , ```show bgp ipv4 unicast neighbors``` and ```show bgp ipv6 unicast neighbors``` ?

natieitan:
Hi,
Sure, I just had to change part of the IP addresses to X.


show bgp all summary
For address family: IPv4 Unicast
BGP router identifier 178.249.100.124, local AS number 11054
BGP table version is 3671167, main routing table version 3671167
288197 network entries using 71472856 bytes of memory
293932 path entries using 35271840 bytes of memory
98 multipath network entries and 196 multipath paths
50637/2748 BGP path/bestpath attribute entries using 12963072 bytes of memory
[...]

PR #583, opened 2017-12-10
Add test for ipv6 neighbor and fix 572

targuan:
IPv6 regex are bugged for the ios driver and no test exists for it.
This PR fix the IPv6 regex for the "full" IPv6 representation format.

More test data would be welcomed as I mocked most of the data.


dbarrosop:
Fixes #572

dbarrosop:
Thanks a lot for this!

Issue #575, opened 2017-12-06
get_network_driver is not backwards compatible


mirceaulinic:
`get_network_driver` is currently looking for the `driver` class under `napalm.<driver>`, while in the past it was searching under `napalm_<driver>`. Not only this breaks backwards compatibility, but also napalm cannot actually be extended with community drivers. 
In the current form, the developer would need to create the following tree structure: `napalm/<driver>/<driver>.py`. As napalm is a dependency of the community driver, say `napalm-driver`, when executing `pip install napalm-driver` (which firstly install napalm and its structure), would override the `napalm` dir.
So this import doesn't actually work.
I'm wondering if we should instead rename the dirs from `napalm/<driver>` to `napalm_<driver>` which simplifies everything.
Thoughts @dbarrosop @ktbyers @bewing?

dbarrosop:
I think it's easier to just modify `get_network_driver` to search for `napalm_$driver` as well. I guess this ties together with what @itdependsnetworks wanted to do as well for the custom drivers.

mirceaulinic:
That sounds good to me too.

ktbyers:
Yes, I agree I think `get_network_driver` should look for the module as follows (probably in this order)

```
custom_napalm_ios
napalm.ios
napalm_ios
```

i.e. custom_napalm_ios should take precendence if it exists (since you are intentionally trying to override the default behavior), then `napalm.ios` (NAPALM2), then fallback to old drivers `napalm_ios`.

itdependsnetworks:
+1 to @ktbyers

dbarrosop:
I am fine what @ktbyers' proposal

mirceaulinic:
/remind me to work on this in 5 hours

dbarrosop:
if this is what this bot does I think we should remove it. I don’t want to increase the vast amount of emails I already get for napalm. We can each one use our favorite todo app ;)

ktbyers:
I agree with David on this...as it notifies everyone not just the person that set the reminder.

mirceaulinic:
Closed via https://github.com/napalm-automation/napalm/pull/589

PR #589, opened 2017-12-15
update module imports

itdependsnetworks:
This partially addresses: https://github.com/napalm-automation/napalm/issues/575 and https://github.com/napalm-automation-community/napalm-panos/issues/50

On the communituy driver you still get: 

```
>>> driver = get_network_driver('panos')
  File "<stdin>", line 1, in <module>
  File "napalm/base/__init__.py", line 107, in get_network_driver
    .format(install_name=module_install_name))
napalm.base.exceptions.ModuleImportError: No class inheriting "napalm.base.base.NetworkDriver" found in "napalm.panos".
```

> napalm.base.base.NetworkDriver

You still have to do similar on community driver, but wanted to run by you all before adding a bunch of PR's. 

```
try:
    from napalm_base import NetworkDriver
except ImportError:
    from napalm_base.base import NetworkDriver
```


napalm/base/__init__.py
```diff
@@ -79,12 +79,17 @@ def get_network_driver(module_name, prepend=True):
...
             module_install_name = 'napalm.{name}'.format(name=module_install_name)
+        # Order is custom_napalm_os (local only) ->  napalm.os (core) ->  napalm_os (community)
         try:
             module = importlib.import_module("custom_" + module_install_name)
         except ImportError:
-            module = importlib.import_module(module_install_name)
+            try:
+                module = importlib.import_module(module_install_name)
+            except ImportError:
+                module = importlib.import_module(community_module_name)
```
ktbyers: Can we just do a for loop?

custom_install_name = "custom_" + module_install_name
for module_name in [custom_install_name, module_install_name, community_module_name]:
    try:
        ...
        break
    except ImportError:
        pass
    else:
        # Never loaded
        raise ImportError(msg)
```

I think would be clearer to understand than the nested try

**Comment by dbarrosop** [comment]
@mirceaulinic weren't you working on this? Just making sure we don't conflict with each other.

Other than that LGTM

**Comment by mirceaulinic** [comment]
Yes, I did some work, but I didn't test it extensively yet. But I'm happy for @itdependsnetworks to go ahead with this one, no worries.

File: napalm/base/__init__.py [hunk 2]
```diff
@@ -73,24 +73,28 @@ def get_network_driver(module_name, prepend=True):
...
+    # Order is custom_napalm_os (local only) ->  napalm.os (core) ->  napalm_os (community)
+    for module_name in [custom_install_name, module_install_name, community_install_name]:
         try:
-            module = importlib.import_module("custom_" + module_install_name)
+            module = importlib.import_module(module_name)
+            break
         except ImportError:
-            module = importlib.import_module(module_install_name)
-    except ImportError:
+            pass
```
mirceaulinic: I'm not a big fan of doing this, but I guess that works okay.

itdependsnetworks: I did it without forloop, and @ktbyers suggested this way. I'm good either way.

dbarrosop: Not sure if there is a better way to do it but I certainly agree with @ktbyers this is nicer than the nested try and except we had before :)

PR #590, opened 2017-12-15
add recording/replaying commands capability

dbarrosop:
This adds back the capability for recording/replaying commands. Two main use cases are:

1. Compliance.
2. Build easily reproducible tests and demos.

From the previous discussion we had, @mirceaulinic didn't like the use of `camel` library. For reference, `camel` is a lightweight library that builds on top of pyyaml to ensure serialized/deserialized data is safe.

To do:

1. Discuss if `camel` is still an issue and look for alternatives or just "vendor" the library and add it directly without adding it as a requirement.
2. Check with @itdependsnetworks if this does what he needs
3. Upon agreement:
    a. Hook the recorder to rest of core drivers
    b. Remove test/demo code
    c. See how to connect this with the testing framework so people can reproduce issues and provide test cases by using the recording feature.


mirceaulinic:
I am fine to use camel, but I would like to not have it as a hard dependency - i.e., if you want this recording feature, you can install it separately.

Otherwise, I'm totally fine with that.

dbarrosop:
I guess easiest will be then to copy the library directly because I want this to be as easy as just as passing an optional_arg.

mirceaulinic:
I wouldn't have anything against. :-)

dbarrosop:
Ok, done! Or I will forget.

itdependsnetworks:
Does this provide the raw output? whatever it is, meaning ssh or api response.

itdependsnetworks:
I see it now, my personal goal is to not just capture `get_interfaces()` structure data, but what actual commands were ran, and their result. 

```
run_commands.1.yml
raw:
  - cmd: show interfaces 
    text: "long text\nwith\all\config"
output:
  ethernet1:
    input_bytes: 123213 
.....
```

dbarrosop:
It records all the interactions between napalm and the underlying library. So if napalm does `device.send_command("show int")` it will record whatever `send_command` returns (which in this case will be the raw output of the device). However, if the underlying library does any massaging to the data before sending it to us we would see the "massaged" data. This is the case for `pyez` when using the tables, for instance.

dbarrosop:
Well, same as before. You are recording the calls you are making, what I am not saving now is the parameters being passed but that should be easy to add.

itdependsnetworks:
got it, yea, this would be great

dbarrosop:
ok, just added the call parameters as well.

dbarrosop:
Any other comment? Should we merge?

dbarrosop:
Tested with ios, eos and junos (doesn't work on junos). Can someone test on iosxr, nxos and nxos-ssh?

bewing:
nxos throwing exception on open() with commit 633e129:
```
In [1]: from napalm import get_network_driver

In [2]: from getpass import getpass

In [3]: p = getpass()
Password:

In [4]: d = get_network_driver("nxos")("nxos1a", "bewing", p, optional_args={"recorder_mode": "record", "recorder_path": "/tmp/recorder"})

In [5]: d.open()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-5-19d0837c5356> in <module>()
----> 1 d.open()

/mnt/c/Users/bewing/PycharmProjects/napalm/napalm/nxos/nxos.py in open(self)
     85                                    timeout=self.timeout,
     86                                    port=self.port,
---> 87                                    transport=self.transport)
     88             self.device.show('show hostname')
     89             self.up = True

/mnt/c/Users/bewing/PycharmProjects/napalm/napalm/base/recorder.py in __init__(self, cls, recorder_options, *args, **kwargs)
    113         self.record_exceptions = recorder_options.get("record_exceptions", True)
    114
--> 115         self.device = cls(*args, **kwargs)
    116         self.calls = defaultdict(lambda: 1)
    117

TypeError: __init__() got an unexpected keyword argument 'hostname'
```

bewing:
Also, shouldn't call `args` be a list, not a list of lists?  Or am I missing a level here somewhere?

```
>>> import yaml
>>> with open("run_commands.1.yaml") as f:
...     data = yaml.load(f)
...
>>> data
{'call': {'args': [['show clock']], 'func': 'run_commands', 'kwargs': {'encoding': 'text'}}, 'result': [{'output': 'Tue Dec 26 15:56:42 2017\nTimezone: US/Central\nClock source: NTP server (10.203.7.37)\n'}]}
>>>

File: napalm/nxos/nxos.py [hunk 1]
```diff
@@ -69,14 +70,21 @@ def __init__(self, hostname, username, password, timeout=60, optional_args=None)
...
         try:
-            self.device = NXOSDevice(self.hostname,
-                                     self.username,
-                                     self.password,
-                                     timeout=self.timeout,
-                                     port=self.port,
-                                     transport=self.transport)
+            self.device = Recorder(NXOSDevice,
+                                   recorder_options=self.recorder_options,
+                                   hostname=self.hostname,
```
bewing: Should be `host=`, not `hostname=`
https://github.com/networktocode/pynxos/blob/48583d924b0ee171f43545c2c909b5d44aeb91e8/pynxos/device.py#L17

dbarrosop:
@bewing just fixed nxos. Would you mind testing again? Also, could someone test iosxr? It's fine if you don't test the recording capability. I mostly want to make sure the default mode "pass" (do nothing) works so I am not breaking anything.

ktbyers:
Are we still working on this?

ktbyers:
I am going to close this. If someone wants to reopen and actively work on it; they can.

Issue #477, opened 2017-10-25
Inconsistencies between nxos and nxos-ssh driver for get_facts()


ktbyers:
get_facts() on nxos_ssh returns abbreviated interface (from show ip int brief); nx-api returns canonical interface names.

Also model results differs between the two.

Rian83:
Reminder of what is different between two 

on nxos : `"hostname": "XXXX", `
on nxos_ssh : `"hostname": "XXXX.YYY.YYY.YYY?YYY", `

on nxos : `"model": "Nexus 3048 Chassis", `
on nxos_ssh : `"model": "Nexus",`

on nxos : `"os_version": "7.0(3)I4(7)", `
on nxos_ssh : `"os_version": "", `

and interface name different too

ktbyers:
@Rian83 Can you post the output for these four commands from that device above:

show_hostname.txt  
show_hosts.txt  
show_interface_status.txt  
show_version.txt

Note, the model and os_version looks like a bug on NXOS_SSH.

You will also have to provide more details on `interface name different too`

Rian83:
@ktbyers Yes version is a bug which I fixed. I'll do a PR on that. For model also but I need to find someone who have a N5K for example, because I need to have this kind of `sh ver` : 
`Hardware
  cisco Nexus 5596 Chassis ("O2 48X10GE/Modular Supervisor")`

And here's my equipment. I discover that N3K and N9K does not have the same CLI output (as usual :( )
```
XXX# sh hostname 
XXX.y.z.a.com 

XXX# show hosts
DNS lookup enabled
Vrf Name: default Default domain is y.z.a.com
Name servers are   10.48.160.1 10.48.184.1



Host            Address

XXX# sh ver
Cisco Nexus Operating System (NX-OS) Software
TAC support: http://www.cisco.com/tac
Copyright (C) 2002-2017, Cisco and/or its affiliates.
All rights reserved.
The copyrights to certain works contained in this software are
owned by other third parties and used and distributed under their own
licenses, such as open source.  This software is provided "as is," and unless
otherwise stated, there is no warranty, express or implied, including but not
limited to warranties of merchantability and fitness for a particular purpose.
Certain components of this software are licensed under
the GNU General Public License (GPL) version 2.0 or 
GNU General Public License (GPL) version 3.0  or the GNU
Lesser General Public License (LGPL) Version 2.1 or 
Lesser General Public License (LGPL) Version 2.0. 
A copy of each such license is available at
[...]

PR #591, opened 2017-12-15
Fix consistency get_fact() for NXOS and NXOS_SSH drivers

FloLaco:
Fix bug #477 

I've edited the result for the `NXOS_SSH` driver to return the same formating regarting to the `NXOS` driver.
Change I made : 
- The hostname does not return the FQDN anymore
- The model does include the full name until `(brackets for sup module)`
- Consistency interface name using canonical name (Ethernet, port-channel, loopback)
- Fix an empty version for some model


test/nxos_ssh/mocked_data/test_get_facts/missing_domain/expected_result.json
```diff
@@ -1,10 +1,10 @@
 {
     "hostname": "nxos1",
     "fqdn": "",
-    "model": "NX-OSv", 
+    "model": "NX-OSv Chassis",
```
mirceaulinic: To be honest, I'm not sure about this, but looking above at `"model": "Nexus7000 C7009"` I understand your reasoning.

napalm/nxos_ssh/nxos_ssh.py
```diff
@@ -785,7 +791,15 @@ def get_facts(self):
             if line.startswith(' ') or line.startswith('-') or line.startswith('Port '):
                 continue
             interface = line.split()[0]
-            interface_list.append(interface)
+            # Return canonical interface name
+            if re.search(r'^Eth', interface):
```
mirceaulinic: I like having this consistent.

mirceaulinic:
If we are going to accept this, I think we'll get it into a major release only (scheduled for 2.4.0 if everything works as planned).

test/nxos_ssh/mocked_data/test_get_facts/missing_domain/expected_result.json
mirceaulinic: Oh I just realised @ktbyers has the same changes in #476.

Rian83: @mirceaulinic about model, I take into reference the nxapi output. I've installed a titanium switch (nxos virtual) and try the get_facts on both.

Rian83: Oh yes, @ktbyers didn't link his PR to his issue so I didn't know about it oups.
But it doesn't work the same way, specially for the hostname

ktbyers:
I don't think it should go in today. I need to review what I did in that previous PR and there was an issue on that PR that I created that needed fixed.

For any consistency issues we need to look at napalm-base and decide how it should be implemented and then document it there. For example `hostname` and whether it should include the domain-name or not.

In general, I think nxos_ssh is probably a more trustworthy source of truth than nx-api. It was implemented more recently and likely has more test data.

For any interface name expansion we should be using the `canonical_name` support that @itdependsnetworks added in the recent hackathon.

mirceaulinic:
I'm checking in here:

> For example hostname and whether it should include the domain-name or not.

I think it should be like that, yes.

> In general, I think nxos_ssh is probably a more trustworthy source of truth than nx-api. It was implemented more recently and likely has more test data.

+1

> For any interface name expansion we should be using the canonical_name support that @itdependsnetworks added in the recent hackathon.

I agree. @Rian83 are you okay with that?

dbarrosop:
> For example hostname and whether it should include the domain-name or not

I don't think it should. hostname != fqdn.

mirceaulinic:
I just realised my answer is confusing. I also vote +1 for "I don't think it should. hostname != fqdn." - just to clarify what I meant above.

Rian83:
> In general, I think nxos_ssh is probably a more trustworthy source of truth than nx-api. It was implemented more recently and likely has more test data.

Totally agree with that.

@mirceaulinic about canonical interface, yes we should use the it because it's more trustful for the futur. And there is so possibility about the contracted form ("eth", "e", "ether", etc ..)
And more and more i'm working latest nexus 9000, and more I discover some new interface name, like BDI, etc. We definitly should use the canonical name ("Ethernet") instead of the contracted name ("Eth" or E).

itdependsnetworks:
Can you make use of: https://github.com/napalm-automation/napalm/blob/68185ac9cec5350b2d461b00ca03563b4ddfc428/napalm/base/helpers.py#L271

FloLaco:
@itdependsnetworks what's the best practice for adding  some elements in your `base_interfaces = { `

mirceaulinic:
Hey @FloLaco can you please elaborate? If you have a list of interfaces and you're unsure how to map them into `base_interfaces`, just share them here and we or @itdependsnetworks will be able to advise.

itdependsnetworks:
@FloLaco you should be able to just add to the at mapping. For each possible short name, create a key value pair in base_interfaces, and in reverse_mapping pick a standard short name to reverse to.

FloLaco:
I'm using the `canonical_interface_name()` function, but for port-channel, it's returning `PortChannel17` while in a Cisco syntax, it is more like `Port-Channel17`. Which format should we use ?

dbarrosop:
@itdependsnetworks would you mind answering that question?

FloLaco:
Ok i'll fix it.

FloLaco:
@ktbyers I've make the changes about the `canonical_interface_name()` utilization

ktbyers:
This PR is superceded by: https://github.com/napalm-automation/napalm/pull/674

PR #601, opened 2017-12-21
POC CDP: do not merge

dainok:
Implementing get_layer2_neighbors and get_layer2_neighbors_detail using LLDP, falling back on CDP.

### DIFF
--- a/napalm/ios/ios.py
+++ b/napalm/ios/ios.py
@@ -693,6 +693,141 @@ class IOSDriver(NetworkDriver):
 
         return optics_detail
 
+    def get_neighbors(self, proto = None):
+        """IOS implementation of get_neighbors."""
+        if not proto or proto == 'lldp':
+            return self.get_lldp_neighbors_detail()
+        elif proto == 'cdp':
+            return self._get_cdp_neighbors()
+        raise ValueError("Unsupported protocol:\n{}".format(proto))
+
+    def get_neighbors_detail(self, proto = None):
+        """IOS implementation of get_neighbors_detail."""
+        if not proto or proto == 'lldp':
+            return self.get_lldp_neighbors_detail()
+        elif proto == 'cdp':
+            return self._get_cdp_neighbors_detail()
+        raise ValueError("Unsupported protocol:\n{}".format(proto))
+
+    def _get_cdp_neighbors(self):
+        """IOS implementation of get_lldp_neighbors."""
+        cdp = {}
+        neighbors_detail = self._get_cdp_neighbors_detail()
+        for intf_name, v in neighbors_detail.items():
+            cdp[intf_name] = []
+            for cdp_entry in v:
+                cdp_dict = {
+                    'port': cdp_entry['remote_port'],
+                    'hostname': cdp_entry['remote_system_name'],
+                }
+                cdp[intf_name].append(cdp_dict)
+
+        return cdp
+
+    def _get_cdp_neighbors_detail(self):
+        """IOS implementation of get_cdp_neighbors_detail."""
+        cdp = {}
+        command = 'show cdp neighbors detail'
+
+        cdp_detail_output = self._send_command(command)
+        # Check if router supports the command
+        if '% Invalid input' in cdp_detail_output:
+            raise ValueError("Command not supported by network device")
+        cdp_detail = re.split(r"^------------------.*$", cdp_detail_output, flags=re.M)[1:]
+
+        for cdp_entry in cdp_detail:
+            cdp_fields = self._cdp_detail_parser(cdp_entry)
+
+            # Convert any 'not advertised' to 'N/A'
+            for field in cdp_fields:
+                for i, value in enumerate(field):
+                    if 'not advertised' in value:
+                        field[i] = 'N/A'
+            number_entries = len(cdp_fields[0])
+
+            # re.findall will return a list. Make sure same number of entries always returned.
+            for test_list in cdp_fields:
+                if len(test_list) != number_entries:
+                    raise ValueError("Failure processing show cdp neighbors detail")
+
+            # Standardize the fields
+            local_intf, port_id, port_description, chassis_id, system_name, system_description, \
+                system_capabilities, enabled_capabilities, remote_address = cdp_fields
+            standardized_fields = zip(local_intf, port_id, port_description, chassis_id, \
+                                      system_name, system_description, system_capabilities, \
+                                      enabled_capabilities, remote_address)
+
+            for entry in standardized_fields:
+                local_intf, remote_port_id, remote_port_description, remote_chassis_id, \
+                    remote_system_name, remote_system_description, remote_system_capab, \
+                    remote_enabled_capab, remote_mgmt_address = entry
+            if local_intf:
+                local_intf = canonical_interface_name(local_intf)
+
+            cdp.setdefault(local_intf, [])
+            cdp[local_intf].append({
+                'parent_interface': u'N/A',
+                'remote_port': remote_port_id,
+                'remote_port_description': remote_port_description,
+                'remote_chassis_id': remote_chassis_id,
+                'remote_system_name': remote_system_name,
+                'remote_system_description': remote_system_description,
+                'remote_system_capab': remote_system_capab,
+                'remote_system_enable_capab': remote_enabled_capab})
+
+        return cdp
+
+    def _cdp_detail_parser(self, output):
+        # Cisco generally use : for string divider, but sometimes has ' - '
+        local_intf = re.findall(r"Interface\s*?[:-]\s+(.+),.*", output)
+        port_id = re.findall(r".*Port ID \(outgoing port\)\s*?[:-]\s+(.+)", output)
+        port_description = re.findall(r".*Port ID \(outgoing port\)\s*?[:-]\s+(.+)", output)
+        chassis_id = ["not advertised"]
+        system_name = re.findall(r"Device ID\s*?[:-]\s+([^(\n]+)", output)
+        system_description = re.findall(r"Platform\s*?[:-]\s+([^,]+)\s*.*", output)
+        system_capabilities = re.findall(r".*Capabilities\s*?[:-]\s+(.+)", output)
+        if system_capabilities:
+            system_capabilities[0] = system_capabilities[0].replace("Router", "R")
+            system_capabilities[0] = system_capabilities[0].replace("Trans Bridge", "T")
+            system_capabilities[0] = system_capabilities[0].replace("Trans-Bridge", "T")
+            system_capabilities[0] = system_capabilities[0].replace("Source Route Bridge", "B")
+            system_capabilities[0] = system_capabilities[0].replace("Source-Route-Bridge", "B")
+            system_capabilities[0] = system_capabilities[0].replace("Switch", "S")
+            system_capabilities[0] = system_capabilities[0].replace("Host", "H")
+            system_capabilities[0] = system_capabilities[0].replace("IGMP", "I")
+            system_capabilities[0] = system_capabilities[0].replace("Repeater", "r")
+            system_capabilities[0] = system_capabilities[0].replace("Phone", "P")
+            system_capabilities[0] = system_capabilities[0].replace("Remote", "D")
+            system_capabilities[0] = system_capabilities[0].replace("CVTA", "C")
+            system_capabilities[0] = system_capabilities[0].replace("Two-port Mac Relay", "M")
+            system_capabilities[0] = system_capabilities[0].strip()
+        else:
+            system_capabilities = ["not advertised"]
+        remote_address = re.findall(r"Management address\(es\)\s*[:-]\s*(\n.+)", output)
+        # get management addresses and fall back to entry addresses
+        new_remote_address = []
+        for val in remote_address:
+            val = val.strip()
+            pattern = r'\s*IP address\s*[:-]\s+(.+)'
+            match = re.search(pattern, val)
+            if match:
+                new_remote_address.append(match.group(1))
+            else:
+                new_remote_address.append(val)
+        if not new_remote_address:
+            remote_address = re.findall(r"Entry address\(es\)\s*[:-]\s*(\n.+)", output)
+            for val in remote_address:
+                val = val.strip()
+                pattern = r'\s*IP address\s*[:-]\s+(.+)'
+                match = re.search(pattern, val)
+                if match:
+                    new_remote_address.append(match.group(1))
+                else:
+                    new_remote_address.append(val)
+        remote_address = new_remote_address
+        return [local_intf, port_id, port_description, chassis_id, system_name, system_description,
+                system_capabilities, system_capabilities, remote_address]
+
     def get_lldp_neighbors(self):
         """IOS implementation of get_lldp_neighbors."""
         lldp = {}
--- a/napalm/ios/ios.py
+++ b/napalm/ios/ios.py
@@ -693,7 +693,7 @@ class IOSDriver(NetworkDriver):
 
         return optics_detail
 
-    def get_neighbors(self, proto = None):
+    def get_layer2_neighbors(self, proto = None):
         """IOS implementation of get_neighbors."""
         if not proto or proto == 'lldp':
             return self.get_lldp_neighbors_detail()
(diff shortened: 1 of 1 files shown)


dbarrosop:
So three things:

1. We should rename the old method. Or rather do something like:

```
def get_layer2_neighbors(self):
....

get_lldp_neighbors = get_layer2_neighbors
```

Then, for the tests it should be a matter of just adding a symlink.

2. As @ktbyers mentioned we are missing the defintion in the base driver.
3. If we want to discuss a different data format (which I think this implements) we should have a  discussion around the definition in the base driver.

dainok:
Thank you both. Sorry but I'm currently overbusy, I'll be back (hopefully) in few weeks.

ktbyers:
@dainok No worries.

orgito:
I'd like to suggest a different approach:

Add `get_cdp_neighbors` and `get_cdp_neighbors_details` methods .

The data format for `get_cdp_neighbors` is the same as `get_lldp_neighbors`. But `get_cdp_neighbors_details` has to be a different format as CDP does not return the same information as LLDP does.

Having the explicit methods allows the user to choose which method to use based on his knowledge of the network, or try both.

The `get_layer2_neighbors` and `get_layer2_neighbors_details` should return the combination of both commands. This is necessary to cover a scenario where a device has both LLDP and CDP enabled and have CDP neighbors in some ports and LLDP neighbors in others.

orgito:
Hello @dainok . This is a feature I'm interested in. Are you getting back to it? Can I help you?

dainok:
Hello @orgito currently I'm busy with other things. The only thing about merging CDP and LLDP outputs, is that they are using a different naming convention on remote node type. So a merge will require to convert CDP into LLDP.
Beside that, to be accepted, a patch must work for ios, asr and nxos.

orgito:
@dainok. By remote node type I believe you mean the system capabilities, right? Yes, that is the problem I see when merging. I don't believe is possible to translate between LLDP capabilities and CDP capabilities. It is not just different names, it is also different types. It would also be confusing to the user. 

Maybe the solution is to leave `remote_system_capab` and `remote_system_enabled_capab` for LLDP and add a new field `remote_system_cdp_capab`. It will be a little confusing, but would allow to merge the output from cdp and lldp in the layer2 method.

dainok:
@orgito yes you're right it was system cap. I think some cap can be translated to LLDP. But I guess @ktbyers / @dbarrosop / @mirceaulinic should comment on that too.

ktbyers:
1. I don't think we should have a Cisco specific napalm method i.e. we want methods that will work across all core platforms (i.e. we should use the `get_layer2_neighbors`).

2. I am a bit leery about combining the output results i.e. you should either get LLDP or CDP, but not simultaneously both. Is the only issue as far as the data the remote system capability? If we combine them, we should definitely have some sort of field added that indicates the data source i.e. 'lldp' or 'cdp'. 

3. The CDP parsing will need to work on all of the napalm-cisco core platforms (ios, nxos, nxos_ssh, iosxr).

orgito:
All right. That simplified things a bit. So we will have `get_layer2_neighbors` that will return LLDP neighbors by default. I think that also solve the problem of the capabilities: There is no need to translate the CDP capabilities initials since the user will know how to interpret it since he explicitly set the protocol to CDP.

jobec:
For my use case (mapping out a layer 2 topology of a network) it would be more preferable to have cdp and lldp combined. I don’t really care where the data came from, as long as I can identify the remote system/interface.

If you run it on thousands of devices of multiple vendors, it’s an advantage not having to worry about the type of device you’re speaking with. In some cases you also don’t know. For example when all you have is a napalm driver instance and no other info about the device. (Think Celery tasks, thousands and thousands of celery tasks 😬)

So, cdp & lldp as a default would be nice, or at least a parameter to get them both at the same time in the same data structure.

ktbyers:
So what would the data structure of a combined LLDP + CDP look like. Here are a couple of options:

```
{  'lldp': {
    'Ethernet2':
                [
                    {
                      'hostname': u'rtr1',
                      'port': u'Gi0/0',
                    },
                ],
    }
    'cdp': {
      'Ethernet2':
                [
                    {
                      'hostname': u'rtr1',
                      'port': u'Gi0/0',
                    },
                ],
    }
}
```

Or

```
{
    u'Ethernet2':
                [
                    {
                      'hostname': u'rtr1',
                      'port': u'Gi0/0',
                      'protocol': u'lldp',
                    },
                    {
                      'hostname': u'rtr1',
                      'port': u'Gi0/0',
                      'protocol': u'cdp',
                    },
                ],
}
```

I prefer option1.

jobec:
Let's have a look at both LLDP and CDP fields from the official standards.

# LLDP

LLDP TLV | -
-------- | -
Chassis ID | Mandatory
Port ID | Mandatory
Time To Live | Mandatory
Port Description | Optional
System Name | Optional
System Description | Optional
System Capabilities Available | Optional
System Capabilities Enabled | Optional
Management Address | Optional

ref: https://ieeexplore.ieee.org/document/7433915/

# CDP

CDP TLVs |
-------- |
Address
Application
Capabilities
Device-ID
Full or Half Duplex
IP Network Prefix
Location
Location-Server
Native VLAN
Platform
Port-ID
Version
VTP Management Domain

Ref:
https://www.cisco.com/c/en/us/td/docs/ios-xml/ios/cdp/configuration/15-mt/cdp-15-mt-book/nm-cdp-discover.html
http://www.rhyshaden.com/cdp.htm


# Cross reference with NAPALM

If you look at the table below, I think we can represent both LLDP and CDP info in the structure currently used for LLDP.

get_lldp_neighbors()        |         LLDP                  |       CDP
----------------------------|-------------------------------|------------------
parent_interface            | N/A                           | N/A
remote_chassis_id           | Chassis ID                    | Device-ID
remote_port                 | Port ID                       | Port-ID
N/A                         | Time To Live                  | Time To Live
remote_port_description     | Port Description              | Port-ID
remote_system_name          | System Name                   | Device-ID
remote_system_description   | System Description            | Platform, Version
remote_system_capab         | System Capabilities Available | 
remote_system_enabled_capab | System Capabilities Enabled   | Capabilities
N/A                         | Management Address            | Address

Add in a field that describes the source of the data, some kind of generalized translation of the capability fields (a bitmask maybe instead of text?), a TTL and address field and this should be rather complete.

Resulting in this example:
```
{
    'TenGigE0/0/0/8': [
        {
            'parent_interface': u'Bundle-Ether8',
            'remote_chassis_id': u'8c60.4f69.e96c',
            'remote_system_name': u'switch',
            'remote_port': u'Eth2/2/1',
            'remote_port_description': u'Ethernet2/2/1',
            'remote_system_description': u'Cisco Nexus Operating System (NX-OS) Software 7.1(0)N1(1a)',
            'remote_system_capab': 3,
            'remote_system_enable_capab': 3,

            'source': u'LLDP',
            'TTL': 123,
            'address': "10.1.1.1",
        }
    ]
}
```

jobec:
As for the mapping between CDP and LLDP capabilities, here's a first analysis of it.

### LLDP

From IEEE document about LLDP

Bit   | Cisco Code | Juniper Code | Capability
----- | ---------- | ------------ | ------------
0x001 | `O`        |              | Other
0x002 | `P`        |              | Repeater
0x004 | `B`        |              | MAC Bridge component
0x008 | `W`        |              | 802.11 Access Point (AP)
0x010 | `R`        | Router       | Router
0x020 | `T`        | Telephone    | Telephone
0x040 | `C`        |              | DOCSIS cable device
0x080 | `S`        |              | Station Only
0x100 |            |              | C-VLAN component
0x200 |            |              | S-VLAN component
0x400 |            |              | Two-port MAC Relay component


### CDP

From https://wiki.wireshark.org/CDP

Bit  | Cisco Code | Capability
---- | ---------- | ------------
0x01 | `R`        | Router
0x02 | `T`        | Transparant Bidge
0x04 | `B`        | Source Route Bridge
0x08 | `S`        | Switch
0x10 | `H`        | Host 
0x20 | `I`        | IGMP capable
0x40 | `r`        | Repeater
0x80 | `P`        | Phone (Not documented but visible in show command on IOS)

ktbyers:
The above doesn't discuss the main issue?

I.E. how should we differentiate between CDP and LLDP in the data structure i.e. should it be the highest level key or should be a key embedded in the inner dictionary:

```
'lldp': { inner dictionary },
'cdp': { inner dictionary },
```

Or

```
'Ethernet2': [
     { inner dict where field indicates source }
]
```
   
Advantages of the first are clear indication of source. Makes it easier (probably) to handle the duplicate entry problem. Makes it easier to transition get_lldp_neighbors() output to this data structure.

Advantages of the second: makes it easier to get the keys of all the interfaces that have data.


Other points:
1. We shouldn't do translation between LLDP and CDP on capabilities (i.e. NAPALM won't attempt to normalize this). If people want to do this, they would have to post process it.
2. No new fields would be be added to the returned data structure (there would have to be really strong need for new fields). The only exception to this potentially being the 'source' if we add the LLDP/CDP differentiator here.
3. No filtering for duplicate entries i.e. if an entry is reported by both CDP and LLDP.

jobec:
Concerning the normalisation: 

https://github.com/napalm-automation/napalm/blob/d7961e8d0efbe6db3cfd4e373edc4d3f0d54f62f/test/junos/mocked_data/test_get_lldp_neighbors_detail/qfx5110/expected_result.json#L8
Vs
https://github.com/napalm-automation/napalm/blob/d7961e8d0efbe6db3cfd4e373edc4d3f0d54f62f/test/ios/mocked_data/test_get_lldp_neighbors_detail/alternate4/expected_result.json#L10

I’m rather new to NAPALM, but when I started with it, I expected the output of something like get_lldp_neighbors_detail() to be device/vendor agnostic. The platform specific post processing should be very limited.

In the example above, if I want to only process specific capabilities, I need to know for every platform how it’s being displayed. For an abstraction layer, that’s unexpected. And again, when making driver instances for thousands of devices and processing them, you don’t always know what kind of device it is (except when regexing the class name of the driver) and what post processing you have to perform.

Not meant as criticism, but for my understanding, what is napalm’s strategy on this subject? Provide standard functions, implement the logic to get the result and give them back in whatever format the device displays it?

My idea on this is, that if NAPALM would give back the same result, irrespective from the platform, it’s strength would much larger. But then again, I’m looking at napalm with 1 month of experience 😇

orgito:
@ktbyers, I did not understand your point about option 1 making it easier to transition  get_lldp_neighbors() to the new data structure. I think it will break any current code if it returns data way.
On the other hand if we simply add the source field than  get_lldp_neighbors() =  get_layer2_neighbors(proto=lldp). Or I'm getting it wrong.

So can I suggest having simply: 
 get_layer2_neighbors having the source field indicating the protocol. If the protocol is specified only that protocol data is returned. If the proto parameter is None we return both data. 

One advantage is that get_lldp_neighbors would simply be get_layer2_neighbors = proto defined as lldp.

Also, is it really necessary the chage the data structure for get_lldp_neighbors? Why not just leave it as it is?

ktbyers:
@jobec NAPALM generally provides uniform data structures. We do normalize some data like IP addresses, and MAC addresses (NAPALM is a bit mixed on this). There is some data normalization; there are other cases where we intentionally don't normalize data.

I think we shouldn't try to normalize here: both in the LLDP case and across LLDP/CDP. It is too much effort for not enough value.

For context, the maintenance work on the existing getters is already too much (way too much). The intent was to move to napalm-yang instead of implementing any more getters (but we have been equivocating on this so we need to revisit it).

So I would say the behavior for LLDP is correct:

```
napalm/test/junos/mocked_data/test_get_lldp_neighbors_detail/qfx5110/expected_result.json

Line 8 in d7961e8

 "remote_system_capab": "Bridge Router",  

Vs
napalm/test/ios/mocked_data/test_get_lldp_neighbors_detail/alternate4/expected_result.json

Line 10 in d7961e8

 "remote_system_capab": "B,R", 
```

We are providing the `remote_system_capab` reported by the system. But we are not trying to rationalize it across platforms or OS versions.

ktbyers:
@orgito All I meant by that is that we could just do this:

```
{
    `lldp`: { inner LLDP dictionary },
   `cdp`: { inner CDP dictionary },
}
```

where inner LLDP dictionary exactly matches what is returned from `get_lldp_neighbors`.

Yes, if people call `get_layer2_neighbors` it will be different than `get_lldp_neighbors`, but `get_layer2_neighbors['lldp'] would be identical to `get_lldp_neighbors`.


```
So can I suggest having simply:
get_layer2_neighbors having the source field indicating the protocol. If the protocol is specified only that protocol data is returned. If the proto parameter is None we return both data.
```

If you return `both` data, can you post what you envision the returned data structure would look like?

I think the issue here is the `both` case i.e. the inner data structure would somehow need to indicate the source as being `lldp` or `cdp` (due to the duplicate entry problem and also possibly to post process differently).

```
Also, is it really necessary the change the data structure for get_lldp_neighbors? 
Why not just leave it as it is?
```

Yes, I agree we should try to keep it the same (if we can).

jobec:
I think it boils down to deciding whether to make the CDP structure the same as the LLDP structure or explicitly choose a different structure.

#### If **A** (the same)
From the point of view: _I just want to know the neighbor_

In this case it makes more sense to add a `source` field. Because your loop logic can remain the same whether it's CDP of LLDP. 

#### if **B** (different)
From the point of view: _The data in the fields are too different and need to be handled separately_

Then it makes more sense to put it in 2 separate keys, as you'll be processing both protocols in different loops, extracting different fields. This keeps you from needing to filter out LLDP from CDP or vise versa and then do the processing.

Who's the BDFL of napalm to make the final decision? 😁

ktbyers:
@jobec How would you handle the duplicate entries i.e. the same device being reported via both CDP and LLDP?

That is the main thing that makes me not like commingling the data.

ktbyers:
@jobec I disagree with your last statement...in both cases the inner data structure between LLDP and CDP will be identical. 

Generally @mirceaulinic @dbarrosop and I decide. Looping them into the decision, here are the options that I see (I will just focus on `get_layer2_neighbors` as I think the same answer will apply to `_details`).


Options:

1. get_layer2_neighbors() has a protocol option. For example,

```
get_layer2_neighbors(proto='lldp')
get_layer2_neighbors(proto='both')
```

** If this is the solution, do we need a field added to the returned data structure that indicates the source of the data ('lldp' or 'cdp').

2. get_layer2_neighbors() just returns a single data structure that contains both sets of data.

```
{
     'lldp': inner_data_structure_identical_to_get_lldp_neighbors,
     'cdp': inner_data_structure_identical_to_get_lldp_neighbors,    # except for CDP data
}
```

3. get_layer2_neighbors() returns a single data structure with CDP and LLDP data commingled:

```
{
        'Ethernet2':
                [
                    {
                      'hostname': u'rtr1',
                      'port': u'Gi0/0',
                      'protocol': 'lldp',
                    },
                ],
    }

**Comment by jobec** [comment]
You won't have duplicate entries. The "source" field will be different.

e.g.
```
{
    'Ethernet2':
    [
        {
            'hostname': u'rtr1',
            'port': u'Gi0/0',
            'protocol': 'lldp',
        },
        {
            'hostname': u'rtr1',
            'port': u'Gi0/0',
            'protocol': 'cdp',
        },
    ],
}
```

**Comment by ktbyers** [comment]
Yes, I meant duplicate entries from the device perspective i.e. the same end-system being reported on the same interface once via LLDP and once via CDP (potentially with slight differences).

**Comment by ktbyers** [comment]
We will say what @dbarrosop and @mirceaulinic chime in with...then we can finalize this definition and move on.

**Comment by ktbyers** [comment]
My vote is for either number 1 or number 2; I like number 3 the least...though maybe number1 with a `proto=<value>` would give both ways (i.e. separate data if you want it; commingled data if you want it).

**Comment by jobec** [comment]
option 3 = option 1 with `proto='both'`

**Comment by ktbyers** [comment]
```
option 3 = option 1 with proto='both'
```

Yes...but option1 gives you the other ways (i.e. proto='lldp' and proto='cdp' which is what I want). 

In other words, you want the commingled data; I don't...option1 gives you both ways.

ktbyers:
I am going to close this. I don't think anyone is working on it. If someone is actively working on it, then we can revisit it.

PR #603, opened 2017-12-25
Track BGP state to report is_up

### LINKED ISSUE #573: Issue with BGP neighbors status
I'm using napalm to show and change BGP configuration.
I noticed that it always shows 'true' under 'us_up' even if the neighbor is down.
I saw on the code, the following:
            # Uptime should be -1 if BGP session not up
            is_up = True if uptime >= 0 else False

If a neighbor is down there's still a number there, for example:
Neighbor        V           AS MsgRcvd MsgSent   TblVer  InQ OutQ Up/Down  State/PfxRcd
10.0.0.10       4        11055   24061   24287       22    0    0 2w1d           11
10.0.0.14       4        11055       0       0        1    0    0 00:19:49 Idle

Before using napalm I did it in another way, I tracked the last column, true if it's a number, false if it's not, that should be the accurate way

The function: 

def represents_int(s):
    try:
        int(s)
        return True
    except ValueError:
        return False

The code for the actual check:
                                if represents_int(line.split()[9]):
                                    lines_status[isp['name']][4] = 'UP'
                                else:
                                    lines_status[isp['name']][4] = 'DOWN'

line - represent each line of the show command: 'show ip bgp sum'

targuan:
This should fix #573 by tracking the real state of the BGP session instead of inferring it based on calculated values.

@ktbyers could you confirm that the tests data where incomplete ? I had to add the BGP state line to some tests but the mocked data seemed to have been shortened to what was only needed.


test/ios/mocked_data/test_get_bgp_neighbors/572/show_bgp_ipv4_unicast_neighbors.txt
```diff
@@ -1,6 +1,7 @@
 BGP neighbor is 192.168.0.2,  remote AS 65001, internal link
  Inherits from template iBGP for session parameters
   BGP version 4, remote router ID 0.0.0.0
+  BGP state = Idle
```
dbarrosop: I don't think this is correct. We shouldn't manipulate mocked data to have tests pass. If this line wasn't here is because this device doesn't report it.

targuan:
@dbarrosop that was the point of my question to @ktbyers. The mocked data seems odd to me. I've never seen such output on a cisco device for this command, it seems truncated.

ktbyers:
@targuan The mocked data came from someone else so we would need to track down that person and verify it was not originally modified.

I think we should start with that.

ktbyers:
@benmaddison Can you verify the validity of this test data...see the files changed for what might not be correct:

https://github.com/napalm-automation/napalm/blame/develop/test/ios/mocked_data/test_get_bgp_neighbors/normal/show_bgp_ipv4_unicast_neighbors.txt

ktbyers:
Here is what an IOS router I have shows:

```
pynet-rtr1#show bgp ipv4 unicast neighbors 
BGP neighbor is 10.220.88.38,  remote AS 44, external link
  BGP version 4, remote router ID 10.220.88.38
  BGP state = Established, up for 20w2d
  Last read 00:00:59, last write 00:00:19, hold time is 180, keepalive interval is 60 seconds
  Neighbor sessions:
    1 active, is not multisession capable (disabled)
  Neighbor capabilities:
    Route refresh: advertised and received(new)
    Four-octets ASN Capability: advertised and received
    Address family IPv4 Unicast: advertised and received
    Graceful Restart Capability: received
      Remote Restart timer is 120 seconds
      Address families advertised by peer:
        none
    Enhanced Refresh Capability: advertised
    Multisession Capability: 
    Stateful switchover support enabled: NO for session 1
  Message statistics:
    InQ depth is 0
    OutQ depth is 0
    
                         Sent       Rcvd
    Opens:                  1          1
    Notifications:          0          0
    Updates:               40          1
    Keepalives:        226215     205516
    Route Refresh:          3          0
    Total:             226259     205518
  Default minimum time between advertisement runs is 30 seconds

 For address family: IPv4 Unicast
  Session: 10.220.88.38
  BGP table version 59, neighbor version 59/0
  Output queue size : 0
  Index 1, Advertise bit 0
  1 update-group member
  Slow-peer detection is disabled
  Slow-peer split-update-group dynamic is disabled
  Interface associated: FastEthernet4
                                 Sent       Rcvd
  Prefix activity:               ----       ----
    Prefixes Current:               0          0
    Prefixes Total:                53          0
    Implicit Withdraw:             24          0
    Explicit Withdraw:             29          0
    Used as bestpath:             n/a          0
    Used as multipath:            n/a          0

                                   Outbound    Inbound
  Local Policy Denied Prefixes:    --------    -------
    Total:                                0          0
  Number of NLRIs in the update sent: max 2, min 0
  Last detected as dynamic slow peer: never
  Dynamic slow peer recovered: never
  Refresh Epoch: 1
  Last Sent Refresh Start-of-rib: never
  Last Sent Refresh End-of-rib: never
  Last Received Refresh Start-of-rib: never
  Last Received Refresh End-of-rib: never
				       Sent	  Rcvd
	Refresh activity:	       ----	  ----
	  Refresh Start-of-RIB          0          0
	  Refresh End-of-RIB            0          0

  Address tracking is enabled, the RIB does have a route to 10.220.88.38
  Connections established 1; dropped 0
  Last reset never
  Transport(tcp) path-mtu-discovery is enabled
  Graceful-Restart is disabled
Connection state is ESTAB, I/O status: 1, unread input bytes: 0            
Connection is ECN Disabled, Mininum incoming TTL 0, Outgoing TTL 1
Local host: 10.220.88.20, Local port: 179
Foreign host: 10.220.88.38, Foreign port: 35105
Connection tableid (VRF): 0
Maximum output segment queue size: 50

Enqueued packets for retransmit: 0, input: 0  mis-ordered: 0 (0 bytes)

Event Timers (current time is 0xB40C407E0):
Timer          Starts    Wakeups            Next
Retrans        226254          0             0x0
TimeWait            0          0             0x0
AckHold        205518     202075             0x0
SendWnd             0          0             0x0
KeepAlive           0          0             0x0
GiveUp              0          0             0x0
PmtuAger            0          0             0x0
DeadWait            0          0             0x0
Linger              0          0             0x0
ProcessQ            0          0             0x0

iss: 2656486589  snduna: 2660786510  sndnxt: 2660786510
irs: 4254025851  rcvnxt: 4257930738

sndwnd:  31419  scale:      0  maxrcvwnd:  16384
rcvwnd:  16251  scale:      0  delrcvwnd:    133

SRTT: 1000 ms, RTTO: 1003 ms, RTV: 3 ms, KRTT: 0 ms
minRTT: 0 ms, maxRTT: 1000 ms, ACK hold: 200 ms
uptime: -1 ms, Sent idletime: 19016 ms, Receive idletime: 18784 ms 
Status Flags: passive open, gen tcbs
Option Flags: nagle, path mtu capable
IP Precedence value : 6

Datagrams (max data segment is 1460 bytes):
Rcvd: 431069 (out of order: 0), with data: 205518, total data bytes: 3904886
Sent: 431088 (retransmit: 0, fastretransmit: 0, partialack: 0, Second Congestion: 0), with data: 226255, total data bytes: 4299920

 Packets received in fast path: 0, fast processed: 0, slow path: 0
 fast lock acquisition failures: 0, slow path: 0
TCP Semaphore      0x89CA56DC  FREE 
```

ktbyers:
test_get_bgp_neighbors/572           `Created by @targuan, doesn't look like real data`
test_get_bgp_neighbors/normal	 `Created by @benmaddison doesn't look like real data`
test_get_bgp_neighbors/real_life     `Looks real`
test_get_bgp_neighbors/regex_strip       `Created by @benmaddison doesn't look like real data`


I think we should move `real_life` to `normal` and then see if we can capture real data for the other scenarios (i.e. if there are real situations/problems we are trying to handle in the other test cases).

benmaddison:
@ktbyers as far as I can remember, i added `real_life` when i re-implemented `get_bgp_neighbors` based on output from a real peering router, but changed addresses, ASNs and router-ids to reserved values for privacy reasons. I don't recall changing the pre-existing mock data, but i could be miss-remembering.

afourmy:
when I was working on adding support for all AFI, I made https://github.com/napalm-automation/napalm/pull/494. I added some unmodified mock data for the new tests there, maybe that helps.

PR #607, opened 2018-01-02
More flexible options for the get_network_driver function

mirceaulinic:
In https://github.com/napalm-automation/napalm/pull/526 @itdependsnetworks added support for libraries the user has in it's own environment. While I agree with this and I am a big supporter of this flexibility, it enforces you to name the library `custom_napalm` (i.e., everyone needs to name it that way).
In this PR I'm adding the possibility so you can import the class from whatever library without having any naming constraints, e.g.:

```python
>>> get_network_driver('junos', lib='my_custom_library')
<class 'my_custom_library.junos.CustomJunOSDriver'>
>>> get_network_driver('junos', lib='my_custom_library', prepend=False)
<class 'my_custom_library.CustomJunOSDriver'>
>>> get_network_driver('junos', prefix='my')
<class 'my_napalm.junos.MyJunOSDriver'>
>>> get_network_driver('junos', prefix='my', prepend=False)
<class 'my_napalm.MyJunOSDriver'>
```

### DIFF
--- a/napalm/base/__init__.py
+++ b/napalm/base/__init__.py
@@ -35,7 +35,10 @@ __all__ = [
 ]
 
 
-def get_network_driver(module_name, prepend=True):
+def get_network_driver(module_name,
+                       prepend=True,
+                       prefix=None,
+                       lib=None):
     """
     Searches for a class derived form the base NAPALM class NetworkDriver in a specific library.
     The library name must repect the following pattern: napalm_[DEVICE_OS].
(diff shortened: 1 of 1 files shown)


ktbyers:
@mirceaulinic I think the idea is fine, but can we simplify the arguments? The arguments are getting overly confusing.

Currently, you propose:

```
def get_network_driver(module_name,
                       prepend=True,
                       prefix=None,
                       lib=None)
```

So the `prepend=True` is converting `ios` to `napalm.ios`? Do we need this any longer or can we deprecate that in the future? With the current code, when would prepend=False ever happen?

get_network_driver('ios') will currently look for this:

```
(Pdb) custom_install_name
'custom_napalm.ios'
(Pdb) module_name
'ios'
(Pdb) module_install_name
'napalm.ios'
(Pdb) community_install_name
'napalm_ios'
```

Can we just add one argument:

```
custom_lib='custom_napalm'
```

Then have the `custom_install_name` do this:

```
custom_install_name = "{lib}.{name}".format(lib=custom_lib, name=module_install_name)
```

I guess we can have one additional attempt if the custom import fails, where it just tries to load straight `custom_lib`...so that you can do:

custom_lib = 'whatever_foo'

and have files of

```
whatever_foo/junos.py
whatever_foo/ios.py
```

or a file named `whatever_foo.py` (though in that case it would totally ignore the driver name in the `get_network_driver`).

Anyways, am open to the idea...just would like to simplify the arguments.

dbarrosop:
I agree with everything @ktbyers said. We should just provide a standard pattern people can reproduce to discover drivers without having to overcomplicate this method. I personally think we should have only two arguments:

```
def get_network_driver(driver_name, module_path=None, fallback=True)
```

* driver_name is the driver you want (eos, ios, junos, etc)
* module_path is the path for an alternative module, ie., `my_custom_napalm` or `whatevs.napalm`
* fallback is if we want to fallback to our core drivers if none is found in `module_path`

We could even make `module_path` a list so users can specify multiple potential paths, ie, `['my_custom_napalm.ios', 'my_custom_napalm.eos', ...]`

mirceaulinic:
I agree, great ideas.

ktbyers:
@mirceaulinic Is there anything more to do on this PR or should we just close it--i.e. you resubmit a new PR if there are changes you are proposing?


Kirk

ktbyers:
I am going to close this PR...just re-open it if it is still active.

Issue #600, opened 2017-12-21
get_users(): can we get root/admin (junos) and enable password (ios)?


sincerywaing:
In junos it seems that there are two static users - root and admin, can we get them when we do get_users()?
also can we also get enable password in ios as well if it does not break anything?

DadaHiT:
get_users() can get admin but cannot get root-authentication

In [15]: device.get_users()
Out[15]:
{**'admin'**: {u'level': 0,
  u'password': u'***************',
  u'sshkeys': []},
 **'remote'**: {u'level': 0, u'password': u'', u'sshkeys': []}}

dbarrosop:
I'd recommend having two different issues to track these changes. The `root` issue should be quite straightforward to implement (or at least it only involves changes in the junos driver). The other one requires modifying the `base` definition of the method and implementing the change in all the drivers so it requires more work. In any case, I think both changes are fine, we just need someone to implement them :P

sincerywaing:
@dbarrosop I'm changing this to target Junos. Will open a new one on ios.

sincerywaing:
@dbarrosop sent a PR to this but it seems can't get Travis passed...can you take a look?

PR #608, opened 2018-01-03
Issue #600: junos add root user to get_users()

sincerywaing:
this is a fix for issue #600: add root user to `get_users()` method.

1. add a private method `_get_root(self)` to specifically handle root user;
2. add a new view `junos_root_table` for above method.
3. put `_get_root()` as an embedded method for `get_users()`


mirceaulinic:
The issue with this is that the root user is not necessarily named `root` on every platform. It is constrained to be named `root` on Junos, or `admin` on Arista, for example. What I would suggest here is that we should maybe have a flag for the root, or a way to distinguish that.
For example, in the [openconfig-system](http://ops.openconfig.net/branches/master/openconfig-system.html) model, there's a separate hierarchy (on the path `system` > `aaa` > `authentication` > `admin-user`).

sincerywaing:
but when we login we use root as username, so I’d still see its a static user at least for junos.

dbarrosop:
> What I would suggest here is that we should maybe have a flag for the root, or a way to distinguish that.

I'd avoid doing that. It's just a user and the privilege level should give away the information.

napalm/junos/junos.py
```diff
@@ -1823,6 +1823,38 @@ def ping(self, destination, source=C.PING_SOURCE, ttl=C.PING_TTL,
 
         return ping_dict
 
+    def _get_root(self):
+        """get root user password."""
+        _DEFAULT_USER_DETAILS = {
+            'level': 0,
```
dbarrosop: why don't you just set this number to 15 here and avoid later the `user_details.update({'level': 15})`?

test/junos/mocked_data/test_get_users/normal/expected_result.json
```diff
@@ -1 +1 @@
...
...Uq6YbUEcgUTRzzXfmelJKuVJTJIlMXii7h2xkbQp0YZIEs4P8ipwnRBAxFfk/ZcDsN3mjep4/yjN56eorF5xs7zP9HbqbJ1dsqk1p3A/9LIL7l6YewLBCwJj6D+fWSJ0/YW+7oH17Fk2HH+tw0L5PcWLHkwA4t60iXn16qDbIk/ze6jv2hDGdCdz7oYQeCE55CCHOHMJWYfN3jcL4s0qv8/u6Ka1FVkV7iMmro7ChThoV/5snI4Ljf2wKqgHH7TfNaCfpU0WvHAnTs8zhOrGScSrtb mircea@master-roshi", "ssh-dss AAAAB3NzaC1kc3MAAACBAIiXYWuYhY7Wyd3E4JYpT/ilm7Q6G/wLlxPgbx4BWRQL8SbVvP/DVLv+1/4aHk4SvPJmYbcGQbx0tz0pHa3YLmGieLQLVZo44K3IvUXaLf7noUKogwDdGVJ6NRxCROo/uUaU9m8mm4vEQkPgkmrRzNHq1B56AvPmTrRFJc5xPewXAAAAFQDiqOM9bw03MbT4jf59lFSlexRiiQAAAIA0ZJ2dNV9IzeWNexsoQfSz2ABx3INnFvjUAMlWZQtIXAqwUhKAvQBFi6vTpWL9nK4nCUvENyHIgX8S8eG2F1+pRFUHILajiJWkb0+0hGXIgE+rtdsdSCbMzggTOnhr4nMuUqedtas7uwGgOkNoEtsC6uSe0iSf3QbGjyrp8423lAAAAIAR0X7xTJCnV25i14SfelgVztVCl44C/E2A9+EBBR2Dp4CEg3cbstf6Z768S2TxOMpWTr7c81zKI6Ij7qLWxiUlTO7yS3ikiBOpEHbihGeHLn01rHcuGh9+SmOFKeIoq6WJHfZQInO/++WLTtwaZgZzIa1mw18SiFlHFeIX0MfYcA== mircea@master-roshi", "ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBFAvVeQxnECAx/HcSfcLdNqCqsjdm64QfVdxBT15pmvzw4eV2RWibqZTfCNbybM5FsB0/jEb2EH2JND4GCBzxUg= mircea@master-roshi"], "level": 0}, u'root': {u'level': 15, u'password': u'$1$LiP5a8I9$feewr345W3H2342', u'sshkeys': []}}
```
dbarrosop: the bits you added are not valid JSON, that's why the CI is broken.

sincerywaing: right... was just copy pasting... 
fixed in recent commit 32e5661

napalm/junos/junos.py
sincerywaing: right... was just copy pasting...
fixed in recent commit 32e5661

test/junos/mocked_data/test_get_users/normal/expected_result.json
sincerywaing: is that because what I returned is unicode?

sincerywaing: I removed `u` in the expected result.

dbarrosop:
Tests are still broken:

```
level:
model_class: <type 'int'>
data_class: <type 'unicode'>
```

That means you are return the `level` of a user as a `unicode` object and it should be an `int`.

mirceaulinic:
> I'd avoid doing that. It's just a user and the privilege level should give away the information.

I understand, and it's easier to introduce this change, but it's weak because:

- Privilege level: there can be other users having the same privilege level (15). We can eventually have a dedicated privilege level (0? 16? 100?) that is only for the root user.
- The username is not platform agnostic. As I previously mentioned, it is called `root` on Junos, but `admin` on other platforms, then you need to treat the platforms differently, which is in opposition to NAPALM principles.

So shall we assign a different privilege number for the root user? I think that would be the easiest change to introduce right now.

dbarrosop:
> We can eventually have a dedicated privilege level (0? 16? 100?) that is only for the root user.

I think this is the best option.

> The username is not platform agnostic. As I previously mentioned, it is called root on Junos, but admin on other platforms

I understand but this would be like standardizing interface names. It's kind of nice but terrible from a user point of view. Imagine you are trying to login as an admin user for a particular use case and we are returning a `napalm_define_agnostic_root_user` string instead of "root" or "admin", I would have to include in my code something like:

```
if username == napalm_define_agnostic_root_user and device_os in ["junos", "ios"]:
    login_with = "admin"
elif username == napalm_define_agnostic_root_user and device_os in ["eos"]:
    login_with = "root"
# maybe some other cases for some other platforms
else:
    login_with = userename
```
So masquerading the name gives you more trouble than anything. I think the vendor agnostic thing to do here is either:

1. mark the user as an admin with a special privilege level as you suggested
2. have a dedicated method to return only admin users

sincerywaing:
I’m fine with assigning a unique previledge to ‘root’ or ‘admin’. However it sounds like we are doing this just to differentiate the root user as it’s special (I know it is!) However in most (or all?) use cases, it’s just a super user with name ‘root’ or ‘admin’. so why not using our mapping to privilege 15? 
@dbarrosop I’ll try to fix the test issue after I come back from vacation next week.

dbarrosop:
So three things:

1. It's a special user you can't remove for instance. So it's good to be able to easily differentiate it without having to resort to string matching.
2. Looks like CI is still broken.
3. Provide a test case with your problem or it will be hard to help to look into it :P

mirceaulinic:
Hi @sincerywaing - did you see @dbarrosop's comments above? He already replied, but just to make sure the questions are answered:

> However it sounds like we are doing this just to differentiate the root user as it’s special (I know it is!) However in most (or all?) use cases, it’s just a super user with name ‘root’ or ‘admin’. so why not using our mapping to privilege 15?

As David already mentioned, "It's a special user you can't remove for instance". You username can be removed, and it can also have the privilege level 15. If we assign the same level to your username and the root, then there wouldn't be any other way to distinguish between the root and the other users. Does this answer your question?

sincerywaing:
@dbarrosop @mirceaulinic Should we give it a priviledge 20? I can change to 20 for sure.
Also double checked the CI but still can’t understand where it breaks...
What I can say is that I tried same code on real devices of srx and ex series all work.

mirceaulinic:
> Should we give it a priviledge 20? I can change to 20 for sure.

Yes @sincerywaing 20 is fine.

mirceaulinic:
@sincerywaing the tests are currently failing as the `level` key must be int. See https://travis-ci.org/napalm-automation/napalm/jobs/332751578#L1395

```
test/junos/test_getters.py::TestGetter::test_get_users[normal] <- napalm/base/test/getters.py key: level
model_class: <type 'int'>
data_class: <type 'unicode'>
key: password
model_class: <type 'unicode'>
data_class: <type 'unicode'>
FAILED [ 77%]
```

Thanks!

sincerywaing:
@mirceaulinic that’s what confuses me - I do return int in the code...

ktbyers:
@sincerywaing Are you still working on this?

sincerywaing:
@ktbyers I actually has been using this module for a long time in my custom_napalm module.
It's just I can't get Travis passed annoyed me...

ktbyers:
I updated this PR to fix the testing in the referenced PR.

PR #609, opened 2018-01-03
Interface is always enabled, unless explicitly specified in config

Ichabond:
An interface should always be considered enabled, unless the `shutdown` stanza is used in the configuration. If that’s the case, the XML API will set the state as `IM_STATE_ADMINDOWN`, opposed to `IM_STATE_DOWN` when the port is enabled, but there is no physical connectivity.


dbarrosop:
Huh, interesting this bug has been here for so long... would you mind fixing the CI?

Ichabond:
@dbarrosop All done :)

PR #610, opened 2018-01-04
Add napalm logo (previously removed probably during reunification)

mirceaulinic:
It's also missing from the README & the organisation icon.


dbarrosop:
I see, worse than Meltfown and Spectre

PR #621, opened 2018-01-11
Junos fix apply-groups in get_bgp_config()

ckishimo:
In Junos the `apply-groups` was never updated when using `get_bgp_config()`. I ignore if that was broken at one point as the existing test case have a `group` applied in the configuration

Anyway this PR is fixing the `apply-groups`. Another test case was added

The thing is when performing the `get()` operation, it's using by default the options `groups` and `inherit`. In that case the `apply-groups` are not reported but all the configuration inherited from the groups is reported which is needed!

In order to get the `apply-groups` being applied, we need to call `get()` without the `inherit` option. Something like:
```
options = {
                'groups': 'groups',
                 # 'inherit': 'inherit'
          }
bgp.get(options=options)
```
in that case all configuration inherited from a group will not be reported, so that's the reason why we need to call `get()` twice. 

Not sure I'm making myself clear :smile:  ... pls let me know I can post an example if needed... Thanks


mirceaulinic:
Hi @ckishimo. I'm not sure I understand the context:

> In Junos the apply-groups was never updated when using get_bgp_config(). I ignore if that was broken at one point as the existing test case have a group applied in the configuration

Are you saying that when a groups is applied, the configuration XML is different? Can you also paste some examples please?

ckishimo:
Hi @mirceaulinic 

With the following junos configuration:
```
carles@vmx# show groups    
test3 {
    protocols {
        bgp {
            group <*> {
                multipath;
            }
        }
    }
}
test2 {
    protocols {
        bgp {
            group <*> {
                multihop;
            }
        }
    }
}
[edit]
carles@vmx# show protocols bgp 
local-as 65001;
group internal {
    apply-groups [ test2 test3 ];
    type internal;
    neighbor 10.10.10.1 {
        export [ nhs static ];
    }
}
```
this is the output I got:
```
{
    "internal": {
        "apply_groups": [],
        "description": "",
        "export_policy": "",
        "import_policy": "",
        "local_address": "",
        "local_as": 0,
        "multihop_ttl": 64,
        "multipath": true,
        "neighbors": {
            "10.10.10.1": {
                "authentication_key": "",
                "description": "",
                "export_policy": "nhs static",
                "import_policy": "",
                "local_address": "",
                "local_as": 0,
                "nhs": false,
                "prefix_limit": {},
                "remote_as": 0,
            }
        },
        "prefix_limit": {},
        "remote_as": 0,
        "remove_private_as": false,
        "type": "internal"
    }
}
```
everything is there except the `apply-groups`. I'm running Junos 14.1R1.10

mirceaulinic:
Bump @ckishimo

dbarrosop:
> The "apply-groups" seems to be of type list of chars. In the existing mocked data it's already like this. is that correct ?

That looks like a bug to me :S

> As the multipath is configured, the test result should be true.

I am a bit confused here. Is it a problem with the testing framework or with the test itself? If it's with the test you can fix the expected result here:

https://github.com/napalm-automation/napalm/tree/develop/test/junos/mocked_data/test_get_bgp_config

But you might know that already. 

If it's a problem with the testing framework, would you care to elaborate more and provide some example? Your tests seem to be failing only on py3 so I guess you didn't push the code.

ckishimo:
@dbarrosop Ok, I can fix that and the mocked data 

Regarding the second issue, I would say it's a problem with the testing framework. If I'm not mistaken It is only taking into account the commands configured explicitly, and it is not taking into account the configuration **inherited** via the `apply-groups` command like in the example I posted where `multipath` is configured via a `group`. I´ll try to elaborate on the example I posted already if still not clear or we can discuss within slack. Thanks

mirceaulinic:
Yes, that's correct @ckishimo: for the moment, the testing framework only works with a limited set of arguments you can send to the method.
I think you should open an issues and discuss this separately, as that's an internal issues (not strictly related to this bug fix).
Regarding the mocked data: yes, please fix that, it's certainly a bug that should be fixed. Thanks!

dbarrosop:
> and it is not taking into account the configuration inherited via the apply-groups

Well, in my opinion that's the correct behavior. If you are using `apply-groups` to configure certain parameters those parameters don't belong to the child objects, they belong to the apply-group. You are getting the BGP config so you get what you have configured there. If we were talking about "state" I'd agree with you but given we are talking about "config" only configuration directly under the objects should apply.

ckishimo:
@dbarrosop fair point.. so I believe there is nothing left here. To sum up, this PR fixes the `apply-groups` which was never updated, and also it fixes the mocked data from `list of char` to `str`. Thanks

ckishimo:
@dbarrosop I just realized that without this PR all commands configured via `apply-groups` are indeed reported. Example is the `multipath` and `multihop` not explicitly configured but inherited via the `apply-groups`. I hope that is ok....
```
group internal {
    apply-groups [ test1 test2 ];
    type internal;
    neighbor 192.168.2.17 {
        export nhs;
    }
}
```
wthout this PR the `get_bgp_config()` is reporting `multipath` and `multihop` enabled despite they are not explicitly configured, unless I did something wrong... 
```
    "internal": {
        "apply_groups": [],
        "description": "",
        "export_policy": "",
        "import_policy": "",
        "local_address": "",
        "local_as": 0,
        "multihop_ttl": 64,
        "multipath": true,
        "neighbors": {
            "192.168.2.17": {
                "authentication_key": "",
                "description": "",
                "export_policy": "nhs",
                "import_policy": "",
                "local_address": "",
                "local_as": 0,
                "nhs": false,
                "prefix_limit": {},
                "remote_as": 0,
            }
        },
        "prefix_limit": {},
        "remote_as": 0,
        "remove_private_as": false,
        "type": "internal"
```

dbarrosop:
> this PR fixes the apply-groups which was never updated

I am a bit confused. I thought we agreed apply-groups don't belong to this method.

ktbyers:
Is this still being actively worked on?

ckishimo:
@ktbyers according to the last comment from @dbarrosop the bgp config should not be updated based on the `apply-groups`

I would suggest to close this issue and then I can create 2 different PRs. One to fix the bug on the existing test case (type list of chars) and another one to have at least the `apply-groups` listed but without modifying the configuration

ktbyers:
Closing as per comment by @ckishimo

Issue #579, opened 2017-12-07
Salt Proxy minion drops the connection with XRv 9000 after 1 minute


adampav:
Hello, i noticed this problem while experimenting on an XRv 9000.
After approximately one minute of normal operation the proxy minion is not able to interact with the specific device.
I tested the same on an IOSXE device and i didn't face such problems.
I attach some DEBUG log lines
2017-12-02 13:28:44,563 [salt.utils.lazy                                            ][DEBUG   ][20954] LazyLoaded status.proxy_reconnect
2017-12-02 13:28:44,564 [netmiko                                                    ][DEBUG   ][20954] Sending the NULL byte
2017-12-02 13:28:44,564 [netmiko                                                    ][DEBUG   ][20954] write_channel: 
2017-12-02 13:28:44,564 [/usr/lib/python2.7/dist-packages/salt/proxy/napalm.pyc     ][DEBUG   ][20954] Is xrv1 still alive? Yes.

As soon as those lines above appear i am not able to interact with the device.

2017-12-02 13:28:53,163 [salt.minion                                                ][INFO    ][20954] Starting a new job with PID 20954
2017-12-02 13:28:53,364 [netmiko                                                    ][DEBUG   ][20954] read_channel: 
2017-12-02 13:28:53,365 [netmiko                                                    ][DEBUG   ][20954] write_channel: <?xml version="1.0" encoding="UTF-8"?><Request MajorVersion="1" MinorVersion="0"><CLI><Exec>show ver</Exec></CLI></Request>
2017-12-02 13:28:53,365 [netmiko                                                    ][DEBUG   ][20954] read_channel: 
2017-12-02 13:28:53,566 [netmiko                                                    ][DEBUG   ][20954] read_channel: 
2017-12-02 13:28:53,766 [netmiko                                                    ][DEBUG   ][20954] read_channel: 

workaround as suggested by @mirceaulinic:  i have set the always_connected flag to false.

mirceaulinic:
That's unfortunate. I've noticed the same some time ago, but I forgot to submit the report.

Apparently our method `is_alive` does more harm than good, as besides checking the state of the SSH connection we also send the NULL byte, which destroys the connection: https://github.com/ktbyers/netmiko/blob/master/netmiko/base_connection.py#L248
[...]

PR #623, opened 2018-01-12
IOS-XR: return the is_alive flag for the SSH channel

mirceaulinic:
As a temporary fix for #579


napalm/iosxr/iosxr.py
```diff
@@ -106,10 +106,10 @@ def close(self):
...
     def is_alive(self):
         """Returns a flag with the state of the connection."""
-        # Simply returns the flag from Netmiko
-        return {
-            'is_alive': self.device.device.is_alive()
-        }
+        if self.device is None:
+            return {'is_alive': False}
+        # Simply returns the flag from pyIOSXR
+        return {'is_alive': self.device.is_alive()}
```
ktbyers: In `__init__` method should we just set:

self.device = None

and then not create the IOSXR object until the open() call? In other words, move lines 90 - 96 to be in the open() method.

Also should the close() method be setting `self.device = None`. I looked in the NAPALM IOS driver and in `__init__` I am setting self.device = None (and don't create the Netmiko object until open() call). I am not doing the self.close though (i.e. I think we should be setting self.device = None in close).

What do you think on this?

dbarrosop: I don't think it shouldn't matter whether we put do the `self.device = None` or not as long as the API behaves as documented. Would be nice to be consistent amongst devices that use netmiko though.

ktbyers:
+1

PR #624, opened 2018-01-13
Fix NX-OS not saving config to startup-config

ktbyers:
Also add structure to share code between NX-API and NX-OS SSH code.


ktbyers:
Tested both a merge and a replace operation including startup config being updated using both NX-API and SSH.

PR #625, opened 2018-01-16
Fix NX-OS rollback bug

ktbyers:


ktbyers:
Will rebase and resubmit.

PR #626, opened 2018-01-16
Fixing NX-OS rollback bug

ktbyers:
`self._save()` method referenced in rollback doesn't exist any longer. Also make code more consistent between SSH and NX-API


ktbyers:
Tested rollback on virtual NX-OS device using both SSH and NX-API.

ktbyers:
restarted tests...as Travis was having issues on the their side.

mirceaulinic (approved):

PR #629, opened 2018-01-21
junos implementation of commit confirm

sdanelli:
Adds commit confirm for junos. 
Procedure:
if `confirmed` is passed, 
-generates random string to add to the commit log, then it can be used to find if pending commit is happening. 
-send `confirm=<whatever value was sent>, message=<napalm_confirm_random_string` (junos device handles passing None) 
if successful, random string is stored in local private object attribute. 

`has_pending_commit_confirm`
Grabs the commit history, and if the last commit has the stored string + rollback, then we know there is a rollback happening (do we want to catch *any* rollback or the one we generated?)

`commit_confirm` 
checks if pending confirm and if True commits without confirm.

`revert_commit_confirm`

checks if pending confirm and if True, does a rollback to the last committed configuration and commits it.


test/junos/test_confirm_commit.py
```diff
@@ -0,0 +1,119 @@
+"""Test confirm commit calls and results"""
```
dbarrosop: This is **very** interesting. Would you mind looking into generalizing this set of tests? Configuration related parameters are here: https://github.com/napalm-automation/napalm/blob/develop/napalm/base/test/base.py

The goal is to make sure all drivers test the same workflow to ensure they behave consistently. So it would be great if the tests were defined in a common file and each driver would just have to define the mocked data.

napalm/junos/junos.py
```diff
@@ -235,14 +237,75 @@ def compare_config(self):
...
-        if confirmed is not None:
-            raise NotImplementedError
+        """
+        Commit configuration.
+
+        if confirmed is passed, random string is generated for the commit log
+        """
+        pending_commit_string = None
+        if confirmed:
+            pending_commit_string = str(uuid.uuid4()).split('-')[0]
```
mirceaulinic: I find this strange. Why are you doing this?

napalm/junos/junos.py
```diff
@@ -235,14 +237,75 @@ def compare_config(self):
...
+            if message:
+                message = '{}_napalm_confirm_{}'.format(message, pending_commit_string)
```
mirceaulinic: I don't think it's a good idea to commit with a different message than the user requested.

napalm/junos/junos.py
```diff
@@ -235,14 +237,75 @@ def compare_config(self):
...
+            else:
+                message = 'napalm_confirm_{}'.format(pending_commit_string)
```
mirceaulinic: Also, I think we shouldn't commit a gratuitous message if you didn't request one.

test/junos/test_confirm_commit.py
sdanelli: Sure, let me look and see what I can do.

napalm/junos/junos.py
```diff
@@ -235,14 +237,75 @@ def compare_config(self):
...
+
+        Will send the commit even without a pending
+        """
+        if self.has_pending_commit_confirm:
+            self.commit_config()
+            return
+        log.warning('no pending commit to confirm on %s', self.hostname)
+
+    @property
+    def has_pending_commit_confirm(self):
```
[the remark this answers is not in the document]
sdanelli: Relying on an object attribute for a non locked, non atomic transaction seems brittle to me.

napalm/junos/junos.py
sdanelli: So I know the pending commit is mine. I can remove this and rely on the 'rollback' string of the commit message, however, I won't know for sure that rollback is from our `commit confirm`.

napalm/junos/junos.py
sdanelli: Justification above, don't see a huge deal but noted.

ktbyers:
FYI, I did some work on this earlier this weekend (prior to your PR). Here is the naming of the methods:

```
    def commit_config(self, confirmed=None, message=None):
        """  
        Commits the changes requested by the method load_replace_candidate or load_merge_candidate.
        """
        raise NotImplementedError

    def commit_confirm(self):
        """Confirm pending commit."""
        raise NotImplementedError

    def commit_confirm_revert(self):
        """Abort pending commit confirm immediately and rollback to previous state"""
        raise NotImplementedError

    def has_pending_commit(self):
        """  
        :return A boolean indicating whether any pending commit confirms exist.
        """
        raise NotImplementedError
```

So we will need to make things consistent.

test/junos/test_confirm_commit.py
ktbyers: @sdanelli No worries...I just didn't want you doing work that we might not keep.

sdanelli: @ktbyers thanks, FWIW it explains the approach I took for Junos on this PR, but I agree, if you are testing against real devices, this seems like a waste.

sdanelli:
@ktbyers @dbarrosop @mirceaulinic Just making sure this is still alive. Is there something left for me to do here?

ktbyers:
@sdanelli Yes, I was going to integrate what you did into this branch in some way.

napalm/junos/junos.py
```diff
@@ -235,14 +236,70 @@ def compare_config(self):
             return diff.strip()
 
     def commit_config(self, confirmed=None, message=None):
-        """Commit configuration."""
+        """
+        Commit configuration.
+
+        if confirmed is passed, random string is generated for the commit log
```
dbarrosop: How so? Is this something done by pyez?

napalm/junos/junos.py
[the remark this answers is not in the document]
dbarrosop: I guess it depends if we want to return if there is a confirm pending in the system or by napalm. So I guess that's something we should agree on first and add it to documentation as a note.

test/junos/test_confirm_commit.py
dbarrosop: Yeah, let's hold onto this.

napalm/junos/junos.py
mirceaulinic: I think this refers to my comment above regarding `pending_commit_string = str(uuid.uuid4()).split('-')[0]` which I still consider we shouldn't do that. But this again depends on what we agree whether the commit revert should be limited to napalm or the broad system.

napalm/junos/junos.py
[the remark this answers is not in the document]
mirceaulinic: I think we cannot realistically aim to return if there's a confirm pending in the system, but limit to napalm only: it is doable for Junos, but for the other platforms (particularly IOS and NXOS) it might be impossible and we'd aim way too far. I would put the following example

Junos
------

1. Commit confirmed some changes
2. Another user from the CLI commits some changes (bad!!!)
3. You commit is not confirmed and it will revert. With your current implementation it would revert *only* your changes. Which I agree is the right thing to do.

IOS/Other stupid platform
--------------------------

1. Commit confirmed some changes
2. Another user from the CLI commits some changes (bad!!!)
3. I don't think it's possible to revert *only* your changes, but instead we might revert the changes made from a different process / session as well. If that's the case (@ktbyers to confirm here please), in order to keep the drivers in-line, we might need to make a compromise.

napalm/junos/junos.py
dbarrosop: We shouldn't create a random commit message if the user doesn't specify one. Basically, `compare_config()` should behave exactly the same way before and after this PR.

napalm/junos/junos.py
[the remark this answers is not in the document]
dbarrosop: Not sure the problem you are describing is really relevant or not. I understand it's a shitty problem for IOS but the question is if `commit_confirm`, `commit_confirm_revert` and `has_pending_commit` should work at the system level or not:

1. If it works at the system level it is the user responsibility to make sure it's done under controlled circumstances because in that case those methods are just going to blindly run the commands for those operations in the system.
2. If it works under napalm only same as before but we do a check like `self.pending_commit`.

The problem of option 2 is that this is not going to work with ansible or if your connection drops as you'd lose that variable.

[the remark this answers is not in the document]
mirceaulinic: Are we actually able to implement (1) properly on IOS/NXOS?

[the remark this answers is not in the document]
dbarrosop: That I don't know and I will defer to our resident expert @ktbyers but if we can't I think this feature is going to be useless for most of our users and dangerous for other. I'd certainly don't use it if I can't recover from a connection drop and can confirm it.

napalm/junos/junos.py
sdanelli: @dbarrosop sure, I removed that part of the code, will also update the docstring. I am not sure what you mean with `compare_config()` however. It always behaved the same way regardless.

napalm/junos/junos.py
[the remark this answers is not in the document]
ktbyers: I will try to review this thread sometime today. I believe I had a reasonable solution for IOS, NX-OS, and EOS (with EOS I think being the most problematic IIRC). 

I need to read your comments though to see how the solutions addresses the points brought up.

napalm/junos/junos.py
dbarrosop: Yeah, sorry, I meant `commit_config`.

napalm/junos/junos.py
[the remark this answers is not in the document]
dbarrosop: EOS should have native support for the commit confirm operations I think.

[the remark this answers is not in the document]
ktbyers: Just adding comments as I read through this thread:
1. IMO it should be system commit-confirm not a NAPALM specific commit-confirm
2. It is already done on IOS (i.e. this branch has a solution implemented in IOS that does a system commit-confirm).
3. I will have to check the solution I had in mind for NX-OS, but I am pretty sure I came to the conclusion a reasonable solution could be implemented on NX-OS.
4. IIRC the EOS solution requires a certain version of EOS which I have noted somewhere (or can figure it out again).
5. I agree with @dbarrosop point here that we should require this (I think this is consistent with what is realistically already required).

```
If it works at the system level it is the user responsibility to make sure 
it's done under controlled circumstances because in that case those 
methods are just going to blindly run the commands for those operations
 in the system.
```

6. The solution should survive a NAPALM connection drop / re-establishment.

ktbyers:
I am going to close this. Right now there is no active work going on with commit confirmed. It has been on hold for a long time so I don't think it makes sense to keep this open :-(

PR #632, opened 2018-01-24
Update get_interfaces_ip to deal with unnumbered interfaces

### LINKED ISSUE #630: get_interfaces_ip for NXOS does not handle unnumbered interfaces
When an interface is an unnumbered interface (https://www.cisco.com/c/en/us/support/docs/ip/hot-standby-router-protocol-hsrp/13786-20.html#unnumbered_iface), napalm will raise an exception: 
```netaddr.core.AddrFormatError: failed to detect a valid IP address from None```
This is due to https://github.com/napalm-automation/napalm/blob/develop/napalm/nxos/nxos.py#L700, which passes None to IPAddress.

Either we should skip an interface that does not have a prefix assigned, and is designated unnumbered, or the prefix from the interface it is borrowing from should be used.

Ichabond:
Unnumbered interfaces are listed in `show ip interface`, but have no `prefix` key. Take precautions for this, and inherit the IP of the interface it is borrowing from.
Take the same precaution for ipv6. This fixes #630.


mirceaulinic:
Bump @ktbyers - does this PR look good to you?

mirceaulinic:
Bump @ktbyers - these changes look good. Let me know if you have any objections against getting this in.

ktbyers:
@mirceaulinic I haven't reviewed it, but if you have reviewed it and it looks good...then I would move forward and merge it.

Issue #633, opened 2018-01-25
nxos_ssh.py parsing for arp table dies if flags present in command output


rbcollins123:
This line forces the arp table to only have at most 4 columns in output:
https://github.com/napalm-automation/napalm/blob/cb5fbd8de58c7f3971deacfd287c62e124155588/napalm/nxos_ssh/nxos_ssh.py#L1057

But there's a valid 5th column for flags that could be ignored, but shouldn't void parsing the result:
```
# sh ip arp

Flags: * - Adjacencies learnt on non-active FHRP router
       + - Adjacencies synced via CFSoE
       # - Adjacencies Throttled for Glean
       D - Static Adjacencies attached to down interface

IP ARP Table for context default
Total number of entries: 2302
Address         Age       MAC Address     Interface
172.20.10.82    00:02:03  f4cf.e2e0.cf41  Ethernet3/36
172.16.252.10   00:05:44  002c.c87d.743f  Ethernet1/5
172.16.253.2    00:02:54  2c5a.0ff0.457f  Ethernet1/7
172.20.110.6    00:09:11  002c.c87d.757f  Ethernet1/9
172.16.200.46   00:00:08  70df.2f07.20e4  Ethernet1/11
192.32.10.1     00:07:23  0000.0c07.ac0a  Vlan10
192.32.10.3     00:15:35  4055.3922.ccc1  Vlan10
192.6.125.22    00:11:32  00e0.8602.a5e9  Vlan125
192.6.125.23    00:09:39  00e0.8602.c3d0  Vlan125
192.6.125.25    00:06:01  001b.d502.a744  Vlan125
192.6.125.32    00:04:50  0050.569a.79dd  Vlan125          *
192.6.125.36    00:00:26  0024.81e1.e512  Vlan125          *
192.6.125.37    00:00:03  0025.b323.f35e  Vlan125          *
192.6.125.44    00:06:31  0050.56b2.003f  Vlan125
```

a simple `elif` to catch the case with 5 columns and parse the 5th column into _ or some unused var would suffice?

Issue #441, opened 2017-10-24
Parsing issue on get_mac_address_table for nxos_ssh


mirceaulinic:
From show mac address-table. Parser will choke on the extra space in the interface name

```
 500      0023.e940.4f83    dynamic     ~~~      F    F  vPC Peer-Link
  500      0025.b5b0.304e    dynamic     ~~~      F    F  vPC Peer-Link
```

Fix for this is to just preprocess 'vPC Peer-Link' this string and eliminate the space:

```
vPC-Peer-Link
```

---

Issue moved from https://github.com/napalm-automation/napalm-nxos/issues/140

PR #634, opened 2018-01-25
Fixes for CAM and ARP parsing issues in nxos_ssh.py

rbcollins123:
I happened to hit both of these today and saw an old issue open on one (#441) and created a new issue for discussion on the other (#633).  Attached are simple fixes for each that I patched into my local instance...

### DIFF
--- a/napalm/nxos_ssh/nxos_ssh.py
+++ b/napalm/nxos_ssh/nxos_ssh.py
@@ -1054,10 +1054,8 @@ class NXOSSSHDriver(NXOSDriverBase):
 
         arp_entries = arp_list[1].strip()
         for line in arp_entries.splitlines():
-            if len(line.split()) == 4:
-                address, age, mac, interface = line.split()
-            elif len(line.split()) == 5:
-                address, age, mac, interface, flag = line.split()
+            if len(line.split()) >= 4:
+                address, age, mac, interface = line.split()[:4]
             else:
                 raise ValueError("Unexpected output from: {}".format(line.split()))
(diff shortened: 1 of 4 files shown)


napalm/nxos_ssh/nxos_ssh.py
```diff
@@ -1056,6 +1056,8 @@ def get_arp_table(self):
         for line in arp_entries.splitlines():
             if len(line.split()) == 4:
                 address, age, mac, interface = line.split()
+            elif len(line.split()) == 5:
+                address, age, mac, interface, flag = line.split()
```
itdependsnetworks: Does it make sense to test if the flag is one of the four options?

rbcollins123: I do not see the point in doing that personally unless we plan to use that value for something. Currently it is just assigned and later garbage collected.  But, if there is a use case for collecting it and using it in the code later I could do that and add appropriate testing. What did you have in mind?

rbcollins123: Or is your concern more that down the road some other value could land in there as Cisco changes the output some day?

dbarrosop: Yeah, something like:

```
address, age, mac, interface = line.split()[:4]
```

Might be more resilient.

rbcollins123: Fair enough!  I’ll change it here in a bit

PR #642, opened 2018-02-07
NX-OS ping implementation using more shared common code

ktbyers:
This is a restructuring of the PR that @mirceaulinic submitted.


ktbyers:
@mirceaulinic I will still review this (unless you have objections) as this is still your code. I just made a `_send_command` wrapper so the identical code could be shared between NX-API and SSH.

PR #649, opened 2018-02-11
improved exceptions - added new base exception

cwoodfield:
Hi! I just had first go at writing a network automation script using napalm; I wanted to wrap my execution block in a try/except block, catching napalm-specific exceptions so that it would not exit the script, but to exit on other exceptions. In order to do this, I would have had to catch almost every listed exception class defined in base/exceptions.py individually.

I wrote a simple patch here to create a new generic NapalmException class, which inherits from Exception, and then modified each other exception class to subclass NapalmException instead. This way, I can list NapalmException as the only exception I need to catch, and then interrogate further for specific exception types if needed.

I've run the docs/test.sh script and all is good; let me know if there's further testing I should be doing before submitting a pull request. Thanks!

### DIFF
--- a/napalm/base/exceptions.py
+++ b/napalm/base/exceptions.py
@@ -16,6 +16,7 @@
 from __future__ import print_function
 from __future__ import unicode_literals
 
+
 class NapalmException(Exception):
     '''
     Base Exception Class.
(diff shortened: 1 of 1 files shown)


dbarrosop:
This is completely reasonable and I am actually quite surprised we haven’t done this sooner... Let’s wait to see what others think before merging though.

Issue #652, opened 2018-02-13
napalm referencing attribute that is deprecated in py3.6 (maybe py3)


sdanelli:
When `ConnectTimeoutError` is attempted in "napalm/napalm/junos/junos.py" line 109, it references an attribute called `message` which has been deprecated. 

In python2.7:
```
In [1]: e = Exception('boo')

In [2]: e.message
/Users/sdanelli/.virtualenvs/containermanager/bin/ipython:1: DeprecationWarning: BaseException.message has been deprecated as of Python 2.6
  #!/Users/sdanelli/.virtualenvs/containermanager/bin/python
Out[2]: 'boo'
```

In python3.6:

```
In [6]: e = Exception('boo')

In [7]: e.message
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-7-7e9b9f9f4c6d> in <module>()
----> 1 e.message

AttributeError: 'Exception' object has no attribute 'message'
```
Which means is raising an `AttributeError` instead of a `ConnectionException`:

More details:
```
...
    device.open()
  File "/Users/sdanelli/devwork/napalm/napalm/junos/junos.py", line 109, in open
    raise ConnectionException(cte.message)
AttributeError: 'ConnectTimeoutError' object has no attribute 'message'
```

sdanelli:
fix might to just cast to str if interested in the details: `raise ConnectionException(str(cte))`

dbarrosop:
Good catch, would you mind opening a PR yourself?

sdanelli:
@dbarrosop sure, #657

ktbyers:
Updated fix here:  https://github.com/napalm-automation/napalm/pull/726

dbarrosop:
Fixed via #726

PR #657, opened 2018-02-18
replacing `message` with cast to `str`

sdanelli:
As noted [here ](https://github.com/napalm-automation/napalm/issues/652#issuecomment-366434168)

le old perl to the rescue `find . -type f -a -name '*py' -exec egrep -l 'raise.*\.message' {} + | xargs -I{} perl -wlpi -e 's/(raise.*?\()(\w+)\.message/$1str($2)/g' {}`

### DIFF
--- a/napalm/base/helpers.py
+++ b/napalm/base/helpers.py
@@ -81,9 +81,9 @@ def load_template(cls, template_name, template_source=None, template_path=None,
         )
     except (jinja2.exceptions.UndefinedError, jinja2.exceptions.TemplateSyntaxError) as jinjaerr:
         raise napalm.base.exceptions.TemplateRenderException(
-            "Unable to render the Jinja config template {template_name}: {error}".format(
+            "Unable to render the Jinja config template {template_name}: {error!s}".format(
                 template_name=template_name,
-                error=jinjaerr.message
+                error=jinjaerr
             )
         )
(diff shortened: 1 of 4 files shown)


ktbyers:
We should cast to str as that will be a unicode in Python3 and a ASCII string in Python2.

Instead we should do:

```
from napalm.base.utils import py23_compat

py23_compat.text_type(obj)
```

sdanelli:
@ktbyers So, no problem doing this if you really want to, however can you tell me why? `.message` was a `str` and not unicode (in py2 anyways) AFAICT, and is every `str` in the codebase unicode?

ktbyers:
Yes, every string in the code-base is unicode (well should be anyways). Obviously, things outside of NAPALM might do something unexpected.

Typecasting things as ASCII-strings has a way of causing problems latter on. 

Also I don't see a good reason for having it be an ASCII-string in PY2 and a Unicode String in PY3 (though there is a good chance it wouldn't cause any problem in this case...so it is mainly for consistency i.e. all of NAPALMs strings are unicode).

sdanelli:
@ktbyers sounds good, if its consistent with the rest the code then it makes sense.

ktbyers:
This PR was replaced by this one:

https://github.com/napalm-automation/napalm/pull/726

PR #662, opened 2018-02-28
get_mac_address_table fix including unit tests

ktbyers:
.


dbarrosop:
I think we should start considering subdrivers. This is becoming a bit insane...

ktbyers:
@dbarrosop Unfortunately, I think sub-drivers would be even more insane (i.e. you would have to try to determine which output corresponded to which IOS version which I think would be close to impossible).

But agreed the code is getting really ugly and hard to manage.

PR #670, opened 2018-03-09
Adding Junos commit comment functionality

ipv6freely:
Adding a commit comment to the junos driver is incredibly simply since the PyEZ commit() function takes a comment natively.

I noticed that in some other branches it seems napalm devs want to use "message" instead of "comment", so that's what I used. Have tested this locally already and works perfectly.

### DIFF
--- a/napalm/junos/junos.py
+++ b/napalm/junos/junos.py
@@ -235,9 +235,9 @@ class JunOSDriver(NetworkDriver):
         else:
             return diff.strip()
 
-    def commit_config(self):
+    def commit_config(self, message=None):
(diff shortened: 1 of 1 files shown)


ktbyers:
@ipv6freely It will have to be implemented on all of the core platforms (IOS, EOS, NX-OS, Junos, EOS). We have to try to maintain consistency across platforms (which is hard).

The Travis tests will require the method signature to be the same across all of the platforms.

ipv6freely:
Okay, scrap it then.

dbarrosop:
Linking  #671 for reference

PR #671, opened 2018-03-09
Adding commit comment functionality

ipv6freely:
Adding a commit comment to the junos driver is incredibly simply since the PyEZ commit() function takes a comment natively.

I noticed that in some other branches it seems napalm devs want to use "message" instead of "comment", so that's what I used. Have tested this locally already and works perfectly.

I also added the message=None argument to the commit_config() function for the other drivers to. Most don't do anything, but IOS-XR will raise an exception when the message is used because it hasn't yet been full implemented.


napalm/iosxr/iosxr.py
```diff
@@ -145,7 +145,9 @@ def compare_config(self):
         else:
             return self.device.compare_config().strip()
 
-    def commit_config(self):
+    def commit_config(self, message=None):
+        if message is not None:
+            raise ValueError('Commit message not yet implemented for IOS-XR')
```
dbarrosop: Could you change this to `NotImplementedException`?

napalm/iosxr/iosxr.py
```diff
@@ -145,7 +145,9 @@ def compare_config(self):
         else:
             return self.device.compare_config().strip()
 
-    def commit_config(self):
+    def commit_config(self, message=None):
```
dbarrosop: I think `message=""` as it also gives away the expected type.

napalm/iosxr/iosxr.py
dbarrosop: Would you also mind adding this same check to all the drivers?

dbarrosop:
@mirceaulinic @ktbyers I think it's fine we proceed with this PR with the minor changes I requested. What do you think?

napalm/iosxr/iosxr.py
ktbyers: @dbarrosop I don't have any objections to this, if this changed; then the exception would also need to be changed to:

```
if not message:
    raise NotImplementedException
```

ktbyers:
Yes, I think we should move forward with the changes you proposed. As this will give @ipv6freely what he needs to use this.

ipv6freely:
Sounds good, I'll make those changes.

napalm/iosxr/iosxr.py
ipv6freely: You would prefer `message=""` instead of `message=None` ?

ipv6freely:
I made the requested changes. I added `message=''` to all of the methods (using single quotes as that seems to be what is being used elsewhere) and also changed to `raise NotImplementedError` with a generic message of `Commit message not implemented for this platform`

I believe we're good to go at this point.

napalm/iosxr/iosxr.py
```diff
@@ -145,7 +145,9 @@ def compare_config(self):
         else:
             return self.device.compare_config().strip()
 
-    def commit_config(self):
+    def commit_config(self, message=''):
+        if not message:
+            raise NotImplementedError('Commit message not implemented for this platform')
```
mirceaulinic: Why not? It is supported by IOSXR and the underlying library: https://github.com/fooelisa/pyiosxr/blob/master/pyIOSXR/iosxr.py#L577

All you have to do is:

```python
         if self.replace:
             self.device.commit_replace_config(comment=message)
         else:
             self.device.commit_config(comment=message)
````

File: napalm/nxos/nxos.py [hunk 4]
```diff
@@ -44,7 +44,9 @@
 
 class NXOSDriverBase(NetworkDriver):
     """Common code shared between nx-api and nxos_ssh."""
-    def commit_config(self):
+    def commit_config(self, message=''):
+        if not message:
+            raise NotImplementedError('Commit message not implemented for this platform')
```
mirceaulinic: I think we should do the same for `nxos_ssh`? @ktbyers

ipv6freely: There is no `commit_config()` in `nxos_ssh`.

napalm/iosxr/iosxr.py
ipv6freely: Because @dbarrosop and @ktbyers said to do it this way until somebody who knows IOS-XR can add that functionality.

ipv6freely:
It just occured to me that the logic is backwards. it should be `if message raise NotImplementedError` because the exception should be raised if the user _has_ specified a message

napalm/iosxr/iosxr.py
ipv6freely: I just added it.

ipv6freely:
I fixed the logic fail using `if message:`, and added support for IOS-XR per @mirceaulinic 's request. 

I also changed the methods to have `message=None` as this seems to be the way it's done everywhere else and I think it makes it more consistent.

ipv6freely:
@mirceaulinic Do you have everything you need for this, now?

ktbyers:
I have some changes I am going to make on this PR. I will update it.

ktbyers:
I updated this PR here: https://github.com/napalm-automation/napalm/pull/701

PR #673, opened 2018-03-13
Unify netmiko argument processing across napalm drivers

ktbyers:
.


dbarrosop:
YES, PLEASE! :D

napalm/base/netmiko_helpers.py
```diff
@@ -0,0 +1,47 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+from __future__ import unicode_literals
+import inspect
+from netmiko import BaseConnection
+
+
+def netmiko_args(optional_args):
+    """Check for Netmiko arguments that were passed in as NAPALM optional arguments.
```
dbarrosop: Can you confirm the optional_args documentation (RTD) for netmiko based drivers are in synch with what this supports?

ktbyers: I will update the documentation after this is merged. I expect the documentation is wrong, but I am pretty sure it was wrong both before and after this.

This will make the documentation easier since it will be consistent across all three drivers and will support all Netmiko `__init__` arguments.

@dbarrosop

mirceaulinic (approved):

PR #675, opened 2018-03-16
Lldp neighbors detail bugfix

colsil:
This PR addresses inconsistencies with the xml-rpc calls required by junos on different devices to fetch detailed lldp information.

I'm simultaneously reaching out to juniper to see if we can get some clarification on what we should expect from various junos devices / models / code versions.

[I also cleaned up some failing PEP8 linting tests](https://github.com/napalm-automation/napalm/commit/03ef4a665aed23aad94c84cb0e83649b14e3ad26), but all changes there are whitespace only.


dbarrosop:
Awesome job! Thx!

PR #701, opened 2018-04-14
Updated PR commit_config with comment

ktbyers:
.


ktbyers:
The test failure is due to the PIP10 issue in pyiosxr library:

```
      File "/tmp/pip-install-HDfM5b/pyIOSXR/setup.py", line 21, in <module>
        from pip.req import parse_requirements
    ImportError: No module named req
```

I have tested it in my environment.

ktbyers:
Tested in my lab environment again Junos and IOS-XR verified the comment worked properly (using PY3.6

Also regression tested the standard unit tests using PY3.6 and PY2.7.

PR #704, opened 2018-04-20
Fix #703

### LINKED ISSUE #703: `base.clitools.cl_napalm:main` entrypoint still uses PIP internals
https://github.com/napalm-automation/napalm/blob/f45fc4f8e9108cacd76ce8f5cd9182df45129480/napalm/base/clitools/cl_napalm.py#L146

Should switch to `pkg_resources.find_distributions()` or similar.

bewing:
Use `pkg_resources.working_set` instead of `pip.get_installed_distributions()`


bewing:
Is there a specific reason we don't do coverage/testing of the clitools folder?

PR #717, opened 2018-04-27
Adding a new driver for DOS

mirceaulinic:
This is based on a previous idea of @dbarrosop to be able to show / test small things without actually requiring a network box to run against.

For the configuration management side it simply manipulates some random text, while getters only load some data from the tests (copied from Junos) then returns the object.

There are some minor issues, but submitting this to discuss before moving forward.

Usage:

```python
import napalm
d = napalm.get_network_driver('dummy')
dummy = d('', '', '')
dummy.load_merge_candidate(config='aaaa\nbbbb')
dummy.get_config()
```


mirceaulinic:
As discussed with @dbarrosop in private, there already is a driver with this exact scope:

```python
>>> import napalm
>>> d = napalm.get_network_driver('mock')
>>> d
<class 'napalm.base.mock.MockDriver'>
```

PR #719, opened 2018-04-28
Add test for sphinx documentation

ogenstad:
This PR adds an environment to test that the documentation in Sphinx is generated correctly.

In this first PR the `make html` test ignores warnings. The idea is to later change this and fail on warnings. There are currently a number of warnings generated when creating the documentation.

An example can be seen when looking at get_firewall_policies:
https://napalm.readthedocs.io/en/latest/base.html#napalm.base.base.NetworkDriver.get_firewall_policies

Compare the example of that function to get_interfaces:
https://napalm.readthedocs.io/en/latest/base.html#napalm.base.base.NetworkDriver.get_interfaces

If accepted I can look at those warnings and start to fix then and then set the test to fail on warnings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,16 @@ python:
   - 3.5
   - 3.6
 
+matrix:
+  include:
+      - python: 3.6
+        env: TOXENV=sphinx
+
 install:
+  - pip install tox
   - pip install tox-travis
   - pip install coveralls
+
 deploy:
   provider: pypi
   user: dbarroso
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,7 @@ matrix:
         env: TOXENV=sphinx
 
 install:
-  - pip install tox
-  - pip install tox-travis
-  - pip install coveralls
+  - pip install tox tox-travis coveralls
 
 deploy:
   provider: pypi
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,5 @@
+.PHONY: doctest
+doctest:
+	cd docs && make clean
+	cd docs && sphinx-build -b html -d _build/doctrees . _build/html
+	#cd docs && sphinx-build -W -b html -d _build/doctrees . _build/html
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py27,py34,py35,py36
+envlist = py27,py34,py35,py36,docs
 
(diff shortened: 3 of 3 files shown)


dbarrosop:
Can we make it so it's a different job? The `matrix` functionality of travis should let us do that I think

ktbyers:
Can we only do this `sphinx-build` for Travis-CI and not for Tox?

`sphinx-build` is basically impossible to build on Fedora-based machines (which includes the default Amazon AMI images). So this will make it impossible for me to use Tox in my local environment.

So it would be preferable to only have it coupled to Travis-CI and not to Tox.

ogenstad:
Now it's in a different job.

@ktbyers, David already merged a similar thing for Brigade. The Pandoc issue sounds weird, it's not as easy as just installing an rpm?

Perhaps a workaround could be to use this?

```bash
tox -e py27,py36
```

ktbyers:
Pandoc requires a Haskell compiler so that already puts it in a very edge case.

It doesn't look like we are using pandoc in NAPALM (correct me if I am wrong)...only in Brigade. So this shouldn't be an issue for napalm.

I gave up on getting Pandoc to work on Amazon AMI for Brigade. I probably only spent 30 minutes to an hour on it though. I didn't find an easy solution, but I could have missed it/not seen it. I will potentially circle back on it for Brigade.

.travis.yml
```diff
@@ -6,9 +6,16 @@ python:
   - 3.5
   - 3.6
 
+matrix:
+  include:
+      - python: 3.6
+        env: TOXENV=sphinx
+
 install:
+  - pip install tox
```
dbarrosop: can we just do one line for all three?

dbarrosop:
So you can use TOXENV to limit which tests you want to run:

```
export TOXENV=py27,py36
tox
```

Should run only those two jobs ignoring others. I think that's a better approach than having to complicate the CI.

ktbyers:
@dbarrosop @ogenstad Okay, that works...

```
export TOXENV=py27,py36
tox
```

And @ogenstad `tox -e` should also work.

ogenstad:
Changed the pip line, was that for speed or esthetical reasons?

Should we merge this now or should I start adding changes to the documentation within this PR? Those changes might require a bit more review.

dbarrosop:
Well, just because there is no reason to install every single thing in a different line :P

PR #681, opened 2018-03-20
Fix bug with get_interfaces_ip with multiple ipv6 addrs on interface

DiogoAndre:
Fixes #442

PR #720, opened 2018-04-28
Fix ipv6 test for nxos which causes RTD to fail

ogenstad:
The PR #681 introduced an error which caused the builds for the documentation to fail.
https://readthedocs.org/projects/napalm/builds/7059953/

This PR moves the tests to the correct directory and removed the unnecessary test code.

These things will be caught in the future with #719.


DiogoAndre:
Sorry for breaking things, and thanks for fixing it @ogenstad!

PR #722, opened 2018-05-02
fix getting ip interface from vrf for NXOS_SSH driver

darylturner:
IP interfaces outside of the 'default' routing instance are missed. This looks intentional from the code of the original developer but isn't consistent with the IOS and Junos drivers tested.

No changes to parsing logic were required.

### DIFF
--- a/napalm/nxos_ssh/nxos_ssh.py
+++ b/napalm/nxos_ssh/nxos_ssh.py
@@ -1140,8 +1140,8 @@ class NXOSSSHDriver(NXOSDriverBase):
         }
         """
         interfaces_ip = {}
-        ipv4_command = 'show ip interface vrf default'
-        ipv6_command = 'show ipv6 interface vrf default'
+        ipv4_command = 'show ip interface vrf all'
+        ipv6_command = 'show ipv6 interface vrf all'
         output_v4 = self.device.send_command(ipv4_command)
         output_v6 = self.device.send_command(ipv6_command)
 
--- a/test/nxos_ssh/mocked_data/test_get_interfaces_ip/normal/expected_result.json
+++ b/test/nxos_ssh/mocked_data/test_get_interfaces_ip/normal/expected_result.json
@@ -10,38 +10,45 @@
         "ipv4": {
             "2.2.2.2": {
                 "prefix_length": 27
-            }, 
+            },
             "3.3.3.3": {
                 "prefix_length": 25
             }
         }
-    }, 
+    },
     "Ethernet2/3": {
         "ipv4": {
             "4.4.4.4": {
                 "prefix_length": 16
             }
-        }, 
+        },
         "ipv6": {
             "fe80::2ec2:60ff:fe4f:feb2": {
                 "prefix_length": 64
-            }, 
+            },
             "2001:db8::1": {
                 "prefix_length": 10
             }
         }
-    }, 
+    },
     "Ethernet2/4": {
         "ipv6": {
             "fe80::2ec2:60ff:fe4f:feb2": {
                 "prefix_length": 64
-            }, 
+            },
             "2001:11:2233::a1": {
                 "prefix_length": 24
-            }, 
+            },
             "2001:cc11:22bb:0:2ec2:60ff:fe4f:feb2": {
                 "prefix_length": 64
             }
         }
-    } 
+    },
+    "mgmt0": {
+        "ipv4": {
+            "5.5.5.5": {
+                "prefix_length": 24
+            }
+        }
+    }
 }
--- a/test/nxos_ssh/mocked_data/test_get_interfaces_ip/normal/show_ip_interface_vrf_all.txt
+++ b/test/nxos_ssh/mocked_data/test_get_interfaces_ip/normal/show_ip_interface_vrf_all.txt
@@ -1,0 +1,119 @@
+IP Interface Status for VRF "default"
+Ethernet2/1, Interface status: protocol-up/link-up/admin-up, iod: 37,
+  IP address: 1.1.1.1, IP subnet: 1.1.1.0/24 route-preference: 0, tag: 0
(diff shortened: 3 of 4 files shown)


ktbyers:
@darylturner The returned data structure doesn't look correct. No key for the vrf for the default/global VRF. Or worded differently in one section the same level key refers to the VRF name; in the other it refers to interfaces.

Also looking at the unit test data, have we actually implemented VRFs on the other platforms? I only saw it on Junos and their I am not sure it wasn't in error.

Looking at the entire docstring in base, I suspect VRFs were never implemented:

```
 859     def get_interfaces_ip(self):
 860 
 861         """
 862         Returns all configured IP addresses on all interfaces as a dictionary of dictionaries.
 863         Keys of the main dictionary represent the name of the interface.
 864         Values of the main dictionary represent are dictionaries that may consist of two keys
 865         'ipv4' and 'ipv6' (one, both or none) which are themselvs dictionaries witht the IP
 866         addresses as keys.
 867         Each IP Address dictionary has the following keys:
 868             * prefix_length (int)
 869 
 870         Example::
 871 
 872             {
 873                 u'FastEthernet8': {
 874                     u'ipv4': {
 875                         u'10.66.43.169': {
 876                             'prefix_length': 22
 877                         }
 878                     }
 879                 },
 880                 u'Loopback555': {
 881                     u'ipv4': {
 882                         u'192.168.1.1': {
 883                             'prefix_length': 24
 884                         }
 885                     },
 886                     u'ipv6': {
 887                         u'1::1': {
 888                             'prefix_length': 64
 889                         },
 890                         u'2001:DB8:1::1': {
 891                             'prefix_length': 64
 892                         },
 893                         u'2::': {
 894                             'prefix_length': 64
 895                         },
 896                         u'FE80::3': {
 897                             'prefix_length': u'N/A'
 898                         }
 899                     }
 900                 },
 901                 u'Tunnel0': {
 902                     u'ipv4': {
 903                         u'10.63.100.9': {
 904                             'prefix_length': 24
 905                         }
 906                     }
 907                 }
 908             }
 909         """
```

ktbyers:
Adding @dbarrosop as I know he had communicated with you on this via Slack.

I checked the Cisco IOS napalm code, I don't think VRFs were ever intentionally implemented and I doubt they work properly.

I think this is an enhancement we would need to do for all the drivers.

ktbyers:
I deleted a couple of comments of mine as they were factually wrong. Checking again...

ktbyers:
Have we tested this on the other platforms and verified whether VRFs actually work on them? 

I see it also doesn't work on NX-API (looking at output test data). I would be surprised if it works on Cisco IOS since it was not intentionally designed to support VRFs (i.e. it would probably be an accident if it worked).

The definition in base.py is vague with respect to VRFs; historically, I have interpreted it the other way (i.e. the global VRF/default VRF and not all VRFs). 

Related to this, the data structure is possibly not right if we are supporting multiple VRFs as there no VRF references in the returned data structure.

ktbyers:
@ababson You are worried that the unit tests contain real IP addresses?  Just want to make sure I understand your concern.

darylturner:
The new behaviour is consistent with Junos and IOS drivers. The output doesn’t refer to VRFs at all (although I think this would be helpful). The behaviour with those drivers and what I interpreted from the base class docstring is to return all IP interfaces regardless of VRF.

ktbyers:
@darylturner You tested it on IOS and Junos with multiple VRFs and you get those interfaces returned in the get_interfaces_ip table? I suspect it was just luck on Cisco IOS that happened.

Or worded differently: `show ip interface` on Cisco IOS returns all VRFs by default?

I asked some questions in Slack as I am not confident 1)this was the intended behavior, 2)we are testing this properly on all platforms, 3)and I know it is broke also on NX-API NAPALM driver.

ababson:
@ktbyers First...my bad, I was in a hurry and didn't see this was a public repo and not my team's internal repo.  I really thought I was only commenting to my team.  That said...yes, I believe test and documentation ranges were created for a reason, and using someone else's IP addresses anywhere, even unit tests, is bad practice. It takes minimal effort to avoid doing it.  Unfortunately a lot of engineers, especially newbies, see examples like this and then due to the monkey-see-monkey-do effect, the next thing you know you've got someone else's public addresses in your production network because it was "easier" to type 1.1.1.1 and 1.1.1.2 for the gre tunnel addresses. :-/  By the way, I'm a fan of your contributions to automation and took your class a few years ago.  I'm just a bit of a purist when it comes to these things.

darylturner:
This is why I asked what was intended behaviour earlier and it seems open to interpretation 🙂.

ktbyers:
@ababson No worries.

@darylturner Yes, understood. I have asked David and Mirceau about it (and I saw David's original comments to you about it). That is the problem with it being vague...I think that was not the intent; David thinks it was the intent. I think Mirceau actually implemented quite a bit of it so I will see what he says.

If that is the intent, I think we need to verify that works on all the core platforms (it sounds like you have probably verified it on Junos and IOS).

dbarrosop:
Historically speaking there was no intent. If there had been we would have done what we did for the BGP ones. A solution would be to leave this method as it is, deprecate it, and invest in a new one that is like this method but adding an outer VRF key.

If we prefer to do the lazy thing and patch this method I personally don't have a strong opinion for doing one thing or the other but we remove any ambiguity in the docstring.

ktbyers:
@dbarrosop Let's start with determining which drivers currently support VRFs in get_interfaces_ip and which don't. If most of them already support VRFs, then we can just update NX-OS.

If most of them don't support VRFs, then we can circle back and decide what to do.

ktbyers:
You should make all regular expression patterns raw strings so:

```
pattern = r"^\s.some.regex.pattern.*$"
```

ktbyers:
This was fixed in PR #890

PR #725, opened 2018-05-12
Updated fix for Catalyst 6500 reporting VLAN as N/A

ktbyers:
Added unit test data

Also updated fix so that the entry wasn't silently ignored.


ExaneServerTeam:
@ktbyers Thanks for this PR ;)

PR #726, opened 2018-05-15
Fixing deprecated use of e.message

ktbyers:
.


ktbyers:
On the change in `__init__.py` I manually tested that on PY27 and PY36 (and the new solution should work properly).

PR #734, opened 2018-05-21
Raise SessionLockedException when there's already a session in progress

mirceaulinic:
In the context of concomitant configuration changes, there's a corner case when due
to a race condition, the device releases the config session (thus the session is
not listed under ``show configuration sessions``), however it didn't actually
finish committing. If there's another configuration change request in this exact
interval (i.e., between the moment we invoked ``commit_config`` and the moment
``commit_config`` finished executing), we will attempt to make a configuration
change and ``_lock`` in its current form won't detect it.
With this change, we check the ``config_session`` variable and in case NAPALM
didn't finish committing yet, would raise ``SessionLockedException``.
I run this change for a few months, and didn't notice any issues.


bewing:
LGTM once the PEP8 check is fixed.

dbarrosop:
@mirceaulinic what @bewing said :D

napalm/eos/eos.py
```diff
@@ -138,9 +138,11 @@ def is_alive(self):
     def _lock(self):
         if self.config_session is None:
             self.config_session = 'napalm_{}'.format(datetime.now().microsecond)
+        else:
+            raise SessionLockedException('Session is already in use: {}'.format(self.config_session))
         sess = self.device.run_commands(['show configuration sessions'])[0]['sessions']
         if [k for k, v in sess.items() if v['state'] == 'pending' and k != self.config_session]:
-            raise SessionLockedException('Session is already in use')
+            raise SessionLockedException('Session is already in use: {}'.format(k))
```
ktbyers: The variable k won't be known outside of the list comprehension. So that will need to be corrected.

ktbyers:
@mirceaulinic Is this still active?

ktbyers:
This had an issue in the PR--if it is still an issue, feel free to re-submit.

PR #741, opened 2018-06-02
Re-write of get_address_table()

PseudotsugaMeziesii:
TLDR; Used the header lines from command output to determine location of relevant information in subsequent rows.

Sorry if it's a bit presumptuous, but I decided to try and tackle a re-write of the get_mac_address() function. There was a lot of reliance on a custom regex for every single edge case that came up. It was starting to get pretty tough to follow. Cisco certainly didn't make it easy for us...

For the proposed changes, an attempt was made to leverage the header line of the command output to provide a reference for the subsequent rows containing MAC addresses (or interfaces with commas). There are still some kooky workarounds, but hopefully there might be fewer custom considerations required in the future.

Please take a look at what was changed and hit me up with any concerns. All testing was completed using the existing test files, and I tried to follow current conventions.

### DIFF
--- a/napalm/ios/ios.py
+++ b/napalm/ios/ios.py
@@ -1849,29 +1849,52 @@ class IOSDriver(NetworkDriver):
         All    1111.2222.3333    STATIC      CPU
         """
 
-        RE_MACTABLE_DEFAULT = r"^" + MAC_REGEX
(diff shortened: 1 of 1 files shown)


PseudotsugaMeziesii:
NM. I'm a douche. Failed testing...

PR #745, opened 2018-06-09
add get_environment to nxos_ssh

cspeidel:
power and fan are not currently implemented (they aren't implemented in the ios driver either).

### DIFF
--- a/napalm/nxos_ssh/nxos_ssh.py
+++ b/napalm/nxos_ssh/nxos_ssh.py
@@ -807,10 +807,9 @@ class NXOSSSHDriver(NXOSDriverBase):
         environment['cpu'][0]['%usage'] = 0.0
         system_resources_cpu = napalm.base.helpers.textfsm_extractor(self, 'system_resources',
                                                                      sys_resources)
-        from decimal import Decimal
         for cpu in system_resources_cpu:
-            cpu_dict = {cpu.get('cpu_id'): {'%usage': float(Decimal(100) -
-                        Decimal(cpu.get('cpu_idle')))}}
+            cpu_dict = {cpu.get('cpu_id'): {
+                        '%usage': round(100 - float(cpu.get('cpu_idle')), 2)}}
             environment['cpu'].update(cpu_dict)
 
         # memory
--- a/napalm/nxos_ssh/nxos_ssh.py
+++ b/napalm/nxos_ssh/nxos_ssh.py
@@ -788,6 +788,73 @@ class NXOSSSHDriver(NXOSDriverBase):
             cli_output[py23_compat.text_type(command)] = output
         return cli_output
 
+    def get_environment(self):
+        """
+        Get environment facts.
+
+        power and fan are currently not implemented
+        cpu is using 1-minute average
+        """
+
+        environment = {}
+        # sys_resources contains cpu and mem output
+        sys_resources = self._send_command('show system resources')
+        temp_cmd = 'show env temp'
+
+        # cpu
+        environment.setdefault('cpu', {})
+        environment['cpu'][0] = {}
+        environment['cpu'][0]['%usage'] = 0.0
+        for line in sys_resources.splitlines():
+            # Load average:   1 minute: 0.14   5 minutes: 0.18   15 minutes: 0.21
+            match = re.search(r'1 minute: [0-9]+(.[0-9]+)?', line)
+            if match is not None:
+                cpu = match.group(0).split(':')[1]
+                cpu = '{0:.2f}'.format(float(cpu))
+                environment['cpu'][0]['%usage'] = float(cpu)
+                break
+        else:
+            raise ValueError("Unexpected output from: {}".format(line))
+
+        # memory
+        for line in sys_resources.splitlines():
+            # Memory usage:   16401224K total,   4798280K used,   11602944K free
+            if 'Memory usage:' in line:
+                proc_total_mem, proc_used_mem, _ = line.split(',')
+                proc_used_mem = re.search(r'\d+', proc_used_mem).group(0)
+                proc_total_mem = re.search(r'\d+', proc_total_mem).group(0)
+                break
+        else:
+            raise ValueError("Unexpected output from: {}".format(line))
+        environment.setdefault('memory', {})
+        environment['memory']['used_ram'] = int(proc_used_mem)
+        environment['memory']['available_ram'] = int(proc_total_mem)
+
+        # temperature
+        output = self._send_command(temp_cmd)
+        environment.setdefault('temperature', {})
+        for line in output.splitlines():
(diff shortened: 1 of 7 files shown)


napalm/nxos_ssh/nxos_ssh.py
```diff
@@ -1008,6 +1008,69 @@ def cli(self, commands):
...
+        """
+        Get environment facts.
+
+        power and fan are currently not implemented
+        cpu is using 1-minute average
+        """
+
+        environment = {}
+        cpu_cmd = 'show system resources'
+        mem_cmd = 'show system resources'
```
itdependsnetworks: Why run the command two times?

napalm/nxos_ssh/nxos_ssh.py
```diff
@@ -1008,6 +1008,69 @@ def cli(self, commands):
...
+        environment.setdefault('cpu', {})
+        environment['cpu'][0] = {}
+        environment['cpu'][0]['%usage'] = 0.0
+        for line in output.splitlines():
+            # Load average:   1 minute: 0.14   5 minutes: 0.18   15 minutes: 0.21
+            match = re.search(r'1 minute: [0-9]+(.[0-9]+)?', line)
+            if match is not None:
+                cpu = match.group(0).split(':')[1]
+                cpu = '{0:.2f}'.format(float(cpu))
+                environment['cpu'][0]['%usage'] = float(cpu)
```
itdependsnetworks: A common pattern is to fail if you have not seen the data before, so adding something like the below, right here would help here.

```
            else:
                raise ValueError("Unexpected output from: {}".format(line))
```

cspeidel: Good idea, added in eb01e0810ebb2ad222f6350c4ca26ced7f8f2157

napalm/nxos_ssh/nxos_ssh.py
cspeidel: Thanks! I should have done one more pass over. I updated and squashed commits into 3ca6028

napalm/nxos_ssh/nxos_ssh.py
```diff
@@ -1008,6 +1008,73 @@ def cli(self, commands):
...
+        environment['cpu'][0]['%usage'] = 0.0
+        for line in sys_resources.splitlines():
+            # Load average:   1 minute: 0.14   5 minutes: 0.18   15 minutes: 0.21
+            match = re.search(r'1 minute: [0-9]+(.[0-9]+)?', line)
+            if match is not None:
+                cpu = match.group(0).split(':')[1]
+                cpu = '{0:.2f}'.format(float(cpu))
+                environment['cpu'][0]['%usage'] = float(cpu)
+                break
+        else:
```
itdependsnetworks: This should be indented in. Just to clarify the goal, it is that if you see a line you are not expecting, error out.

napalm/nxos_ssh/nxos_ssh.py
```diff
@@ -1008,6 +1008,73 @@ def cli(self, commands):
...
+                is_alert = float(is_alert)
+                temp = float(temp)
+                env_value = {'is_alert': temp >= is_alert,
+                             'is_critical': temp >= is_critical,
+                             'temperature': temp}
+                location = line.split()[1]
+                environment['temperature'][location] = env_value
+
+        # Initialize 'power' and 'fan' to default values (not implemented)
+        environment.setdefault('power', {})
```
itdependsnetworks: Napalm core team, are you ok with this?

napalm/nxos_ssh/nxos_ssh.py
cspeidel: The full output of show system resources is:
```
nx01-top# show system resources
Load average:   1 minute: 0.05   5 minutes: 0.28   15 minutes: 0.27
Processes   :   642 total, 1 running
CPU states  :   1.02% user,   2.57% kernel,   96.40% idle
        CPU0 states  :   1.03% user,   4.12% kernel,   94.84% idle
        CPU1 states  :   1.05% user,   2.10% kernel,   96.84% idle
        CPU2 states  :   1.02% user,   1.02% kernel,   97.95% idle
        CPU3 states  :   2.04% user,   2.04% kernel,   95.91% idle
Memory usage:   16401224K total,   5005012K used,   11396212K free
Current memory status: OK
```
I'm not expecting the load average to be the first line, knowing Cisco I'm allowing the flexibility that it could be on a different line on a different NX-OS. So this loops through the whole output looking for the regex and if found breaks out of the loop, if the regex isn't found on any line then ValueError is raised.

itdependsnetworks: I'll defer to @ktbyers but my opinion and experience is if you don't know the data you will get more false positives. Obviously doing it the way you intend would less likely to fail, but again, more likely to have "needle in the haystack" type issues.

cspeidel: I'll try using the textfsm extractor instead

ktbyers:
@cspeidel I think this PR will need rebased (otherwise it will become prohibitively hard to review).

napalm/nxos_ssh/nxos_ssh.py
```diff
@@ -1008,6 +1008,70 @@ def cli(self, commands):
...
+        sys_resources = self._send_command('show system resources')
+        temp_cmd = 'show env temp'
+
+        # cpu
+        environment.setdefault('cpu', {})
+        environment['cpu'][0] = {}
+        environment['cpu'][0]['%usage'] = 0.0
+        system_resources_cpu = napalm.base.helpers.textfsm_extractor(self, 'system_resources',
+                                                                     sys_resources)
+        from decimal import Decimal
```
cspeidel: @ktbyers no problem I went ahead and rebased. Let me know if there is a better way here to subtract the decimal idle value from 100 to get the cpu usage.

cspeidel:
I rebased again and removed the Decimal dependency, so I think this is good to go.

ktbyers:
Okay, let me see if I can review it.

cspeidel:
@ktbyers rebased again to keep it ready for review :)

cspeidel:
@ktbyers I rebased again in order to keep this PR ready for approval. I seemed to have pulled in some extra commits that aren't mine. Any tips on how to fix?

ktbyers:
@cspeidel How did you do the rebase? It looks like that possibly wasn't done correctly.


Regards, Kirk

cspeidel:
@ktbyers I rebased again. I'd like to contribute more to napalm, but this first pull request has taken quite some time so I'm hesitant to do more.

ktbyers:
@cspeidel Yes, understood...unfortunately we are short of reviewers/maintainers so things have bogged down to a certain extent.

ktbyers:
PR updated here:

https://github.com/napalm-automation/napalm/pull/973

PR #757, opened 2018-07-01
get route table of ios

NeilJudson:
add function get_route_table() in napalm/ios/ios.py


ktbyers:
@NeilJudson Any new getter needs to be first proposed in a GitHub issue and has to have a returned data structure agreed upon in that issue and also get agreement we are going to do that getter (and how many platforms it needs to be implemented on).

I am going to close this PR for now and we can revisit it in a GitHub issue.

PR #767, opened 2018-07-11
Fix #651

### LINKED ISSUE #651: On logical interface the "is_enabled" is FALSE with link down on Juniper
On Juniper MX, if the physical link is down, the "is_enabled" status for the logical interface is FALSE instead of TRUE:

servis@router1.re1> show configuration interfaces xe-2/2/5
description test;
mtu 9192;
unit 0 {
    family inet {
        mtu 9000;
        address 10.200.7.100/31;
    }
}

servis@router1.re1> show interfaces terse
xe-2/2/5                up    down
xe-2/2/5.0              up    down inet     10.200.7.100/31

But with the get facts, the result is:

 "xe-2/2/5": {
                    "description": "test",
                    "is_enabled": true,
                    "is_up": false,
                    "last_flapped": 68971.0,
                    "mac_address": "40:71:83:A2:81:A1",
                    "speed": 10000
                },
                "xe-2/2/5.0": {
                    "description": "",
                    "is_enabled": false,   <----- this should be TRUE
                    "is_up": false,
                    "last_flapped": 68971.0,
                    "mac_address": "40:71:83:A2:81:A1",
                    "speed": 10000
                },

ckishimo:
Fix #651


ktbyers:
@ckishimo Thanks for updating that JSON file...so much easier to read now.

ktbyers:
It doesn't look like the unit tests actually have the test case of iff-down. Is it possible to add this?

You should be able to make a directory here:

```
test/junos/mocked_data/test_get_interfaces/iffdown/expected_result.json
test/junos/mocked_data/test_get_interfaces/iffdown/get-interface-information.xml
```

You would need to be able to get the XML from a device or from the original issue and add it into (get-interface-information.xml). I think it would be good to have this covered in the unit test data.

Let me know if I missed something here, but I didn't see it when I looked through the current unit test output.

ckishimo:
@ktbyers You are totally right! I only fixed the issue. 
I just added another test case with logical interfaces disabled (with `iff-down` flag). Thanks

PR #770, opened 2018-07-17
Add NAPALM Dockerfile

mirceaulinic:


Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+## Install min deps
+RUN apt-get update
+
+## Install NAPALM & underlying libraries dependencies
+RUN apt-get install -y python-cffi python-dev libxslt1-dev libssl-dev libffi-dev \
+    && apt-get install -y python-pip \
+    && pip install -U cffi \
+    && pip install -U cryptography \
+    && pip install napalm
```
bewing: Change this to `pip install .` ?  Do we want the Dockerfile to install the latest release, or install the latest dev?

mirceaulinic: `pip install .` might be a good idea indeed

mirceaulinic: Updated - see 723fd72

PR #774, opened 2018-07-19
ios get_network_instances support for 12.4 and minor fixes

LukeDRussell:
Hi, this is my first PR, please be gentle :)

This PR does a few things, but my main focus has been to add support for legacy devices (12.4 ish). There is a current bug when run against legacy devices that the instance name is always returned with `;` appended.

1. Expected results had GigabitEthernet0 in both "default" and "mgmt-intf" VRFs. Removed Gi0 from default.
2. Had to expand `Gi0` out to `GigabitEthernet0` so that the outputs from show vrf detail and show ip int br matched.
3. Added a step where interfaces found in a VRF get removed from the default VRF
4. Expanded the show vrf detail regex to work on both newer and older operating systems
5. Additional mocked_data for 12.4
6. Variable naming consistency for "sh / show"

Tox tests are passing in 2.7 and 3.6.


LukeRussell:
I have triggered this build again on psnmc/napalm and it builds fine. https://travis-ci.org/psnmc/napalm/builds/406444877

Could one of the maintainers trigger another build?

LukeRussell:
I will need to add more interface short name to long name mappings.

Lo = Loopback
Tu = Tunnel
etc

LukeRussell:
Update for squash message:

Adds support for get_network_instances on legacy ios devices (~12.4) and fixes various problems with existing method.

1. Removes interfaces from "default" instance when they are found in a VRF.
2. Normalises interface names to their "long" format (GigabitEthernet0 instead of Gi0)
3. Adjusts and adds to test cases
4. Improves regex to work on older ios versions

More interface types can be easily added.

ktbyers:
@LukeRussell I started reviewing this. 

One item I noticed is that there already exists a function that maps short interface name to long interface name. It is called `canonical_interface_name`.

You could use it as follows:

```
long_name = canonical_interface_name(short_name)
```

The source for this function is in: `napalm/base/helpers.py`

Because of that I don't think we should have the `def _expand_interface_names(interfaces):` function and instead should use the `canonical_interface_name`

Can you update the PR to start using this?

LukeRussell:
Ah, I did think there might be one already but I obviously didn't look hard enough. Will replace it shortly.

ktbyers:
No worries it is pretty hidden and we aren't consistently using it across methods.

LukeRussell:
Sorry I'm probably not going to get this updated for another week. I'm moving home this week, so I'm a little short on time both inside and outside work.

ktbyers:
No worries

mirceaulinic:
Bump @LukeDRussell - are you still working on this branch? :smile:

LukeDRussell:
I can't say I am @mirceaulinic. It's probably way out of date. I may come back to it at some stage, but don't wait on me if you are wanting to clean up.

mirceaulinic:
@LukeDRussell - let me know when you're back on this, and I'll be happy to reopen.

PR #782, opened 2018-08-01
Fix Travis and PY34 Tox issue

ktbyers:


bewing (approved):
Looks just as good to me

PR #789, opened 2018-08-16
Add driver_path to make it easier to find templates

ogenstad:
The path for the textfsm templates was wrong when you inherited from
a driver class.

As I mentioned regarding the problem I saw with napalm-inspector and anyone who would want to extend a driver might experience. Looks like ios was the only driver that needed the new method for the tests to run, but I added it to all the drivers which had a textfsm folder under the utils directory.

### DIFF
--- a/napalm/base/base.py
+++ b/napalm/base/base.py
@@ -1629,3 +1629,8 @@ class NetworkDriver(object):
             return napalm.base.helpers.canonical_interface_name(interface, addl_name_map=None)
         else:
             return interface
+
+    @staticmethod
+    def _driver_path():
+        """Return the path to the current driver."""
+        raise NotImplementedError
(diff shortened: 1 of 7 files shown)


ktbyers:
This will break the napalm community drivers that use templates. At a minimum should notify them that they will need to make this update if they are using textfsm templates.

Also we can remove the utils directories that are inside of ./test with this change (except for the one under `base`).

ktbyers:
I think we probably should put a try/except in base and fallback to the old way (if the new way fails) and have a warning get printed.

ktbyers:
@bewing I will see if I can review your PR today or tomorrow (hopefully today).

ktbyers:
Replaced by this PR https://github.com/napalm-automation/napalm/pull/798

PR #790, opened 2018-08-16
New feature - get_vni()

FloLaco:
Hello guys,

I'm working on a Cisco Nexus VXLAN fabric so I'm pushing one method concerning VXLAN.
VNI is : Virtual Network Identifier, as VLAN in a classic ethernet network.

Please tell me if something is unclear
Florian


ktbyers:
Unfortunately, we are not accepting new getters. The issue is we don't have enough maintainers on the current getters so the backlog/amount of work has gotten too large. 

All new getters would need to be implemented using napalm-yang.

We might make an exception for a getter that was implemented on all of the core platforms (Junos, EOS, IOS-XR, NX-OS, and NX-OS SSH). 

Sorry about that...but we need a better way of communicating this.

You can alternatively have your own getters that augment what we have and just subclass our classes with your class so you can use your method(s).

Regards, Kirk

FloLaco:
@ktbyers What's the best way to create my own class ?
Should I create a new driver ? If so, what things I need to edit in order to work (base.py ? etc ...)

ktbyers:
In your Python module:

```
$ python
Python 2.7.14 (default, May  2 2018, 18:31:34) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-11)] on linux2
Type "help", "copyright", "credits" or "license" for more information.

>>> from napalm.nxos import NXOSDriver
>>> class MyClass(NXOSDriver):
...   pass
... 
>>> from napalm.nxos_ssh import NXOSSSHDriver
>>> class MyClass2(NXOSSSHDriver):
...   pass
... 
```

Depending on whether you are using the SSH driver or the NX-API driver. You can then add your new methods inside of your class.

Basically, it will first try to use the methods in your class. In any case where a given method doesn't exist in your class, it will try to use the method in the parent class NXOSDriver or NXOSSSHDriver.

Hope this helps.

FloLaco:
@ktbyers Thank you.
But how can I use the new class ?

Right now, I use this method to use NAPALM : 
```
driver = get_network_driver("nxos")
....
```

I tried this kind of hack, but it don't works : 
```
    driver = MyNXOSDriver.get_network_driver("nxos")
AttributeError: type object 'MyNXOSDriver' has no attribute 'get_network_driver'
```

ktbyers:
Let me try to follow up with how to do this. Ping me if I forget.

FloLaco:
@ktbyers After reading the __init__.py file in `base/`, I saw that the module that is try to being loaded is `custom_napalm.{myfilename}`. So in my current directory, I've added this directory : 
custom_napalm/
-- mynxos.py
-- __init__.py

Now it works doing this : 
```
driver = get_network_driver("mynxos")
```

PR #791, opened 2018-08-16
NXOS - Add received_prefixes in return value and multiple AF

FloLaco:
Hello,

I've improved (I hope so) the `get_bgp_neighbors()` method : 
- Change the `received_prefixes` from `-1` to the real value
- Add support of multiple address family (ipv4, ipv6, vpn4, vpnv6, l2vpn)

Florian


napalm/nxos/nxos.py
```diff
@@ -982,7 +1003,7 @@ def get_ntp_stats(self):
 
     def get_interfaces_ip(self):
         interfaces_ip = {}
-        ipv4_command = 'show ip interface'
+        ipv4_command = 'show ip interface vrf all'
```
ktbyers: Why is this changing? That probably should be in a separate change if it needs to change?

napalm/nxos/nxos.py
```diff
@@ -1032,7 +1053,7 @@ def get_interfaces_ip(self):
                     'prefix_length': secondary_address_prefix
                 })
 
-        ipv6_command = 'show ipv6 interface'
+        ipv6_command = 'show ipv6 interface vrf all'
```
ktbyers: Why is this changing? That probably should be in a separate change if it needs to change?

napalm/nxos/nxos.py
FloLaco: I was adding the vrf support at that time, and yes it's not related to the bgp_neighbors

napalm/nxos/nxos.py
FloLaco: I was adding the vrf support at that time, and yes it's not related to the bgp_neighbors

ktbyers:
@FloLaco Okay, nice...can you back at those changes for the get_interfaces_ip commands?

I think this PR is ready to go besides that.

FloLaco:
@ktbyers sure I will

FloLaco:
@ktbyers it's done. Sorry for the multiple push, I've forgotten the mocked data file

PR #795, opened 2018-08-23
improve validate.py so it only connects once per getter type

jedelman8:
I know this can be optimized, but just wanted to open it to get feedback for now.

Was doing tests using `_name` for several compliance tests like this:

```
- get_facts:
    _name: verify_os_version
    os_version: 15.1F4.15

- get_facts:
    _name: verify_hostname
    hostname: vmx7
```
Same is true for other getters and thought it made sense to just gather facts once and reduce the time it takes to gather data! 

You'll see this list hard-coded (`["ping", "is_alive", "traceroute", "get_route_to"]`) because the results will vary for each of those tests (and ppl should really use `_name` for those too) so the object is one layer deeper to account for that.


ktbyers:
Some unit tests are failing (also, there is a minor linter issue).

mirceaulinic:
Bump @jedelman8

ktbyers:
@itdependsnetworks @jedelman8 Are you still working on this?

itdependsnetworks:
Let me check tomorrow

ktbyers:
I am going to close this since never fully working. Just re-open if you want to update/fix the issues.


Kirk

PR #798, opened 2018-08-23
Allow classes to use parent's textfsm templates

bewing:
Similar to the change in `load_template`, allow classes that inherit
from a NetworkDriver class to optionally override the textfsm templates
used via `textfsm_extractor`, or utilize the MRO to provide a list of paths
to search until a template is found.


ktbyers:
Documenting the behavior:

```python
[<class 'test.eos.conftest.PatchedEOSDriver'>, 
 <class 'napalm.eos.eos.EOSDriver'>, 
 <class 'napalm.base.base.NetworkDriver'>, 
 <class 'object'>]
```

In this example, it would first look in the following directories in the following order:

```
'/path/napalm/test/eos/utils/textfsm_templates'
'/path/napalm/napalm/eos/utils/textfsm_templates'
'/path/napalm/napalm/base/utils/textfsm_templates'
```

Issue #802, opened 2018-09-05
Destination addr error in get_route_to () in junos


Mabuchin:
Using the `get_route_to()` function in junos.
If there is a VPNv4 route in the acquired route, it can not be acquired with an invalid error.

A sample of the error is below.
```
  File "/home/mabuchin/devices/models.py", line 52, in dump_all_routes
    return self.device.get_route_to('0.0.0.0/0')
  File "/home/mabuchin/.local/share/virtualenvs/test-pbAc2kqw/lib/python3.6/site-packages/napalm/junos/junos.py", line 1614, in get_route_to
    destination = napalm.base.helpers.ip (d.pop ('destination', ''))
  File "/home/mabuchin/.local/share/virtualenvs/test-pbAc2kqw/lib/python3.6/site-packages/napalm/base/helpers.py", line 248, in ip
    addr - obj = IPAddress (addr)
  File "/home/mabuchin/.local/share/virtualenvs/test-pbAc2kqw/lib/python3.6/site-packages/netaddr/ip/__init__.py", line 306, in __init__
    'address from% r'% addr)
netaddr.core.AddrFormatError: failed to detect a valid IP address from '198.18.255.2:11:198.18.1.12'
```

This was due to `get_route_to()` in junos, because the IP validation was applied to the destination. ( `Napalm.base.helpers.ip`
Please look at the following.

https://github.com/napalm-automation/napalm/blob/82432cccadd7f9d72e2e95dcc36ff642927d10ab/napalm/junos/junos.py#L1619

VPNv4 is general route information and I think that it can be acquired with `get_route_to()`.
I think that it is better add VPNv4 helpers and apply it to `get_route_to()` or removing IP validate(only destination validate) .

Mabuchin:
It was solved by # 803. Thanks

PR #803, opened 2018-09-06
Delete the ip validation of the destination of get_route_to in junos

Mabuchin:
#802 
Destination in get_route_to other than junos (eg iosxr / eos) does not validate ip.
Also, there is a possibility that non-IP will be included in destionation in any OS. (VPNv 4)

junos has already parsed the value you get from the rpc response using `junos_views.yaml`.(https://github.com/napalm-automation/napalm/blob/develop/napalm/junos/utils/junos_views.yml) 

I think it was not necessary to verify the IP here.

### DIFF
--- a/napalm/junos/junos.py
+++ b/napalm/junos/junos.py
@@ -1616,7 +1616,7 @@ class JunOSDriver(NetworkDriver):
             d = {}
             # next_hop = route[0]
             d = {elem[0]: elem[1] for elem in route[1]}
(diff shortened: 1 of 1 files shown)


ktbyers:
I merged this, but do you have example output where this is not an IP address. 

I say this as I am wondering if the napalm code would actually work if it was not an IP prefix (i.e. whether we are just masking errors that would happen later) and whether the returned result actually makes sense.

I.E. if it is going to fail anyways, it is better to fail earlier with a clear message as to why.

PR #805, opened 2018-09-10
Fix ResourceWarning: unclosed file in base/helpers.py

BoboTiG:
Hello,

A little patch to prevent file descriptor leaks :)


ktbyers:
Probably should use a context manager for fixing this issue.

BoboTiG:
Done. I refactored the method, I think it is a little clearer now.

PR #806, opened 2018-09-10
Fix the use of deprecated inspect.getargspec() for inspect.getfullarg…

BoboTiG:
…spec() in netmiko_helpers.py


ktbyers:
Very minor lint issue that you need to fix (that is causing the unit tests to fail).

bewing:
```
# git grep -l getargspec
napalm/base/mock.py
napalm/base/netmiko_helpers.py
napalm/base/test/getters.py
```
Possibly worth moving this solution to base.util.py23 and importing it from there?

ktbyers:
Will separately fix the use in mock.py and in getters.py

PR #811, opened 2018-09-17
Fix #810

### LINKED ISSUE #810: `EOSDriver.get_environment()` throws keyError on 7010T
Arista 7010T switches have unmanaged power supplies and are missing the "powerOutput" key:

```python
In [4]: d.get_environment()
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-4-b99851033f4f> in <module>()
----> 1 d.get_environment()

~/.pyenv/versions/3.7.0/lib/python3.7/site-packages/napalm/eos/eos.py in get_environment(self)
    545                     'status': data['state'] == 'ok',
    546                     'capacity': data['capacity'],
--> 547                     'output': data['outputPower']
    548                 }
    549         cpu_lines = cpu_output.splitlines()

KeyError: 'outputPower'
```

Looking at the other codebases, it appears we want to set the output to -1 in this case?

Raw JSON from a 7010T:
```json
{'powerSupplies': {'1': {'uptime': 1531344080.3479242,
   'modelName': 'PWR-00194-01',
   'capacity': 65.0,
   'fans': {},
   'tempSensors': {},
   'managed': False,
   'state': 'ok'},
  '2': {'fans': {},
   'modelName': 'PWR-00194-01',
   'capacity': 65.0,
   'tempSensors': {},
   'managed': False,
   'state': 'powerLoss'}}}
```

bewing:
Update dictionary access to provide fallback defaults if keys are missing
when accessing power data.  7010T switches do not have managed supplies,
so they do not report output current. Returns float(-1.0) to mimic the
behavior of the IOS driver.


napalm/eos/eos.py
```diff
@@ -542,9 +542,9 @@ def extract_temperature_data(data):
         if not is_veos:
             for psu, data in power_output['powerSupplies'].items():
                 environment_counters['power'][psu] = {
-                    'status': data['state'] == 'ok',
-                    'capacity': data['capacity'],
-                    'output': data['outputPower']
+                    'status': data.get('state', False) == 'ok',
```
ktbyers: Looking on what we did on Junos (and somewhat on IOS), I think we should probably have status default to True in the unknown case)...i.e. think this is probably more consistent with other drivers and we have no basis for saying there is a problem. Also the -1 will indicate we don't have data.

PR #813, opened 2018-09-18
nxos_ssh Update sh int regexp for 7.3.3.N1

bewing:
Thanks to Bill Nelson for reporting.


connifer:
Thanks for fixing!  Getting the interface info to my user will make him very happy!

I assume this fix will take time before it can be installed via pip.  Is there a development/early release pkg to use?  If yes, where would I get it/how would I install?

PR #826, opened 2018-10-04
junos cli function: all pipes ignored ?

afourmy:
That's strange but it would seem that with the current code, because of that one line, all pipes are ignored for junos `cli` function. 
I don't have access to any juniper device so I cannot test. If I'm right, it probably means that there should be unit tests with pipes for that `cli` function.


mirceaulinic:
Thanks @afourmy - please feel free to add some unit tests for this.

PR #829, opened 2018-10-08
Fixing username prompting issue.

ktbyers:
Needs tested for both replace and merge operation.

Needs regression tested also against real device(s)


ktbyers:
@jwbensley My bad...I have updated with the change I intended over the weekend. I somehow messed up the commit over the weekend and didn't actually submit the change that I intended, but it is there now.

PR #830, opened 2018-10-12
Update pynxos to 0.0.4 and add verify optional_args to nxos

targuan:
Update current version of pynxos >= 0.0.4: #784
Verify argument is exposed via ``verify`` key in ``optional_args``. It defaults to ``False`` for compatibility and can be set to ``True`` or any value allowed by requests library.
I kept the InsecureRequestWarning warnings from urllib3on purpose. These warnings can be easily blacked out.

I don't knwo how this code can be tested as the mocked driver use for the test bypass the open method.
I have just tested it against a live device.

### DIFF
--- a/napalm/nxos/nxos.py
+++ b/napalm/nxos/nxos.py
@@ -199,6 +199,8 @@ class NXOSDriver(NXOSDriverBase):
         elif self.transport == 'http':
             self.port = optional_args.get('port', 80)
 
+        self.verify = optional_args.get('verify', False)
+
     def open(self):
         try:
             self.device = NXOSDevice(self.hostname,
@@ -206,7 +208,8 @@ class NXOSDriver(NXOSDriverBase):
                                      self.password,
                                      timeout=self.timeout,
                                      port=self.port,
-                                     transport=self.transport)
+                                     transport=self.transport,
+                                     verify=self.verify)
             self.device.show('show hostname')
             self.up = True
         except (CLIError, ValueError):
--- a/napalm/nxos/nxos.py
+++ b/napalm/nxos/nxos.py
@@ -199,7 +199,7 @@ class NXOSDriver(NXOSDriverBase):
         elif self.transport == 'http':
             self.port = optional_args.get('port', 80)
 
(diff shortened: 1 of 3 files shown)


ktbyers:
I will try to see if I can get this merged this weekend.

ktbyers:
@targuan Can you update the optional_arg to be `ssl_verify`? I think `verify` is just a bit too generic. Obviously this is only on the napalm side; it will still be what it is in pynxos.

It will be easier to have you update it then me (so I can just review and approve it).

Thanks.

targuan:
Thanks for the review.
I've updated the optional_arg to ssl_verify.

napalm/nxos/nxos.py
```diff
@@ -199,14 +199,17 @@ def __init__(self, hostname, username, password, timeout=60, optional_args=None)
         elif self.transport == 'http':
             self.port = optional_args.get('port', 80)
 
+        self.ssl_verify = optional_args.get('verify', False)
```
ktbyers: You didn't change the second half of this i.e. optional_args.get is still looking for verify.

targuan:
Sorry for the time wasted.
I did correctly tested it this time:
It raise SSLError when the check is enabled
```
(.venv) [root@localhost napalm]# napalm --user admin --pass Cisco123 --vendor nxos --optional_args 'ssl_verify=True' 10.10.20.58 call get_facts
2018-10-14 00:33:08,456 - napalm - ERROR - open - Failed: HTTPSConnectionPool(host='10.10.20.58', port=443): Max retries exceeded with url: /ins (Caused by SSLError(SSLError(1, u'[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:579)'),))
```
It complete when it ssl_verify is disabled:
```
(.venv) [root@localhost napalm]# napalm --user admin --pass Cisco123 --vendor nxos --optional_args 'ssl_verify=False' 10.10.20.58 call get_facts &> /dev/null 
(.venv) [root@localhost napalm]# echo $?
0
```

Issue #475, opened 2017-10-24
NXOS fails to create checkpoint file if file already exists


ktbyers:
This would fail (basically take over 12 minutes to complete...potentially never finish):

```
    def _create_sot_file(self):
        """Create Source of Truth file to compare."""
        commands = ['terminal dont-ask', 'checkpoint file sot_file']
        self._send_config_commands(commands)
```

If the `sot_file` already existed.

This was observed on NX-OS 6.2.16

ktbyers:
Probably should just delete `sot_file` before trying to write it.

ktbyers:
Should be fixed here:

https://github.com/napalm-automation/napalm/pull/833

ktbyers:
This was fixed as per the above PR.

Issue #474, opened 2017-10-24
napalm-nxos SSH send_config_commands() method is not using config_mode


ktbyers:
Should be using the `send_config_set()` method in Netmiko.

ktbyers:
Fixed here https://github.com/napalm-automation/napalm/pull/833

Issue #447, opened 2017-10-24
Atomic Merge


mirceaulinic:
Could the merge process be made atomic by taking a snapshot that could be rolled back just before the merge? I thought about writing an enhancement, but I wanted to check here first to see if there is some other technical reason that merges cannot be atomic on this platform.

@jedelman8:

> Makes sense.  This is actually how we do atomic changes with Ansible on
nxos using the nxos_rollback and ntc_rollback modules.  So yes, in
agreement for merges:  Step 1.  Create a checkpoint file.  Step 2: send
commands.  Step 3: rollback on error

---

Issue moved from https://github.com/napalm-automation/napalm-nxos/issues/77

ktbyers:
I think we should do this i.e. before merge operation create a checkpoint that could be rolled back to.

It still wouldn't be atomic as 1. if we are disconnected, there will be no rollback, 2. Partial changes could go through which you would have to manually handle and rollback, 3. You potentially would have to do this manually.

Though for both items 2 and 3 we could check for errors in the merge attempt and auto-rollback which we probably should do.

I probably view this as a bug i.e. we should already be doing some of this.

ktbyers:
This should be fixed here https://github.com/napalm-automation/napalm/pull/833

PR #833, opened 2018-10-13
NX-OS refactor

ktbyers:
I want to release napalm 2.3.3 before merging this.

This is a big change to both the NX-OS drivers

- Refactor SCP to use Netmiko
- Significantly reduce redundant code between two drivers
- Use standardized file names for rollback file and replace file (same file names as IOS)
- Added a platform attribute to indicate 'nxos', 'ios', 'junos' et cetera
- Improve NX-OS merge atomic behavior
- Making it easier for potentially future adding a get_file and put_file using Netmiko
- Making it easier for potentially adding XML support in NX-OS (and replacing/elminating NX-OS jsonrpc and SSH support).
- Also fixes issues #475 #447 #474


ktbyers:
Ran napalm_custom_test on nxos and nxos_ssh for this updated driver. I also did quite a bit of testing using pdb and a real device.

https://github.com/ktbyers/napalm_custom_test

ktbyers:
Added unit tests to test that automatic rollback on a failed merge operation worked properly for NX-OS (this is in the testing against real devices). Currently, I only ran it as a one off test; not running on an ongoing basis (it has to be merged into 'develop' before I can test on an ongoing basis).

https://github.com/ktbyers/napalm_custom_test/commit/37a961fb6595b74b0e90360fb72126c25874e4a6#diff-16b358213f85e70673686e009a79b591R93

mattschwen:
+1 

I have reviewed the code briefly for a sanity check and have tested the PR changes against physical hardware. I used multiple getters for `NXOS` and `NXOS_SSH` as well as config operations on both `9K` and `5k` with no tracebacks or odd behavior.

bewing (approved):
LGTM.

PR #837, opened 2018-10-22
NX-OS get_network_instances

auerlukas:
Implemented get_network_instances(...) for NX-OS using RESTful API


auerlukas:
not quite sure if I corrected the merge conflict the right way. please review

PR #852, opened 2018-10-30
IOS ignore invalid input

TomCos:
While this line seems harmless, and even helpful, there are a few things to consider:

IOS is old and a mess.  Unlike NXOS, the commands are not consistent across hardware platforms.  For example, some have 'show module' and 'show inventory', while others do not.  When trying to build automation for these older devices, it's important to poke around a bit and see what works and what does not.  Shooting back and exception breaks the poking around.

Personally, I think the exception should be removed altogether, as there are other spots in the code where we explicitly back out of the task (such as commit_config and get_interfaces_ip) if it is seen, however, without knowing the background, I think my solution provides the best of both worlds compatibility wise.  Keep in mind, we're not doing this check in the NX-OS platform for 'cli', so we should be consistent, whatever the solution is.


TomCos:
Digging real fast, NXOS doesn't have anything like it, neither does JunOS, but EOS does have a CommandError exception, also containing the word "Invalid".

ktbyers:
Feel free to resubmit, but the behavior/method signature needs to be consistent across all of the core napalm platforms.

TomCos:
I'm struggling with this one, because it's already *not* consistent right now.  Would anyone be opposed to removing the exception?

ktbyers:
I am open to removing the exception being raised as long as it is consistent with the behavior of the other platforms.

@mirceaulinic any thoughts?

PR #858, opened 2018-11-05
Remove invalid exception cli IOS

TomCos:
nxos_ssh - No exception for invalid command
nxos - No exception for invalid command
junos - No exception for invalid command
iosxr - No exception for invalid command
ios - Does have one, but will be removed after this merge
eos - Has an exception for pyeapi.eapilib.CommandError (https://pyeapi.readthedocs.io/en/latest/client_modules/eapilib.html), but I'm not familiar enough with the platform to decide if this is inconsistent or not, so leaving that one alone.


ktbyers:
@TomCos Why do you have those LLDP changes in this PR?

PR #865, opened 2018-11-13
change junos lock method

borisovano:
Allow users to skip locking in order to manage it elsewhere. 
One use case would be to manage locking externally to perform several config_load without unlocking the configuration in between.


ktbyers:
Why can't you just set config_lock to False and do your own locking?

Can you expand some more on why you need to do this? How you would use this?

borisovano:
@ktbyers that's exactly what we wanted to do! (set config_lock to False and manage own locking)

However it appears that config_lock name is deceiving, 
if True it means it will lock the connection for the entire session.
if False it won't avoid lock altogether as the name might suggest.

Instead it won't lock on session level, but still lock after each load_template:

```
    def _load_candidate(self, filename, config, overwrite):
        if filename is None:
            configuration = config
        else:
            with open(filename) as f:
                configuration = f.read()

        if not self.config_lock:
            # if not locked during connection time
            # will try to lock it if not already aquired
self._lock()

```

borisovano:
@ktbyers I would be grateful if you could throw a quick glance at the code & review 
(the change is pretty small and self contained)

ktbyers:
@borisovano I will see if I can get something done on this in the next couple of days. Ping me if I forget (i.e. just comment on this PR).

borisovano:
@ktbyers Thanks

ktbyers:
Re-opening as it hasn't been merged.

borisovano:
@ktbyers Thanks! closed by mistake

(PS - ping:)

napalm/junos/junos.py
```diff
@@ -888,7 +901,7 @@ def _process_pipe(cmd, txt):
             Process CLI output from Juniper device that
             doesn't allow piping the output.
             '''
-            if txt is None:
+            if txt is not None:
```
mirceaulinic: What's that?

ktbyers:
This is on me...I was supposed to look at this.

I think I am going to reimplement this in a slightly different way (where the end-objective of the original PR was still achieved).

ktbyers:
Replaced by this PR:

https://github.com/napalm-automation/napalm/pull/881

There are also some re-base issues in this PR

PR #872, opened 2018-11-30
Fix nxos_ssh LLDP neighbors implementation

### LINKED ISSUE #834: nxos_ssh get_lldp_neighbors using TextFSM template is unlikely to be correct
I am basing this on IOS, but that first field in IOS is a fixed length field (20 characters) and can have spaces in it:

```
$ cat ./nxos_ssh/utils/textfsm_templates/lldp_neighbors.tpl
Value NEIGHBOR (\S+)
Value LOCAL_INTERFACE (\S+)
Value NEIGHBOR_INTERFACE (\S+)

Start
  ^Device.*ID -> LLDP

LLDP
  ^${NEIGHBOR}\s+${LOCAL_INTERFACE}\s+\d+\s+[\w+\s]+\S+\s+${NEIGHBOR_INTERFACE} -> Record

```

jobec:
Fixes #834


jobec:
Couple of remarks:

1.  There was an inconsistency between the interface names from `get_interfaces()` and `get_lldp_neighbors()`/`get_lldp_neighbors_detail()`. I changed it (and all mock data) to using the canonical interface name. This is what `get_interfaces()` returns

1. Because the remote Chassis ID isn't necessarily a MAC address (it depends on what the neighbor implemented as a chassis id type), I removed the turn-chassis-id-into-mac piece of the code.

ktbyers:
I vote we switch the `super` calls back to the previous format as it simpler, the standard way we do super in napalm, and probably/possibly the more proper pythonic way to do this in Python3.

https://docs.python.org/3/library/functions.html#super

jobec:
Actually, it is the “pythonic” way to do it like this when you want to keep python 2.7 compatibility.

* https://docs.python.org/2/library/functions.html#super
* https://github.com/paramiko/paramiko/search?q=super%28&type=Code
* https://github.com/django/django/blob/stable/1.11.x/django/db/models/aggregates.py#L19

The way it was, is how the python-future package advertises it. But it’s not something I’ve ever seen in other packages.

Anyway, I’ll revert it.

jobec:
The super call code is reverted

PR #873, opened 2018-12-03
Black format napalm repository

ktbyers:
Also dropped Python3.4 from testing and rolled Tox version to be a newer version.


ktbyers:
Superceded by #885

PR #875, opened 2018-12-06
Update get_lldp_neighbors_detail function

jnmatlock:
Added two additional device SWITCH models including EX4300-32F and EX2300-48T.

### DIFF
--- a/napalm/junos/junos.py
+++ b/napalm/junos/junos.py
@@ -760,11 +760,13 @@ class JunOSDriver(NetworkDriver):
                 'EX9208': rpc_call_without_information,
                 'EX3400': rpc_call_without_information,
                 'EX4300-48P': rpc_call_without_information,
+                'EX4300-32F': rpc_call_without_information,
                 'EX4600-40F': rpc_call_without_information,
                 'QFX5100-48S-6Q': rpc_call_without_information,
                 'QFX5110-48S-4C': rpc_call_without_information,
                 'QFX10002-36Q': rpc_call_without_information,
                 'QFX10008': rpc_call_without_information,
+                'EX2300-48T': rpc_call_without_information,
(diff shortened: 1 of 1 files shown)


ktbyers:
Can we switch this to more of a regex-pattern solution (for example, 2300.*, 4300.*)?

By this I mean instead of updating this per exact-model number, update it using a model-pattern where the model-pattern wouldn't care about the number of ports or whether they are powered or not, et cetera.

I am referring here to the entire solution and not just your specific fix.

```
             'SWITCH': {
                 'default': rpc_call_with_information,
                'EX9208': rpc_call_without_information,
                'EX3400': rpc_call_without_information,
                'EX4300-48P': rpc_call_without_information,
                'EX4300-32F': rpc_call_without_information,
                'EX4600-40F': rpc_call_without_information,
                'QFX5100-48S-6Q': rpc_call_without_information,
                'QFX5110-48S-4C': rpc_call_without_information,
                'QFX10002-36Q': rpc_call_without_information,
                'QFX10008': rpc_call_without_information,
                'EX2300-48T': rpc_call_without_information,
                'EX2300-24P': rpc_call_without_information,
                'EX2300-C-12P': rpc_call_without_information
            },
```

jnmatlock:
Great point. Looking into that, right now.

ktbyers:
Okay, great...yes, I think if we make this a bit more pattern based we can hopefully reduce the amount of churn/updates due to this.

jobec:
See #876 
Instead of just adding yet another model to the list, I propose to automatically handle it, removing the maintenance of this list.

I'll see on opening a PR for this.

ktbyers:
Obsoleted by https://github.com/napalm-automation/napalm/pull/878

PR #877, opened 2018-12-07
Implement VRF support for get_interfaces_ip for nxos_ssh

jobec:
This augments PR #819 and merges into the same branch (not develop)


jobec:
Rebased on current `develop` branch, hence this also pulls in all changes from that branch ever since the target branch for this PR was created.

TheRealBecks:
@ktbyers and @mirceaulinic: I forked the branch from @jobec and did a rebase, you can find it here: https://github.com/TheRealBecks/napalm/tree/becks_vrf
The current status on that branch is as follows:

- EOS: not implemented
- IOS: not implemented
- IOSXR: not implemented
- Junos OS: implemented, but I haven't tested it yet
- NXOS: not implemented
- NXOS_SSH: implemented, but I haven't tested it yet

What's the next step to get this PR merged?

mirceaulinic:
Hey @TheRealBecks - as that's a pretty common feature (i.e., not proprietary), I think it's reasonable to aim to implement it across all the core drivers. If you don't have the resources to implement for everything, I would suggest you go ahead with as much as you possibly can, then try to engage the community to help with the rest. I could probably be able to help with the EOS one if needed.

mirceaulinic:
Hey @jobec - don't suppose you'd be willing to re-submit these changes to get 'em into the `vrf_get_interface_ip` branch? :sweat_smile:

TheRealBecks:
I don't think that it's a good idea to merge the request as the driver has not implemented all VRF cases and there are conceptual problems, for example see the documentation for get_interfaces_ip in the base driver:
"'vrf' of null-string will default to the global or default VRF. 'vrf' of 'all' will
        return the interface IPs for all VRFs. 'vrf' referring to a specific VRF will return
        the interface IPs of that specific VRF. In all cases the same data structure is returned
        and no reference to the VRF that was used is included in the output."

The drivers need to handle >none< ('default'), 'all' and a specific VRF name. But I see a bigger problem, because the VRF name gets thrown away, but when handling VRFs this is a major information. In each VRF IP prefixes can be reused, so it could be possible to have a device with the 'same' IP address on different interfaces with different VRFs. Without that information the return values are useless in some cases and can lead to false data handling when not knowing about this VRF handling in NAPALM.

I would even go one step further: I think the VRF handling is broken in NAPALM, because every function which handles VRFs has its own, but not intuitive VRF handling. These VRF functions exist at the moment:
1. get_bgp_neighbors:
* first dictionary will be the VRF name
* defaults to 'global' (default VRF) if no VRF is specified
* can't handle 'all' VRFs
2. get_bgp_neighbors_detail:
* first dictionary will be the VRF name
* defaults to 'global' (default VRF) if no VRF is specified
* can't handle 'all' VRFs
=> same handling as get_bgp_neighbors
3. get_arp_table:
* VRF name won't be used as first dictionary
* defaults to 'all' VRFs if no VRF is specified
* "Specific 'vrf' will return the ARP table entries for that VRFs" --> That's vendor specific, on Arista EOS the return values are correct for the VRF. I'm sure other modern OSes like NX-OS will handle that accordingly. So this function is not doing it right (or the documentation is old/wrong?)
* Can't handle 'all' VRFs
4. get_interfaces_ip:
* VRF name will not be used as first dictionary
* defaults to 'global' (default VRF) if no VRF is specified
* Can handle >none<, 'all' or a specific VRF
5. get_network_instances
* first dictionary will be the VRF name
* defaults to 'all' VRFs if no VRF is specified

As we can see the function documentation (!) says that the functions are all different in their:
* input handling (>none<, 'all', VRF name)
* default VRF name when no VRF is specified
* dictionary output has or has not the VRF name as first dictionary

I think that a concept of handling VRFs in NAPALM is needed first before merging more VRF related code. Afterwards the already implemented functions need to be changed accordingly. And yeah, that would break the current functions handling in other tools :/ But in my opinion it's worth the effort when the VRF handling gets way easier and the output gets aligned.

And with that being said I would also recommend to change the wording from 'global' to 'default', because 'global' is a historical Cisco IOS wording. As far as I found out it's as follows:
* Cisco IOS: global
* Cisco NX-OS: default
* Arista EOS: default (they are sayinf 'default VRF', they live the VRF dream! ;) )
* Junos OS: 'master' (instance is 'master', but referenced in commands will be  'inet.0' and 'inet6.0')
=> 'default' fits best, because VRFs and routing tables are device local, that's not 'global' at all :)

What do you think?

mirceaulinic:
I think this branch is pretty much stale at this point.

> I don't think that it's a good idea to merge the request as the driver has not implemented all VRF cases 

@TheRealBecks I was suggesting @jobec to resubmit against a different branch, not getting this merged into out `develop` branch. To the broader points you raised, I'd suggest you opening an issue to discuss.

PR #880, opened 2018-12-12
Move save to netmiko

TomCos:
move napalm ios save config to netmiko's implementation, so we can fix the confirm in netmiko and allow for better implementation in the future


TomCos:
@ktbyers This is step 1, step 2 is to fix the netmiko save config to detect for confirm dynamically.  I'll work to make it backwards compatible, so either work.

TomCos:
If https://github.com/ktbyers/netmiko/pull/1037 gets approved, we need to add the delay factor back in, or should move the save to a function, similar to NXOS

PR #881, opened 2018-12-13
Allow for disable of Junos Locking so that it can be externally managed

ktbyers:


ktbyers:
@borisovano I think this is identical to what you implemented (as far as operation). Let me know if you see any issues.

@mirceaulinic Let me know if this looks good. I fixed a couple of rebase issues that were in the previous PR.

napalm/junos/junos.py
```diff
@@ -82,12 +84,15 @@ def __init__(self, hostname, username, password, timeout=60, optional_args=None)
         if optional_args is None:
             optional_args = {}
 
-        self.config_lock = optional_args.get("config_lock", False)
-        self.port = optional_args.get("port", 22)
-        self.key_file = optional_args.get("key_file", None)
-        self.keepalive = optional_args.get("keepalive", 30)
-        self.ssh_config_file = optional_args.get("ssh_config_file", None)
-        self.ignore_warning = optional_args.get("ignore_warning", False)
+        self.port = optional_args.get('port', 22)
```
mirceaulinic: These look reordered after resolving the conflicts with the develop branch.

ktbyers: Okay, that was mostly just black wanted the strings to be double quoted and the original code had them as single quoted (i.e. the code from the merge conflict). I fixed it so it was compliant with black requirements.

PR #889, opened 2018-12-19
Catch ValueError exception when vlan is not set to an integer.

tstoner:
When vlan is not set to an integer it is set to '-' which causes int() to throw a ValueError exception.  Adding try/except around int().


napalm/nxos/nxos.py
```diff
@@ -1186,7 +1186,10 @@ def get_mac_address_table(self):
         for mac_entry in mac_table_raw:
             raw_mac = mac_entry.get("disp_mac_addr")
             interface = py23_compat.text_type(mac_entry.get("disp_port"))
-            vlan = int(mac_entry.get("disp_vlan"))
+            try:
+                vlan = int(mac_entry.get('disp_vlan'))
+            except ValueError:
+                vlan = '-'
```
ktbyers: VLAN always needs to be an integer...so you should change the vlan assignment in the exception clause to:

```
            except ValueError:
                vlan = 0
```

Also change the quotes around 'disp_vlan' to be double quotes: "disp_vlan" (so that it is compliant with python-black-formatting.

tstoner: Thanks for the review & guidance.

PR #892, opened 2018-12-20
added expect_string support to cli() function in ios.py

mariozz:
Some Cisco IOS commands require confirmation in order to be executed.
For example:
```
clear counters
Clear "show interface" counters on all interfaces [confirm]y
```

_cli(self, commands)_ function in _ios.py_ do not permit to add a custom _expect_string_ which is used to identify next prompt. 
From netmiko documentation:

> expect_string(str) – Regular expression pattern to use for determining end of out-
> put. If left blank will default to being based on router prompt.

_expect_string_ is a paramenter of _send_command function_ in _/netmiko/base_connection.py_ (by default it is equal to None)

Since _cli()_ fuction in _ios.py_ do not feature _expect_string_ paramenter, executing the following command hangs on confirmation prompt:
`device.cli(['clear counters', 'y'])`

Adding parameters _expect_string_ to _ios.py_ _cli()_ and __send_command()_ function solve the issue.
`device.cli(['clear counters', 'y'], expect_string='[confirm]')`

In _ios.py_:
```
def cli(self, commands, expect_string=None):
        """
        Execute a list of commands and return the output in a dictionary format using the command
        as the key.

        Example input:
        ['show clock', 'show calendar']

        Output example:
        {   'show calendar': u'22:02:01 UTC Thu Feb 18 2016',
            'show clock': u'*22:01:51.165 UTC Thu Feb 18 2016'}

        """
        cli_output = dict()
        if type(commands) is not list:
            raise TypeError('Please enter a valid list of commands!')

        for command in commands:
            output = self._send_command(command, expect_string)
            if 'Invalid input detected' in output:
                raise ValueError('Unable to execute command "{}"'.format(command))
            cli_output.setdefault(command, {})
            cli_output[command] = output

        return cli_output


def cli(self, commands, expect_string=None):
        """
        Execute a list of commands and return the output in a dictionary format using the command
        as the key.

        Example input:
        ['show clock', 'show calendar']

        Output example:
        {   'show calendar': u'22:02:01 UTC Thu Feb 18 2016',
            'show clock': u'*22:01:51.165 UTC Thu Feb 18 2016'}

        """
        cli_output = dict()
        if type(commands) is not list:
            raise TypeError('Please enter a valid list of commands!')

        for command in commands:
            output = self._send_command(command, expect_string)
            if 'Invalid input detected' in output:
                raise ValueError('Unable to execute command "{}"'.format(command))
            cli_output.setdefault(command, {})
            cli_output[command] = output

        return cli_output

```


ktbyers:
My vote would be not to do this. There are other ways to accomplish this for Netmiko drivers (in Napalm) using the `napalm_object.device` which gives you the Netmiko object (and then call `send_command`).

I don't think it really makes sense to try to support this, but I also don't really ever use napalm-cli so what are other people's thoughts on this?

My main concern is I don't want to over-complicate napalm-cli by how much access/support we give to the underlying driver.

PR #897, opened 2018-12-23
Feature/ios get bgp config

jtishey:
Please review and provide feedback on IOS getter for "get_bgp_config".  Mostly cloned from the iosxr getter, but adds a new dependency for ciscoconfparse in napalm/base/helpers.py.  Also adds an "import re" in napalm/base/helpers.py.

Very likely still needs some work/testing.

```
>>> import napalm
>>> driver = napalm.get_network_driver('ios')
>>> dev = driver('IOS-R1', 'guest', 'password')
>>> dev.open()
>>> b = dev.get_bgp_config()
>>> pprint(b)
{'_': {'apply_groups': [],
       'description': '',
       'export_policy': '',
       'import_policy': '',
       'local_address': '',
       'local_as': 0,
       'multihop_ttl': 0,
       'multipath': False,
       'neighbors': {'10.0.0.1': {'authentication_key': '',
                                  'description': 'vSRX1',
                                  'export_policy': '',
                                  'import_policy': '',
                                  'local_address': 'GigabitEthernet1',
                                  'local_as': '',
                                  'nhs': False,
                                  'prefix_limit': {'inet': {'unicast': {'limit': '',
                                                                        'teardown': {'threshold': '',
                                                                                     'timeout': ''}}}},
                                  'remote_as': '65001',
                                  'route_reflector_client': False},
       'prefix_limit': {},
       'remote_as': 0,
       'remove_private_as': False,
       'type': ''}}
```


ktbyers:
Need to add ciscoconfparse to requirements.txt.

Also can you add unit tests for this?

Finally, it looks like the linters/black are failing.

jtishey:
Thanks @ktbyers .  I've updated the requirements.txt and will work on adding the unit tests and formatting properly.  Wasn't sure if the added dependency would be allowed or if there was perhaps a better way.

jtishey:
Thanks @ktbyers   it should be fixed now

ktbyers:
Added one minor change which is that we should be looking at the current running-config not the startup-config? i.e. startup-config may or may not be what the actual current configuration is and the getter should reflect the current configuration.

I think everything else is probably fine.

jtishey:
yeah, I was just using startup config for speed, but I'm sure it's better to be accurate.

jtishey:
updated for running config instead of startup

PR #898, opened 2018-12-24
optimize usage of junos XML tables to prevent greenlet starvation

### LINKED ISSUE #896: Usage of PyEZ tables in junos is not optimized
We're using celery and gevent (greenlets / green threads) as a pool implementation and we found out that the `get_interfaces()` getter is blocking a greenlet run too long when running a lot of concurrent greenlets and eventually starving the others. 

Digging into this deeper is appears that the Juniper PyEZ tables are used in a suboptimal way in that function. When acting on a PyEZ table as if it is a dictionary, performance starts crippeling.

Left: current code
Right: wrapping the table in a `dict()` call, which is 6x faster.

The example is for a device with 219 interfaces (so 438 entries in total, counting the logic interfaces)

image

jobec:
Fixes #896


jobec:
Still needs more testing.

Originally I fixed it with a 1ms sleep in the loop because that makes a greenlet move to the next instance, effectively breaking the lock.
I still need to see if this is enough of an optimization to fix things.

ktbyers:
@jobec Okay, let us know when this is ready from your perspective.

jobec:
These changes fixed the issue in our setup.
There might be other places that have the same issue, but I'll open new issues/PRs whenever I bump into one.

PR #874, opened 2018-12-03
Add VRF structure in for get_arp_table

jobec:
Related to PR #819 and issue #792

PR #900, opened 2019-01-05
Add structure for VRF support to get_arp_table

ktbyers:


ktbyers:
Updated this PR here https://github.com/napalm-automation/napalm/pull/874

ktbyers:
@jobec Can you double check that my updates are right...basically I kept the "all" for nxos_ssh driver (i.e. all VRFs) and updated the base docstring to reflect the default is all VRFs.

jobec:
@ktbyers It looks right.
I rebased my local branch on develop and I don't see anything wrong between that branch and this one.

Issue #901, opened 2019-01-07
Do Junos and EOS drivers allow for SSH host-key validation and SSL cert validation


ktbyers:
I didn't see the arguments in optional_args to enable this in napalm driver.

bewing:
pyeapi disables SSL verification if an SSL context is not passed in:

https://github.com/arista-eosplus/pyeapi/blob/develop/pyeapi/eapilib.py#L542-L545

Unfortunately, the `pyeapi.client.connect` method does not expose this.  It could be, if we switched to using `pyeapi.client.make_connection` instead, but we'd still need to pass in the kwargs correctly.

bewing:
This is exposed in EOS now, in release 2.4.0

bewing:
@ktbyers does https://github.com/napalm-automation/napalm/blob/9b66011a4235dcb1982467ee621b9c11c27f6346/napalm/junos/junos.py#L90 fulfill the JunOS SSH host-key validation knob well enough to close this issue?

PR #902, opened 2019-01-07
Refactor processing of EOS optional_args (#901)

bewing:
Process optional_args in a way that supports arbitrary underlying
arguments in the pyeapi connection class used.


ktbyers:
@bewing I assume you sanity tested this against a real device and the connection process still works properly?

I reviewed the code and that looked right.

bewing:
I have performed basic testing of the change against 7280's and 7150s.  Running the non-mock test suite shows no new failures with this change, but have provided a whole new rabbithole for me to go down.

PR #911, opened 2019-01-15
Closes #893: improve 'file prompt quiet' efficiency

### LINKED ISSUE #893: [Enhancement] Enable `file prompt quiet` only once at the beginning when configuration occurs for cisco_ios
Essentially what @ktbyers described in [this comment](https://github.com/napalm-automation/napalm/issues/879#issuecomment-448303651) of issue #879 

> On your point 1, I think we should just add an attribute to the napalm object where we track the file_prompt_quiet state. Once we change this once, then we set that flag to True. Then in the napalm close() method, if that file_prompt_quiet flag has been set to True, we re-set it back to `no file prompt quiet`.
> 
> Basically since we are making a configuration change to the device, we should only do it if we need to (i.e. we shouldn't do it in open() as then you would be making this config change even for getter operations). But if we do it for config() operations and then just track the state so we only do it once, that would be a big improvement.

I'll give it a go.

vladola:
I tested this with the following simple snippet of code:

```python


driver = get_network_driver("ios")
dev = driver('172.21.129.3', 'user', 'pass')
dev.open()
dev.load_merge_candidate(config="username xxx secret xxx")
dev.commit_config()
dev.close()
```

Before the change:

```shell
time pipenv run python testing.py
pipenv run python testing.py  1.31s user 0.89s system 2% cpu 1:48.68 total
```

After the change:

```shell
time pipenv run python testing.py
pipenv run python testing.py  1.21s user 0.79s system 2% cpu 1:08.55 total
```

A bigger gain will be obtained especially if other operations happen (e.g if one discards the changes). **At a minimum, this change will speed-up NAPALM (ios) by 40%.**

### DIFF
--- a/napalm/ios/ios.py
+++ b/napalm/ios/ios.py
@@ -124,6 +124,10 @@ class IOSDriver(NetworkDriver):
         # Control automatic toggling of 'file prompt quiet' for file operations
         self.auto_file_prompt = optional_args.get("auto_file_prompt", True)
 
+        # Define if 'file prompt quiet' has been enabled, either by NAPALM or by the user
+        self.file_prompt_quiet_napalm_enabled = False
+        self.file_prompt_quiet_user_enabled = False
+
         self.netmiko_optional_args = netmiko_args(optional_args)
 
         # Set the default port if not set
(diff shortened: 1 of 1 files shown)


vladola:
LMAO just noticed this is issue #911 - this is fate 🤣

PR #912, opened 2019-01-16
Nexus SSH: Support another magic IP address line format

mirceaulinic:
Turns out that besides the following format:
``IPv6 link-local address: fe80::2ec2:60ff:fe4f:feb2 (default) [VALID]``,
Cisco Nexus also presents the IPv6 addresses under the following format:
``IPv6 address: fe80::2ec2:60ff:fe4f:feb2 [VALID]``.

### DIFF
--- a/napalm/nxos_ssh/nxos_ssh.py
+++ b/napalm/nxos_ssh/nxos_ssh.py
@@ -903,13 +903,21 @@ class NXOSSSHDriver(NXOSDriverBase):
             #   2001:cc11:22bb:0:2ec2:60ff:fe4f:feb2/64 [VALID]
             # IPv6 subnet:  2001::/24
             # IPv6 link-local address: fe80::2ec2:60ff:fe4f:feb2 (default) [VALID]
+            # IPv6 address: fe80::a293:51ff:fe5f:5ce9 [VALID]
             if "Interface status" in line:
                 interface = line.split(",")[0]
                 continue
             if "VALID" in line:
                 line = line.strip()
                 if "link-local address" in line:
+                    # match the following format:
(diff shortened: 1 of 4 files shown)


napalm/nxos_ssh/nxos_ssh.py
```diff
@@ -903,14 +903,22 @@ def get_interfaces_ip(self):
...
                 if "link-local address" in line:
+                    # match the following format:
+                    # IPv6 link-local address: fe80::2ec2:60ff:fe4f:feb2 (default) [VALID]
                     ip_address = line.split()[3]
                     prefix_len = "64"
+                elif "IPv6 address" in line:
+                    # match the following format:
+                    # IPv6 address: fe80::a293:51ff:fe5f:5ce9 [VALID]
+                    ip_address = line.split()[2]
+                    prefix_len = "64"
```
ktbyers: Are addresses of this form always link-local? Should we put a check in where if `ip_address` contains a `/` then raise an exception. I am a bit worried that we might have a case where we incorrectly assume a /64 prefix when it isn't.

Maybe a better solution would just be to move the case where we search for '/' up (move it up from the else clause and change the logic a bit). Structurally something like:

 ```
if "link-local address" in line:
    ip_address = line.split()[3]
    prefix_len = "64"
elif "/" in line:
    ip_address, prefix_len = line.split()[0].split("/")
elif "IPv6 address" in line:
    ip_address = line.split()[2]
    prefix_len = "64"
else:
    raise ValueError("We don't know what just happened here")
```

mirceaulinic:
Yes, from what I've seen across a dozen of devices, only link-local have
that format.

PR #913, opened 2019-01-17
File prompt quiet changes to improve Cisco IOS performance

ktbyers:


ktbyers:
@vladola Yes, this solution ended up pretty similar to what you did. 

I need to check it some more as I was working on it in a pretty disjointed way yesterday (I was traveling).

vladola:
I think you did it! Commit https://github.com/napalm-automation/napalm/pull/913/commits/0a2e7ffdd105cab18837866654c6da81ca30c00f still uses 2 variables but I think that's not necessary anymore with your new arrangement.  You use `self.prompt_quiet_changed` in the `close` method but you can easily  just reference `self.prompt_quiet_configured` and delete `self.prompt_quiet_changed` altogether. 

Neat 👍

PR #917, opened 2019-01-25
Reverting IOS prompting change

ktbyers:
Upon additional testing against real devices, this PR didn't work:

https://github.com/napalm-automation/napalm/pull/913

Basically, the above code would fail if you did something like a `load_replace_candidate` and `commit` (and then did some subsequent file operations). Once the replace is committed, then the `file prompt quiet` is no longer in the config. Any subsequent file operations at that point would fail.

Will need to come up with an alternate solution.

### DIFF
--- a/napalm/ios/ios.py
+++ b/napalm/ios/ios.py
@@ -121,13 +121,8 @@ class IOSDriver(NetworkDriver):
         self._dest_file_system = optional_args.get("dest_file_system", None)
         self.auto_rollback_on_error = optional_args.get("auto_rollback_on_error", True)
 
-        # Control automatic execution of 'file prompt quiet' for file operations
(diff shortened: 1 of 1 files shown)


ktbyers:
CC @vladola

vladola:
Interesting catch. A first idea might be to just reset the flag after the replace operation. Thoughts?

ktbyers:
I am going to roll forward with this fix here:

https://github.com/napalm-automation/napalm/pull/918/files

ktbyers:
@vladola Yes, that is basically what I did...basically reset the flag to an unknown state after a commit and rollback operation such that the wrapper will re-execute the `file prompt quiet` when those events happen.

PR #922, opened 2019-01-30
Ios/get authentication

stepxy:
Added method get_auth_sessions to generate 802.1x session table for dot1x enabled switches.

### DIFF
--- a/napalm/ios/ios.py
+++ b/napalm/ios/ios.py
@@ -2773,34 +2773,22 @@ class IOSDriver(NetworkDriver):
         """
         Get 802.1x session information for interfaces.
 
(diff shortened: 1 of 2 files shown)


ktbyers:
Creating a new getter requires:

1. Creating an issue in GitHub and getting agreement we are going to do this and that the returned data structure format makes sense (including for all of the core platforms: Junos, Arista, Cisco IOS, NX-OS, and IOS-XR).

2. After 1 is done you would probably need to write the getter for at least 3 of the 5 core platforms including unit tests for it.

3. Probably a willingness for you to maintain the getter across time.

Getter maintenance is a reasonable amount of work.

If you want to do this, then we should re-start with step 1.

I am going to close this for now, but open to discussing more in a GitHub issue.

PR #923, opened 2019-02-06
New Feature: Added new get_vrrp_groups() method

e3prom:
Hello,

Here's a new getter to pull VRRP information from Cisco IOS devices. It has been tested on a recent IOS-XE platform, with multiple VRRP groups per interface. I think this may be useful as VRRP is one of the most prevalent and used FHRPs.

Let me know if there is anything I can do to improve the code, in regards to the data structure returned, the speed or code readability. I can also provide similar feature for Junos and NX-OS (if applicable).


mirceaulinic:
Hey @e3prom. Sorry for late reply here. We've been debating for a long time whether we want to accept new getters or not. I'm seeing this feature valuable, would you be able to add the equivalent changes for Junos and NX-OS as well? Thanks.

PR #937, opened 2019-02-28
IOS: add support for VRF for get_arp_table func

ExaneNetworkTeam:
Simply takes into account the vrf function arg and exec
the expected ios command.


ExaneServerTeam:
Thanks for merging this guys !

PR #938, opened 2019-02-28
FIX: fix vrf detection when no vrf is defined

ExaneServerTeam:
Right now an exception is raized when no vrf is defined.

It will now return the default instances list.

Also catch a few unhandle errors when vrf name does not exists.
A ValueError exception is now raised with an explaination.


mirceaulinic:
Hey @ExaneServerTeam - would you be able to provide a test case for this please?

mirceaulinic:
FYI: we are planning to release NAPALM 3.0.0 soon. If you'd like your patch to be included in this version, please address the comments above. Thanks!

ExaneServerTeam:
@mirceaulinic 
Hello took me sometime to find out a router with no vrf (as a few things change since my fix report).
I found one, so could able to build a test case. Basically the command 'show vrf' returns nothing.

mirceaulinic:
That's great, @ExaneServerTeam - thanks!

PR #940, opened 2019-03-07
MTU support

yeled:
Problems as I see it:

Junos `Unlimited` MTU reports as `0`, whereas all the other non-MTU interfaces on other platforms are just not reported.

Not sure if this is an interface-bug with expected data.

The rest: tests pass.


ktbyers:
You will need to get the unit tests passing.

Also you will need to update the base.py file with the new definition.

This change might also break the community drivers...which we would have to figure out how to communicate to them.

yeled:
> This change might also break the community drivers...which we would have to figure out how to communicate to them.

Can you point me at these? I'm unsure what you mean by "community"

ktbyers:
@yeled The community drivers are a coordination issue that I need to manage (so it is more something I need to track/handle)...that I wanted to document here.

mirceaulinic:
@yeled - if this hasn't been clarified on Slack or elsewhere in the meanwhile, see here the community drivers: https://github.com/napalm-automation-community

yeled:
Gentle bump, this is working for us in production environment. Is it worth merging?

ktbyers:
@yeled Yes, good job on this.

mnhauke:
While having MTU support is a good thing this change definitely broke all community drivers:
I have just tested the following drivers and they are all in a broken state with napalm 2.5 ...
* napalm-asa
* napalm-fortios
* napalm-exos
* napalm-ros
* napalm-panos

itdependsnetworks:
Ironically noticed the same thing last night. In fairness we can’t hold up progress on the main repo, but perhaps we should setup a group email for maintainers of community repo’s?

ktbyers:
Yes, I have recommended in the NTC slack that maintainers of all the community repos pin their main `napalm` version and then regression test a new release (I did this 6 months ago or so i.e. this is not new). Either that or they could nightly regression test against napalm `develop` branch.

Realistically this has to be on the owners of community drivers to fix/manage.

Near term fix is to downgrade to previous version of napalm and then create an issue/PR in each of those repositories.

ktbyers:
I am open to doing the email-list idea, but I don't think that will fix the fundamental problem.

That MTU PR was merged six months ago...so to me the fundamental question is how do we get better testing/more automated testing rolled into the community drivers (or at a minimum pinned napalm dependency so they don't break). 

In other words, how was the issue not detected and fixed by the community drivers for that length of time.

There are also going to be changes in napalm also that would break some set of community driver where the napalm maintainer would not be aware that it would break the dependent driver. In other words, this is really an automated testing problem (which includes in it making sure the automated tests are executed on a frequent enough basis).

PR #945, opened 2019-03-18
Fix broken get_bgp_neighbors when 6PE configured on eos

### LINKED ISSUE #944: get_bgp_neighbors() gets corrupted if 6PE is enabled in eos
In the case of processing of these lines, if ipv6 is invalid, there will be a gap in the number of processing lines.
As a result, local_asn regex different lines.

https://github.com/napalm-automation/napalm/blob/2e98a86bc8f967256fdc38255a856d2fae40fd68/napalm/eos/eos.py#L484-L509

In my eos environment, I confirmed neighbors where ipv6 stats like below are not displayed.

```
['BGP neighbor is 10.8.176.1, remote AS 64540, internal link', '  BGP version 4, remote router ID 10.8.176.1, VRF default', '  BGP state is Established, up for 172d03h', '    IPv4 Unicast:           0        58', 'Local AS is 38651, local router ID 10.8.176.11']
```

`v6_stats` should also be branched like `desc`'s `next_line`.

Mabuchin:
Fix #944 

This PR addresses the problem that occurs when 6PE is performed on IPv4 bgp neighbor on eos.

For 6PE Prefix displayed in ipv4 neighbor, the regular expression has been changed so that it can be stored in the ipv6 dictionary.
And adding that test.


mirceaulinic:
Thanks @Mabuchin

PR #959, opened 2019-04-10
[IOS] Raise exception if BGP neighbor info requested and BGP not running

gjim83:
per chat with @ktbyers a more common exception could be implemented for protocol/feature info requested on device not running said proto/feature


gjim83:
One comment: currently `get_bgp_neighbors_detail` actually returns the `defaultdict(<function IOSDriver.get_bgp_neighbors_detail.<locals>.<lambda> at 0x110b770d0>, {})` when a device isn't running BGP. This would change that behavior, but IMO it's still better than this obscure-looking `defaultdict`

ktbyers:
Replaced by the PR here which fixes the black issue:

https://github.com/napalm-automation/napalm/pull/970

PR #962, opened 2019-04-15
fix for empty parent_table

carlniger:
this fixes an issue where napalm gets back null from nxapi for some getter -- in this case I ran into this with "get_interfaces_ip" where no ip interfaces were created (mgmt interface being DHCP seems to cause that not to show up).


carlniger:
Just kidding, already fixed in develop. closing :)

PR #971, opened 2019-04-24
Add ping and traceroute capabilities to mock

brandomando:
The `mock` driver does not currently implement `ping` or `traceroute` as mocked_modules. This change adds that functionality to the driver.  

An additional change is to use `ast.listeral_eval()` in place of `json.loads()` to pull in mocked data within `mock.mocked_data` function. This is to allow dicts with integer keys to be loaded, and also to allow data exported directly from napalm to be saved as mock data without modification, such as using double quotes, or string keys as `json.loads()` requires natively.


napalm/base/mock.py
```diff
@@ -68,7 +68,7 @@ def mocked_data(path, name, count):
     filename = "{}.{}".format(os.path.join(path, name), count)
     try:
         with open(filename) as f:
-            result = json.loads(f.read())
+            result = ast.literal_eval(f.read())
```
mirceaulinic, line 71: Hey @brandomando - only now looking at your PR (sorry!). Can you explain why do you think this change is necessary? That'd be a breaking change for many users that have the mock data in JSON format, e.g.,

```python
>>> txt = '{"test": null}'
>>> json.loads(txt)
{'test': None}
>>> ast.literal_eval(txt)
  File "<console>", line 1, in <module>
  File "/usr/lib/python3.6/ast.py", line 85, in literal_eval
    return _convert(node_or_string)
  File "/usr/lib/python3.6/ast.py", line 66, in _convert
    in zip(node.keys, node.values))
  File "/usr/lib/python3.6/ast.py", line 65, in <genexpr>
    return dict((_convert(k), _convert(v)) for k, v
  File "/usr/lib/python3.6/ast.py", line 84, in _convert
    raise ValueError('malformed node or string: ' + repr(node))
ValueError: malformed node or string: <_ast.Name object at 0x7f426efe8518>
>>> 
```

brandomando:
Sorry I totally missed the notification for this

PR #973, opened 2019-04-25
Implement nxos_ssh get_environment

ktbyers:


ktbyers:
Updated and fixed some rebasing issues from earlier referenced PR.

PR #978, opened 2019-05-07
Set device enable just only if have "secret" options

User-Man:
Some time connect to cisco a read-only user and don't need set enable every time.


ktbyers:
The thing I don't really like about is that it will cause failures with configuration operations and some set of getter methods to happen in more obscure ways. In other words, they will require elevated privileges, not have them and then fail with hard to decipher messages.

It would probably be better to add an optional argument that you have to pass in at init for something like `force_no_enable` in other words not have this be the default behavior.

User-Man:
Something like that?

User-Man:
also, for some cisco models (expample: me3400, me3800) and with user limited rights to get config need use "show running-config full" or "show running-config view full"

PR #982, opened 2019-05-16
Add support system host keys

TomCos:
Saw it was missing from the support doc.


docs/support/index.rst
```diff
@@ -127,6 +127,7 @@ ____________________________________
 * :code:`ssh_config_file` (ios, iosxr, junos, nxos_ssh) - File name of OpenSSH configuration file.
 * :code:`ssh_strict` (ios, iosxr, nxos_ssh) - Automatically reject unknown SSH host keys (default: ``False``, which means unknown SSH host keys will be accepted).
 * :code:`ssl_verify` (nxos) - Requests argument, enable the SSL certificates verification. See requests ssl-cert-verification for valide values (default: ``None`` equivalent to ``False``).
+* :code:`system_host_keys` (ios, iosxr, nxos_ssh) - Load host keys from the user's 'known_hosts' file (default: ``False``).
```
mirceaulinic: Hey @TomCos - thanks for this PR. It doesn't seem to me like this option is currently supported at all yet - see, e.g., https://github.com/napalm-automation/napalm/blob/develop/napalm/ios/ios.py.
Or were you willing to contribute it? :-)

ktbyers: I think it is supported as all Netmiko connection options are supported on IOS and NXOS_SSH, but I am not sure it is worth enumerating all of them. Probably better to just document that i.e. that all Netmiko connection arguments are supported via optional_args and reference that (with exception of NAPALM primary arguments like hostname which get mapped to their corresponding Netmiko value).

TomCos: Either is fine, I can close this out if we want to do a more general note.

ktbyers: @TomCos Sounds good

PR #984, opened 2019-05-21
Force no enable

User-Man:
Add optional arguments force_no_enable for disable use enable mode.
Some time connect to cisco a read-only user and don't need set enable every time.


ktbyers:
It should also support NX-OS SSH which uses Netmiko and maybe IOS-XR. IOS-XR will have to look into on whether it makes sense there.

mirceaulinic:
What's the status on this one?

mirceaulinic:
I'm checking in here - @User-Man have you seen Kirk's comment above?

PR #985, opened 2019-05-21
additional options for get_config()

User-Man:
For some cisco models (expample: me3400, me3800) and with user limited rights to get full config
need use "show running-config full" or "show running-config view full" commands.


ktbyers:
You would also need to update `base.py` and all the other drivers to support the extra argument. 

You can see from the above change that unit tests are failing.

You should also have a more specific argument name.

mirceaulinic:
I think this might have been done in https://github.com/napalm-automation/napalm/pull/1029, please reopen otherwise to discuss.

Issue #637, opened 2018-01-29
when using nxos_ssh when changing hostname receive error


omershtivi:
When used nxos_ssh changing device hostname I received the following issue:
`fatal: [LAB_POD2_1]: FAILED! => {"changed": false, "failed": true, "msg": "cannot install config: Search pattern never detected in send_command_expect: LAB\\_POD2\\_1\\-TEST\\#"}`

Rerun the playbook Password OK

The playbook:
```
omers@Ubuntu-Ansible-CTL:~/Naplam-Tests$ more Playbooks/Basic-Config-Checkpoint-Dryrun.yml
---
- name: Configuration replace
  hosts: N9K
  connection: local
  gather-facts: False
  tasks:
      - name: Generate-Configuration N9K
        template: src=../Templates/N9K-Checkpoint.j2 dest=../Configuration/{{ inventory_hostname }}

      - name: N9K Configuration Replace Check
        napalm_install_config:
             hostname: "{{ ip_address }}"
             username: "{{ username }}"
             password: "{{ password }}"
             dev_os: "{{ dev_os }}"
             config_file: ../Configuration/{{ inventory_hostname }}
             commit_changes: False
             replace_config: True
             diff_file: ../Diff/initial.diff
```

lazhao123:
I got same issue on Linux when I config cisco nxos. I am using NAPALM version 2.4.0.
The script call commit_config() to change hostname.
The bug is caused by self.base_prompt, since hostname is changed, while base_prompt is not changed yet. That caused pattern not match, and timeout. Is anyone working on this?

    def _read_channel_expect(self, pattern="", re_flags=0, max_loops=150):
        output = ""
        if not pattern:
            pattern = re.escape(self.base_prompt)           <=== base_prompt not match with hostname.
        log.debug("Pattern is: {}".format(pattern))

traceback
=========================
  File "/root/learn_lab/net-auto/switch_dev.py", line 86, in <module>
    nxos_dev.set_hostname('N3k-1')
  File "/root/learn_lab/net-auto/switch_dev.py", line 52, in set_hostname
    self._driver.commit_config()
  File "/root/global3_venv/lib64/python3.6/site-packages/napalm/nxos/nxos.py", line 224, in commit_config
    self._copy_run_start()
  File "/root/global3_venv/lib64/python3.6/site-packages/napalm/nxos_ssh/nxos_ssh.py", line 496, in _copy_run_start
    output = self.device.save_config()
[...]

PR #990, opened 2019-06-07
FIX: Changing hostname using nxos_ssh causing Netmiko timeout

carlniger:
Had a more complicated "fix" involving a "_commit_handler" similar to the IOS setup, but I believe this fixes the underlying issue more simply. See #637


ktbyers:
I found some additional issues so I made an updated PR here:

https://github.com/napalm-automation/napalm/pull/999

PR #992, opened 2019-06-07
Create NOTICE

eckelcu:


ktbyers:
@eckelcu Why? 

If we do that we should include the Apache2 notice right below that as all the code in here is under the Apache2 license. Also Spotify only is a relatively small part of the code at this point.

eckelcu:
This is an attempt to satisfy this requirement of applying the Apache 2.0 license to your code, https://www.apache.org/licenses/LICENSE-2.0#apply

Admittedly, it would be best to add the Apache 2.0 license boilerplate to every source file, but short of doing that, I am proposing to identify the copyright owner in a single NOTICE file. 

I saw the Spotify copyright in some other files, e.g. https://github.com/napalm-automation/napalm/blob/develop/napalm/base/base.py, so I went with that.

eckelcu:
Here is an example from Apache of a notice file https://apache.org/licenses/example-NOTICE.txt

ktbyers:
Okay, we should also probably add the set of committers to it as well (as Spotify contributed very little of the code at this point). DBarroso has committed a lot, but I think most of it was probably post spotify.

mirceaulinic:
Checking in here @eckelcu, any chance to update as per Kirk's suggestion?

eckelcu:
Happy too. I have updated as suggested @ktbyers @mirceaulinic

NOTICE
```diff
@@ -0,0 +1,7 @@
+napalm
+Copyright 2015 Spotify AB. All rights reserved.
```
mirceaulinic, line 2: This is a matter I thought about it, seems still unclear.
What about adding another line here like:

```
Copyright 2016-2020 NAPALM Automation Team. All rights reserved.
```

Any thoughts @ktbyers @dbarrosop?

eckeltest: I am not a lawyer, but my understanding is there should be a single copyright notice only and that identifies the copyright owner at the time the work was first published. The owner should be a legal entity (a person, their employer, an organization, etc.)

eckelcu: I am not a lawyer, but my understanding is there should be a single copyright notice only and that identifies the copyright owner at the time the work was first published. The owner should be a legal entity (a person, their employer, an organization, etc.)

dbarrosop:
I am no lawyer either and I have no strong opinons but I don't think a NOTICE file is needed. Apache 2 license only requires a LICENSE file in the root of the project and the rest of the content, unless specified otherwise, inherits this LICENSE. From the LICENSE itself:

> If the Work includes a "NOTICE" text file as part of its distribution, then any Derivative Works that You distribute must include a readable copy of the attribution notices contained within such NOTICE file, excluding those notices that do not pertain to any part of the Derivative Works, in at least one of the following places: within a NOTICE text file distributed as part of the Derivative Works; within the Source form or documentation, if provided along with the Derivative Works; or, within a display generated by the Derivative Works, if and wherever such third-party notices normally appear. **The contents of the NOTICE file are for informational purposes only and do not modify the License.** You may add Your own attribution notices within Derivative Works that You distribute, alongside or as an addendum to the NOTICE text from the Work, provided that such additional attribution notices cannot be construed as modifying the License.

PR #994, opened 2019-06-10
Better handling of connection closing upon exceptions in __enter__

### LINKED ISSUE #991: Device connection not closed upon exception in __enter__() method
### Description of Issue

Observed while connecting to a Juniper device over SSH with the Cisco IOS driver. (so wrong driver for the target platform)

When using Napalm as a context manager (or by manually calling `open()` and `close()`) an exception might be raised during the `__enter__()` method. Like for example when enable mode can't be reached.

When this happens, 2 issues arise:

1) When called as a context manager, the code in `__exit__()` is never called and the connection to the device remains open for the remainder of the program.

2) When called with `open()` and `close()`, the connection cannot be closed anymore in `close()` because `self.device` isn't set yet.

Open:
https://github.com/napalm-automation/napalm/blob/e0cc9af966230bcc069cd3ab7e397e906efc04c5/napalm/base/base.py#L81
Close:
https://github.com/napalm-automation/napalm/blob/e0cc9af966230bcc069cd3ab7e397e906efc04c5/napalm/base/base.py#L98

When running inside something like Celery or Django, where the program never closes, the connections remain open indefinitely and eventually occupy all SSH sessions on the target device if it's not configured to close the connection after an idle timeout.

### napalm version
2.4.0

Use the IOS driver to connect to a Juniper device

jobec:
Always try to close netmiko connections.

Closes #991

### DIFF
--- a/napalm/base/base.py
+++ b/napalm/base/base.py
@@ -16,14 +16,16 @@
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import sys
+
+from netmiko import ConnectHandler, NetMikoTimeoutException
+
 # local modules
 import napalm.base.exceptions
-from napalm.base.exceptions import ConnectionException
 import napalm.base.helpers
 from napalm.base import constants as c
 from napalm.base import validate
-
-from netmiko import ConnectHandler, NetMikoTimeoutException
+from napalm.base.exceptions import ConnectionException
 
 
 class NetworkDriver(object):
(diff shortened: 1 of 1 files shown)


jobec:
Might be I'm seeing some side effects of this change.
Will test some more and see if they are related.

jobec:
The side effect is that all drivers need to be able to handle a non-initialized state in their `close()` function. Because it might be called even if there's no connection made already.

I checked all core drivers, and they seem to be fine (on the assumption that the external packages like pyez etc handle it).

It might cause some turmoil with community drivers though.

ktbyers:
This looks right to me.

Drivers (including community drivers) will have to make a best effort to cleanup the connection on a close() call including one that occurs on a non-completed open() call.

@jobec I think we should probably merge this--any objections?

jobec:
No objections.

My tests don't give any issues.

PR #999, opened 2019-06-18
Fix nxos and nxos_ssh hostname change handling

ktbyers:
I ran into issues with both nxos driver and nxos_ssh driver with hostname changes. Made fixes here to incorporate. 

I also updated the napalm_custom_test suite to test this for both nxos and nxos_ssh for both merge and replace operations. This generally runs daily against virtual network devices and the develop branch.

### DIFF
--- a/napalm/nxos/nxos.py
+++ b/napalm/nxos/nxos.py
@@ -218,6 +218,8 @@ class NXOSDriverBase(NetworkDriver):
 
             if self.replace:
                 self._load_cfg_from_checkpoint()
+                # If hostname changes ensure Netmiko state is updated properly
+                self._netmiko_device.set_base_prompt()
             else:
                 self._commit_merge()
 
--- a/napalm/nxos_ssh/nxos_ssh.py
+++ b/napalm/nxos_ssh/nxos_ssh.py
@@ -533,11 +533,9 @@ class NXOSSSHDriver(NXOSDriverBase):
         ]
 
         try:
-            rollback_result = self._send_command_list(commands, expect_string=r"#")
+            rollback_result = self._send_command_list(commands, expect_string=r"[#>]")
         finally:
             self.changed = True
-        # If hostname changes ensure Netmiko state is updated properly
-        self.device.set_base_prompt()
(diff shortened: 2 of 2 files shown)


lazhao123:
The bug has not fixed yet. I changed hostname with load_merge_candidate( ), the base_prompt is not changed, so check_enable_mode( ) still search old hostname and timeout. Attached stack trace.

            if self.replace:
                self._load_cfg_from_checkpoint()
                # If hostname changes ensure Netmiko state is updated properly
                self._netmiko_device.set_base_prompt()         <========= should call when merge config
            else:
                self._commit_merge()

image

lazhao123:
move set_base_prompt( ) before self._copy_run_start( ) fixed it.
        if self.loaded:
            # Create checkpoint from current running-config
            self._save_to_checkpoint(self.rollback_cfg)

            if self.replace:
                self._load_cfg_from_checkpoint()
            else:
                self._commit_merge()

            # If hostname changes ensure Netmiko state is updated properly
            self._netmiko_device.set_base_prompt()
            self._copy_run_start()

lazhao123:
After rollback( ), prompt gets changed, but "copy_run_start( )" still look for old prompt, that caused timeout.
So I added set_base_promt( ) after rollback to fixed it.

image

    def rollback(self):
        if self.changed:
            command = "rollback running-config file {}".format(self.rollback_cfg)
            result = self._send_command(command)
            if "completed" not in result.lower():
                raise ReplaceConfigException(result)
            # If hostname changes ensure Netmiko state is updated properly
            self._netmiko_device.set_base_prompt()
            self._copy_run_start()
            self.changed = False

ktbyers:
@lazhao123 Okay, I opened the referenced new issue. 

Yes, this is still broke. My test case was working for merge, but it was  because I chose poorly on the new hostname (i.e. it had part of the old prompt in the new prompt name).

There is a good chance your fix for the merge part of this won't work (i.e. I think that is probably going to break NX-API as there is no Netmiko connection there for a merge operation...need to double check this though).

PR #1003, opened 2019-06-24
Fix 'Edit in GitHub' button and sync with Read the Docs settings

jdillard:
This changes the "Edit in GitHub" button from editing `master` to `develop`. **Disclaimer:** I've never used Read the Docs so I don't know if adding this setting in **conf.py** will fix the button, break something, or do nothing. It fixes it for local builds though.

This also changes the version in **conf.py** from `0` to `latest` so that the local build matches the live build more closely. Again, I don't know if this will break something on RTD, but it works on local builds.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -62,7 +62,7 @@ copyright = u"2016, David Barroso"
 # built documents.
 #
 # The short X.Y version.
-version = "0"
+version = "latest"
 # The full version, including alpha/beta/rc tags.
 release = "1"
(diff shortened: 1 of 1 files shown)


jdillard:
Ah, my bad I just realized I forgot to check that you publish three versions of the docs...

This should probably be closed, but I'm not sure showing the "Edit in GitHub" button is ideal to leave turned on it if it is going to break the workflow in most cases.

Issue #1000, opened 2019-06-19
get_environment fails on cisco 2960


Tolisan:
### Description of Issue/Question

*Note*: Please check https://guides.github.com/features/mastering-markdown/
to see how to properly format your request.

### Did you follow the steps from https://github.com/napalm-automation/napalm#faq
(Place an ``x`` between the square brackets where applicable)

- [ x ] Yes
- [ ] No

### napalm version
(Paste verbatim output from `pip freeze | grep napalm` between quotes below)

```
napalm==2.4.0
```

### Network operating system version
(Paste verbatim output from `show version` - or equivalent - between quotes below)

```
15.0(2)SE6
12.2(50)SE5
12.2(58)SE2
```

### Steps to Reproduce the Issue

Initialize nornir and run a Napalm getter task for a devices environment and the following traceback error is provided

### Error Traceback
(Paste the complete traceback of the exception between quotes below)

  File "/home/tolisan/Desktop/automation-project/automation/lib/python3.6/site-packages/napalm/ios/ios.py", line 1880, in get_environment
    temp_value = float(line.split(":")[1].split()[0])
ValueError: could not convert string to float: 'Not'

________________________________________________________________

Tested my script on next models:
WS-C2960-48PST-L
WS-C2960-48TC-L
when get_environment executing "show env temperature status" CLI output return string
"Temperature Value: Not Supported" or "System Temperature Value: Not Supported" depends on IOS version. Because method cant convert string to float it's return exception.
________________________________________________________________
Possible solution.

If CLI output return string like "not supported" returh dict with 
{ 'is allert': False, 
  'is critical': False,
  'temperature': 'Sensor_not_supported'}}}}
________________________________________________________________

environment.setdefault("temperature", {})
        re_temp_value = re.compile("(.*) Temperature Value\W\s\d+")
        re_temp_not_suppoted = re.compile("(.*)Not Supported")
        # The 'show env temperature status' is not ubiquitous in Cisco IOS
        output = self._send_command(temp_cmd)
        if "% Invalid" not in output:
            for line in output.splitlines():
                m = re_temp_value.match(line) 
                n = re_temp_not_suppoted.match(line)
                if n:
                    temp_name = n.group(1).lower()
                    temp_value = "Sensor_not_supported"
                    env_value = {
                        "is allert": False,
                        "is critical": False,
                        "temperature": temp_value,
                    } 
                    environment["temperature"][temp_name] = env_value   
                elif m:
[...]

PR #1013, opened 2019-07-01
Check if temperature sensor on Cisco device is not supported

TheRealBecks:
Some older Cisco devices like models from the 2960 platform have no temperature sensor built-in. In that case the get_environment function will throw on exception. That case will now be handeled.


TheRealBecks:
See issue [1000](https://github.com/napalm-automation/napalm/issues/1000)

TheRealBecks:
Has anyone an idea why four out of six checks fail? Looks like tox isn't working properly.
And can anyone review/test my change? Not sure if I did correct.

napalm/ios/ios.py
```diff
@@ -2110,7 +2110,7 @@ def get_environment(self):
         re_temp_value = re.compile("(.*) Temperature Value")
         # The 'show env temperature status' is not ubiquitous in Cisco IOS
         output = self._send_command(temp_cmd)
-        if "% Invalid" not in output:
+        if "% Invalid" and "Not Supported" not in output:
```
mirceaulinic: I think `if "% Invalid" not in output or "Not Supported" not in output:` might be what you want.

TheRealBecks: @mirceaulinic You are 'more right' than me, but still not entirely correct: It's `if "% Invalid" not in output and "Not Supported" not in output:` - `and` instead of `or` :) I will try to update my PR, not sure how to do that...

TheRealBecks:
I branched my repository, so I had to recreate this PR, see #1055.

PR #1025, opened 2019-07-29
Incorrectly taken serial number in NXOS.

VictorPavlushin:
Taken "Processor Board ID".
In some devices, it is equal to the chassis serial number, while others do not.
But inventory and technical support comes from the chassis serial number.

```
show inventory
{
   "TABLE_inv": {
     "ROW_inv": [
       {
         "name": "Chassis",
         "desc": "Nexus 6001 Chassis",
         "productid": "N6K-C6001-64P",
         "vendorid": "V01",
         "serialnum": "FOC000000000"
       },
      {
        "name": "Module 1",
        "desc": "Nexus 64 Supervisor",
        "productid": "N6K-C6001-64P",
        "vendorid": "V01",
        "serialnum": "FOC11111111"
      },
.....
```
simpler device:
```
"TABLE_inv":{
    "ROW_inv":[
    {
        "name":"\"Chassis\"",
        "desc":"\"Nexus9000 C92160YC-X chassis\"",
        "productid":"N9K-C92160YC-X",
        "vendorid":"V02",
        "serialnum":"FDO00000000"
    },
    {
        "name":"\"Slot 1\"",
        "desc":"\"48x10G + 4x40G + 2x100G Ether net Module\"",
        "productid":"N9K-C92160YC-X",
        "vendorid":"V02",
        "serialnum":"FDO00000000"
    },
......
```


napalm/nxos/nxos.py
```diff
@@ -779,10 +779,20 @@ def get_facts(self):
...
+
+        for row in show_inventory_table:
+            if row["name"] == "Chassis":
+                facts["serial_number"] = row.get("serialnum", "")
+                break
+
         show_version = self._send_command("show version")
         facts["model"] = show_version.get("chassis_id", "")
         facts["hostname"] = show_version.get("host_name", "")
-        facts["serial_number"] = show_version.get("proc_board_id", "")
```
mirceaulinic: I'd say that `serial_number` should always be there, while the loop above presents the risk of returning an output that doesn't contain it. Might want to add a line `facts['serial_number'] = None` just above maybe?

napalm/nxos_ssh/nxos_ssh.py
```diff
@@ -584,16 +585,23 @@ def get_facts(self):
...
 
+        show_inventory_table = self._get_command_table(
+            "show inventory | json", "TABLE_inv", "ROW_inv"
+        )
+        if isinstance(show_inventory_table, dict):
+            show_inventory_table = [show_inventory_table]
+
+        for row in show_inventory_table:
+            if row["name"] == "Chassis":
+                serial_number = row.get("serialnum", "")
```
mirceaulinic: Similar comment as above (in `nxos.py`).

napalm/nxos_ssh/nxos_ssh.py
```diff
@@ -1487,3 +1495,34 @@ def get_users(self):
                     continue
                 users[username]["sshkeys"].append(py23_compat.text_type(sshkeyvalue))
         return users
+
+    @staticmethod
+    def _get_table_rows(parent_table, table_name, row_name):
```
mirceaulinic: I was under the impression that this function is already defined in https://github.com/napalm-automation/napalm/blob/develop/napalm/nxos/nxos.py#L705, we might want to have it as a helper function and use it in both drivers, rather than duplicating the code.

napalm/nxos/nxos.py
VictorPavlushin: Added to code

napalm/nxos_ssh/nxos_ssh.py
VictorPavlushin: Already defined in line 580

napalm/nxos_ssh/nxos_ssh.py
VictorPavlushin: Can transfer these three functions to the NXOSDriverBase class?
Or there are other suggestions ....

mirceaulinic: Correct, this is where I would transfer those functions @VictorPavlushin.

VictorPavlushin: Done

Ichabond:
This added a regression:
Cisco Nexus 9000 now reports `None` for serial number, as there appears to be some double quoting going on:
```
{   u'show inventory': {   u'TABLE_inv': {   u'ROW_inv': [   {   u'desc': u'"Nexus9000 93180YC-EX chassis"',
                                                                 u'name': u'"Chassis"',
```

So the `if` test should be:
```            if row["name"] == '"Chassis"':```

VictorPavlushin:
PR [https://github.com/napalm-automation/napalm/pull/1049]

PR #1030, opened 2019-08-04
prototype of ntc_rosetta integration

dbarrosop:
More progress, I think this pretty much illustrate how easy is to integrate rosetta with napalm and write setters and getters for with it.

Quick and dirty demo:

https://github.com/napalm-automation/napalm/blob/yang/docs/tutorials/yang/yang.ipynb

Note that for the demo I had to use a ceos container pretending to be an IOS device because I don't have access to my lab right now :(

TODO:

- address potential feedback
- tests


napalm/eos/eos.py
```diff
@@ -85,7 +87,16 @@ class EOSDriver(NetworkDriver):
...
 
-    def __init__(self, hostname, username, password, timeout=60, optional_args=None):
+    def __init__(
+        self,
+        hostname,
+        username,
+        password,
+        timeout=60,
+        rosetta_driver=None,
+        yang_model="openconfig",
```
dbarrosop: we could also set this to None by default and avoid loading the rosetta driver, that'd save some memory

ogenstad:
One thing here is that ntc_rosetta is written for Python 3.6 only whereas Napalm still is 2.x/3.x. One solution could be to use conditional imports for the ntc_rosetta stuff, on the other hand, it might just confuse things as it would need to be stated in the documentation that some parts of Napalm are only available under Python 3.6.

I guess now would be a good time to drop Python 2.x support, then if Napalm is ready to move to 3.6 only and above the imports won't be a problem.

ktbyers:
NAPALM will probably only have one more PY2 release and then will migrate to PY3.6 or greater. The  last PY2 release will probably be very close to what is currently there.

Netmiko will do the same...there is a good chance I won't do any additional PY2 releases in Netmiko and everything in Netmiko going forward will be >=PY3.6.

napalm/base/base.py
```diff
@@ -28,8 +28,49 @@
 from napalm.base.exceptions import ConnectionException
 
 
+yang_filters = {
```
mirceaulinic: I was just wondering, would it be possible to make this more dynamic and extensible in your own environment - without a fork, that is.

napalm/base/base.py
```diff
@@ -28,8 +28,49 @@
...
+
 class NetworkDriver(object):
-    def __init__(self, hostname, username, password, timeout=60, optional_args=None):
+    def __init__(
+        self,
+        hostname,
+        username,
+        password,
+        rosetta_driver=None,
+        yang_model="openconfig",
```
mirceaulinic: I'm thinking maybe there are other ways to select across different standardization orgs within the same session, i.e., without re-initializing the driver? I think that'd add a pretty good degree of flexibility.

napalm/eos/eos.py
mirceaulinic: Ah, only now seeing this comment. Left a comment around this line on the base driver. I think that'd make sense, and somewhat related to my suggestion.

napalm/ios/ios.py
```diff
@@ -97,7 +99,16 @@
...
     """NAPALM Cisco IOS Handler."""
 
-    def __init__(self, hostname, username, password, timeout=60, optional_args=None):
+    def __init__(
+        self,
+        hostname,
+        username,
+        password,
+        timeout=60,
+        rosetta_driver=None,
```
mirceaulinic: Hmm... what does this argument do? :thinking: I've seen it at the base driver, and assumed that here we pass the name of the NAPALM driver, but now seeing it here, I'm not following anymore (again, I'm not familiar with ntc-rosetta still, and not sure what driver means in that context :sweat_smile:).

napalm/base/base.py
dbarrosop: This is basically our own implementation of "getters" so you could say it's our "business logic". So "no", I am afraid this can't be generated dynamically.

napalm/base/base.py
dbarrosop: Maybe, for the getters and setters it's going to be easier to have a default model associated to it but as an enhancement we could add a parameter to them to override the default one or have a way to override it, something like `device.set_yang_model("something")`.  In any case, I'd suggest treating that as a post enhancement to avoid complicating this initial PR.

napalm/ios/ios.py
dbarrosop: A driver in the context of rosetta is an object that knows how to parse/translate your device's native configuration/state. Ideally, napalm would figure out which driver to use but you may choose to override it yourself if you want to.

mirceaulinic:
Great, now `develop` is Python 3 only. @dbarrosop could you please rebase this PR?

ktbyers:
I will probably do work on this related to `incendo` so depending on `incendio` integrates to `napalm` this will  be there.

Or worded differently, I will require the `rosetta/yangify` integration in `incendio` so I will have it there (and consequently napalm might bind to it there).

I was working on it a few weeks ago, but need to circle back to it.

mirceaulinic:
Sounds good to me, @ktbyers. If David agrees, I'm going to close this PR. Thanks! :smile:

Issue #1026, opened 2019-07-29
get_bgp_config get non committed configuration on JunOS


kpetremann:
### Description of Issue/Question

get_bgp_config is getting the configuration from the candidate configuration (not committed) instead of the real configuration on JunOS.

### Did you follow the steps from https://github.com/napalm-automation/napalm#faq

- [x] Yes
- [ ] No

### napalm version
(Paste verbatim output from `pip freeze | grep napalm` between quotes below)

```
napalm==2.4.0
```

### Network operating system version
(Paste verbatim output from `show version` - or equivalent - between quotes below)

```
Model: qfx5100-48s-6q
Junos: 17.4R1-S3.3
JUNOS Base OS boot [17.4R1-S3.3]
JUNOS Base OS Software Suite [17.4R1-S3.3]
JUNOS Crypto Software Suite [17.4R1-S3.3]
JUNOS Online Documentation [17.4R1-S3.3]
JUNOS Crypto Software Suite [17.4R1-S3.3]
JUNOS Kernel Software Suite [17.4R1-S3.3]
JUNOS Packet Forwarding Engine Support (qfx-ex-x86-32) [17.4R1-S3.3]
JUNOS Routing Software Suite [17.4R1-S3.3]
JUNOS jsd [i386-17.4R1-S3.3-jet-1]
JUNOS SDN Software Suite [17.4R1-S3.3]
JUNOS Enterprise Software Suite [17.4R1-S3.3]
JUNOS py-base-i386 [17.4R1-S3.3]
JUNOS py-extensions-i386 [17.4R1-S3.3]
JUNOS Host Software [17.4R1-S3.3]
Junos for Automation Enhancement
```

### Steps to Reproduce the Issue

**Script:**

```
import napalm

driver = napalm.get_network_driver('junos')
device1 = driver(hostname="", username="", password="")

device1.open()
device1.get_bgp_config()
```

**Execution:**

1. When there is no configuration in protocols bgp:

```# run show configuration protocols bgp
> empty
```

2. Script output is ok:
```
$ python test.py
{}
```

3. We load configuration in the Juniper device without committing it:
```
# load patch terminal
[Type ^D at a new line to end input]
[edit protocols]
+   bgp {
+       group TOTO {
+           neighbor 169.254.254.169;
+       }
+   }
ctrl+D

# show | compare   
[edit protocols]
+   bgp {
+       group TOTO {
+           neighbor 169.254.254.169;
+       }
+   }

[...]

PR #1031, opened 2019-08-08
fix junos issue #1026

ckishimo:
Fix for issue #1026 so Junos BGP config is retrieved from the committed config instead of the candidate

### DIFF
--- a/napalm/junos/junos.py
+++ b/napalm/junos/junos.py
@@ -1086,13 +1086,15 @@ class JunOSDriver(NetworkDriver):
         _DATATYPE_DEFAULT_ = {py23_compat.text_type: "", int: 0, bool: False, list: []}
 
         bgp_config = {}
+        # Retrieve configuration from running config (not candidate)
+        options = {"database": "committed"}
 
         if group:
             bgp = junos_views.junos_bgp_config_group_table(self.device)
-            bgp.get(group=group)
+            bgp.get(group=group, options=options)
         else:
             bgp = junos_views.junos_bgp_config_table(self.device)
-            bgp.get()
+            bgp.get(options=options)
             neighbor = ""  # if no group is set, no neighbor should be set either
         bgp_items = bgp.items()
 
@@ -1105,7 +1107,7 @@ class JunOSDriver(NetworkDriver):
         # The resulting dict (nhs_policies) will be used by _check_nhs to determine if "nhs"
         # is configured or not in the policies applied to a BGP neighbor
         policy = junos_views.junos_policy_nhs_config_table(self.device)
-        policy.get()
+        policy.get(options=options)
         nhs_policies = dict()
         for policy_name, is_nhs_list in policy.items():
             # is_nhs_list is a list with one element. Ex: [('is_nhs', True)]


ktbyers:
@ckishimo Can you fix the linting i.e. use Python black on the file that is failing.

You probably just need to use double quotes on that line 1090.


Regards, Kirk

napalm/junos/junos.py
```diff
@@ -1086,13 +1086,15 @@ def build_prefix_limit(**args):
         _DATATYPE_DEFAULT_ = {py23_compat.text_type: "", int: 0, bool: False, list: []}
 
         bgp_config = {}
+        # Retrieve configuration from running config (not candidate)
+        options = {'database': 'committed'}
```
mirceaulinic: We have other methods that retrieve configuration, what if we have an optional argument allowing you to select from what config DB to select (of course, defaulting to `committed`). Don't mean to block this PR, IMO we should go ahead with this fix as-is, but wanted to have this discussion.

ktbyers: I don't really think the getters should be from the pending state (i.e. should be from the running config state). 

I also think it is probably too much to track for this getter i.e. we need to parse not just the running state, but also the pending state. This might be easy for Juniper, but probably not easy for some of the other NAPALM platforms (and probably not worth the pain that will entail on the other platforms). 

Open to more discussion on why it would be valuable, however.

mirceaulinic: I'm looking at this from the following angle: we have other methods that pull data from the configuration. With this change as-is, `get_bgp_config` becomes the only one gathering from the committed config, while the rest would continue to return the pending config. All I'm suggesting is to make sure we're consistent across the existing methods.

At the same time, with this change, we're changing the current behaviour, so we'd be just one step away from providing a way to switch back to the previous behaviour (if the user desires) through an optional arg - which would be a trivial implementation to gate this flexibility.
It would indeed be available to Juniper only, but it wouldn't be the first optional arg dedicated to a specific platform; our convention is to prefix the optional arg with the platform name in such case, e.g., `junos_config_database` - defaulting to `committed`.

ktbyers: What other getters are pulling from `candidate configuration`? I would have said almost all other napalm getters are using operational state (i.e. ultimately based on running config). 

Though maybe this difference comes from Juniper/IOS-XR way of thinking about things (i.e. where candidate config is more inherent in the thinking). I doubt we are doing this anywhere in the IOS/NX-OS drivers.

I checked EOS's implementation of get_bgp_neighbors and it is using `running-config`. I am not sure about IOS-XR (maybe you know off the top of your head). NX-OS and IOS are not implemented for this getter.

The main thing I don't like is it will be very hard to ever implement this behavior on IOS and NX-OS so we will have built-in inconsistencies in the getters.

I am okay with keeping this as an argument to the method though it is very likely going to get stranded i.e. very unlikely to ever to be implemented on other platforms (IOS, NX-OS). 

I would argue the default should be the running-config, but we could also look to gracefully transition it at napalm3 (since it is a change in behavior on Junos).

Does that work from your perspective?
1. Change the default for Junos to be running-config.
2. Add an argument to the getter that allows you to call get_bgp_config and specify either candidate, startup, or running).
3. potentially transition the behavior at napalm3 (if the new behavior change is going to be problematic). 


Kirk

ktbyers: @ckishimo Feel free to comment here as well since you are doing quite a bit of work on the juniper driver.

mirceaulinic: Sure, that workflow would work well from my perspective.

Just to clarify the question - I think I wasn't clear enough that I was referring to preserving the consistency across the _Junos_ drivers:

> What other getters are pulling from candidate configuration?

- `get_ntp_servers`
- `get_ntp_peers`
- `get_snmp_information`
- `get_probes_config`
- `get_users`
- `get_network_instances`

All of these pull the configuration. From the candidate / pending config, currently. What I was referring to, is that without a patch like this one - for each and every one of those methods - we'd break consistency in this regard within the Junos driver (i.e., with this patch let as-is `get_bgp_config` becomes the only one retrieving from committed, while the other ones I mentioned above would continue to gather from the candidate configuration). Does it make more sense like that?

ckishimo: @mirceaulinic , @ktbyers could you please have a look to #1035  if that suits everyone ?

If ok another option would be to go ahead with these changes now, but making the default to be `candidate` (as today). Then as suggested by @ktbyers just change the default to `committed` in napalm3  it is a change in behavior on Junos. Thanks

mirceaulinic: Thanks for that PR @ckishimo - it's exactly what I had in mind! :-)

ktbyers: @mirceaulinic Okay, that makes sense what you posted above. That would probably want to make me change all those methods, however (to use the running-config and not pending config)

i.e. it is peculiar to me to use `pending configuration` and not actual configuration for all of those. And generally I would not expect that for the rest of the napalm drivers.

mirceaulinic:
Closing this one, as already solved in https://github.com/napalm-automation/napalm/pull/1035.

Issue #1028, opened 2019-08-01
Wrong interface information for Junos virtual-switch w/ bridge domains using get_network_instances 


ferfemar:
### Description of Issue/Question

Output of get_network_instances shows no interfaces in Junos routing instance type virtual-switch when interfaces are assigned under a bridge-domain. It counts them as being part of the master/default instance which is not correct.

### Did you follow the steps from https://github.com/napalm-automation/napalm#faq
(Place an ``x`` between the square brackets where applicable)

- [x ] Yes
- [ ] No

### napalm version
(Paste verbatim output from `pip freeze | grep napalm` between quotes below)

```
ferfecky@lablinux:~/vrf-report (napalm-getters)*$ pip freeze | grep napalm
napalm==2.4.0
```

### Network operating system version
(Paste verbatim output from `show version` - or equivalent - between quotes below)

```
root@vMX1> show version
Hostname: vMX1
Model: vmx
Junos: 14.1R1.10
JUNOS Base OS Software Suite [14.1R1.10]
JUNOS Base OS boot [14.1R1.10]
JUNOS Crypto Software Suite [14.1R1.10]
JUNOS Online Documentation [14.1R1.10]
JUNOS Kernel Software Suite [14.1R1.10]
JUNOS Packet Forwarding Engine Support (M320) [14.1R1.10]
JUNOS Packet Forwarding Engine Support (M/T/EX Common) [14.1R1.10]
JUNOS Routing Software Suite [14.1R1.10]
JUNOS Runtime Software Suite [14.1R1.10]
JUNOS Services AACL PIC package [14.1R1.10]
JUNOS Services Application Level Gateway [14.1R1.10]
JUNOS Services Application Level Gateway (xlp64) [14.1R1.10]
JUNOS Services Application Level Gateway (xlr64) [14.1R1.10]
JUNOS AppId Services PIC Package [14.1R1.10]
JUNOS Services AppId PIC package (xlr64) [14.1R1.10]
JUNOS Border Gateway Function PIC package [14.1R1.10]
JUNOS Services Captive Portal and Content Delivery PIC package [14.1R1.10]
JUNOS Services HTTP Content Management PIC package [14.1R1.10]
JUNOS Services HTTP Content Management PIC package (xlr64) [14.1R1.10]
JUNOS IDP Services PIC Package [14.1R1.10]
JUNOS Packet Forwarding Engine Trio Simulation Package [14.1R1.10]
[...]

PR #1032, opened 2019-08-08
fix junos issue #1028

ckishimo:
Fix for #1028


ktbyers:
@ckishimo Can you explain briefly the fix. I am slightly out of the loop on the behavior of Juniper Views and how adding the pipe character into the YAML file would change the behavior.

Also do we have tests for this use case i.e. can we add a use case where this would fail before doing this fix and would work with the new solution.

We can potentially assist you with getting the tests setup.

Regards, Kirk

ckishimo:
@ktbyers the pipe character is acting as an "OR", as we have different tag names present for different
types of configurations. 

For the existing napalm test case the path `interface/name` is used
```
    <routing-instances>
        <instance>
            <name>DUMMY</name>
            <instance-type>vrf</instance-type>
            <interface>
                <name>lt-0/0/10.1</name>
            </interface>
            ....
        </instance>
    </routing-instances>
```
while now the following configuration is used which needs `bridge-domains/domain/interface/name`
```
    <routing-instances>
        <instance>
            <name>vswitch1</name>
            <bridge-domains>
                <domain>
                    <name>MY-BD</name>
                    <domain-type>bridge</domain-type>
                    <interface>
                        <name>ge-0/0/4.0</name>
                    </interface>
                </domain>
            </bridge-domains>
        </instance>
    </routing-instances>
```
I can try to add a new test based on the configuration provided in #1028 please give me a few days... thanks!

Issue #1041, opened 2019-08-30
lock prevent to load new candidate config


blahu:
### Description of Issue/Question

It's a question rather than an fault description.
We are writing a script that checks if running-config matches the file provided by configuration automation.
To do that we are uploading a new configuration (load_replace_candidate()) and executing compare_config().
Unfortunately sometimes there is already an open session on the box in pending state.
When load_replace_candidate() is executed it runs _lock() function that in case of a pending state will raise an exception SessionLockedException("Session is already in use").

We would like to argue that _lock() should be involved from within commit_config() - this is where it makes more sense to have such check?

Alternatively one could pass a lock/no_lock flag to both or either of the functions

### Did you follow the steps from https://github.com/napalm-automation/napalm#faq
(Place an ``x`` between the square brackets where applicable)

- [x] Yes
- [ ] No

### napalm version
(Paste verbatim output from `pip freeze | grep napalm` between quotes below)

```
napalm==2.4.0
```

### Network operating system version
(Paste verbatim output from `show version` - or equivalent - between quotes below)
I can't do verbatim
```
Arista EOS 4.21.7M
```

### Error Traceback
(Paste the complete traceback of the exception between quotes below)

```

```

ktbyers:
We have a `lock_disable` argument in Juniper. I think it would be fine to add this to EOS (if someone wants to implement the PR on it):

https://github.com/napalm-automation/napalm/blob/develop/napalm/junos/junos.py#L94

blahu:
https://github.com/napalm-automation/napalm/pull/1042

mirceaulinic:
Implemented via #1042.

PR #1042, opened 2019-08-31
Implemented  lock_disable for EOS based on Junos.

blahu:
To fix(?) issue #1041 - https://github.com/napalm-automation/napalm/issues/1041
On branch eos-lock-disable-flag
Changes to be committed:
    modified:   napalm/eos/eos.py

### DIFF
--- a/napalm/eos/eos.py
+++ b/napalm/eos/eos.py
@@ -86,7 +86,26 @@ class EOSDriver(NetworkDriver):
     )
 
     def __init__(self, hostname, username, password, timeout=60, optional_args=None):
-        """Constructor."""
+        """
+        Initialize EOS Driver.
+
+        Optional args:
+            * lock_disable (True/False): force configuration lock to be disabled (for external lock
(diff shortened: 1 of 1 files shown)


mirceaulinic:
Okay, I think I'm going to merge this. If anyone has any objections or remarks, please comment here before we release 2.5.0 when we're going to include this code.

@blahu while this might help providing a way to stay away from the issue (and I think it's right to have this argument, as we already have it in Junos), I seem to recall that I've seen `SessionLockedException("Session is already in use")` myself as well, which seemed to me like a racing condition, but haven't got a chance to dig into it yet. Do you think you'd have the cycles to look into this and try to fix the root cause? The key might be in what you mentioned in the #1041: `Unfortunately sometimes there is already an open session on the box in pending state.` -- why is there a session in pending state? Does it take EOS too long to clean it? Or is there something else? I'd be grateful if you could go down this rabbit hole. Cheers!

Issue #1052, opened 2019-09-18
get_environment not working on IOS-XE using the IOS driver


tcaiazza:
### Description of Issue/Question

When I try to use the `get_environment()` on an ios-xe device of mine I get the following stack trace
```
/home/tim/code/napalm/napalm/ios/ios.py in get_environment(self)
   2098         if "Invalid input detected at" not in output:
   2099             for line in output.splitlines():
-> 2100                 if "Processor" in line:
   2101                     _, _, proc_total_mem, proc_used_mem, _ = line.split()[:5]
   2102                 elif "I/O" in line or "io" in line:

UnboundLocalError: local variable 'proc_total_mem' referenced before assignment
```

The ios driver is looking to run the command `show memory statistics` which isn't valid on an ios-xe device, the command is just `show memory`

*Note*: Please check https://guides.github.com/features/mastering-markdown/
to see how to properly format your request.

### Did you follow the steps from https://github.com/napalm-automation/napalm#faq
(Place an ``x`` between the square brackets where applicable)

- [x] Yes
- [] No

### napalm version
(Paste verbatim output from `pip freeze | grep napalm` between quotes below)

```
napalm==2.4.0
```

### Network operating system version
(Paste verbatim output from `show version` - or equivalent - between quotes below)

```
Cisco IOS Software, IOS-XE Software, Catalyst L3 Switch Software (CAT3K_CAA-UNIVERSALK9-M), Version 03.06.06E RELEASE SOFTWARE (fc1)
Technical Support: http://www.cisco.com/techsupport
Copyright (c) 1986-2016 by Cisco Systems, Inc.
Compiled Sat 17-Dec-16 00:22 by prod_rel_team



Cisco IOS-XE software, Copyright (c) 2005-2015 by cisco Systems, Inc.
All rights reserved.  Certain components of Cisco IOS-XE software are
licensed under the GNU General Public License ("GPL") Version 2.0.  The
software code licensed under GPL Version 2.0 is free software that comes
with ABSOLUTELY NO WARRANTY.  You can redistribute and/or modify such
GPL code under the terms of GPL Version 2.0.
(http://www.gnu.org/licenses/gpl-2.0.html) For more details, see the
documentation or "License Notice" file accompanying the IOS-XE software,
or the applicable URL provided on the flyer accompanying the IOS-XE
software.



ROM: IOS-XE ROMMON
[...]

PR #1053, opened 2019-09-18
Fixes get_environment memory collection on ios-xe devices

tcaiazza:
* fixes issue #1052
* add tests for get_environment on ios-xe devices


mirceaulinic:
Thanks @tcaiazza!

PR #1067, opened 2019-10-09
pin netmiko version to 2.4.2

carlmontanari:
per @ktbyers pinning netmiko to 2.4.2 so that netmiko 3.x doesn't accidentally get installed (netmiko 3.x drops python 2 support) breaking python 2 for next 2.5.0 release!


ktbyers:
Earlier PR beat you to it...closing as duplicate.

PR #1068, opened 2019-10-12
NAPALM with PY3 only support

ktbyers:
Testing against CI-CD


ktbyers:
@mirceaulinic This should be good to go :-)

mirceaulinic (approved):
Fantastic, thanks a lot @ktbyers!

PR #1093, opened 2019-12-05
EOS syntax change updates

bewing:
Arista has released FN 0039 (available to Arista customers with support
contract) detailing changes to command syntax that impacts several show
commands used by NAPALM beginning in EOS Release 4.23.0. NAPALM will
re-try specific show commands if a CommandError is raised with the newer
syntax.


bewing:
Items for discussion:
* Is this the right approach?
* * Should we change the driver to check version on open, and choose command sets appropriately?
* * Should we shim run_commands, and catch CommandErrors for deprecated commands, and feed it a set of regexeps for old-> new replacements?
* How do we test this?
* * Update EOS Mock driver to raise CommandError if specific text is present in mock files?

ktbyers:
Yes, I agree with @mirceaulinic here that it would probably be better if we had this centeralized in the code i.e. some sort of dictionary and a version check.

bewing:
Based on earlier discussions, I believe we wanted to try to avoid blindly changing user input / what is being sent to switches.  Perhaps something like
```python
def _run_commands(self, cmds):
    convert = {
      # old command : new command
      'show user-account': 'show user accounts'
    }
    commands = []
    try:
        return self.device.run_commands(cmds)
    except pyeapi.eapilib.CommandError as e:
        # Check e for bad commands and convert individually
        ...
        return self.device.run_commands(cmds)
```

ktbyers:
Yes, good point. I apologize if I contradicted something I said earlier also...

I was assuming we would only be doing this for hard-coded commands that were used in napalm `getters` and not for config commands nor for `cli` (i.e. arbitrary user commands).

So for `config` commands--I think the issue should be punted onto the end-user (i.e. they need to get their commands correct for their OS-version). Same for `cli` commands (i.e. what you specify is what you get).

So that leaves what do we do for `getters` where we have one set of hard-coded commands that won't work in certain contexts.

bewing:
The other unfortunate part here is that the new commands don't work in older EOS versions -- so we can't blindly use the new commands unless we put a version gate on the EOS driver, and state that older versions aren't supported.  Is that something that we should consider? I don't think so, as the first version that supports the new syntax was released less than a year ago.

tcaiazza:
What about adding something to the `open` function to determine the version number like this below and storing that version number

```           
# let's try to run a very simple command
self.device.run_commands(["show clock"], encoding="text")
version = self.device.run_commands(["show version"])[0]
major, minor, _ = version.get("version").split(".")
self.verion_number = int(major) + int(minor) * .01
```

Then the new `_run_commands` function could be something like this
```
def _run_commands(self, cmds):
    convert = {
      # old command : new command
      'show user-account': 'show user accounts'
    }
    commands = []
    # Convert commands to 4.23+ syntax
    if self.version_number >= 4.23
        cmds = [i if i not in convert.keys() else convert[i] for i in cmds]
    return self.device.run_commands(cmds)
```

mirceaulinic:
Sorry, I dropped the ball here.

> I believe we wanted to try to avoid blindly changing user input / what is being sent to switches.

My comment was probably misleading - let me try to clarify a little: I meant `_run_commands` to be used exclusively by NAPALM internally (of course, without restricting access if the user wants to use it), so the methods would use it out of the box; arbitrary CLI commands via the `cli` method shouldn't be affected. Does this make more sense?

Additionally, I think I like @tcaiazza's suggestion. We could even consider an optional argument if we want to provide a little more flexibility to toggle this feature.

tcaiazza:
Having just learned about distutils.version.LooseVersion, I would change my suggestion to this
```
# let's try to run a very simple command
self.device.run_commands(["show clock"], encoding="text")
version = self.device.run_commands(["show version"])[0]['version']
self.version = distutils.version.LooseVersion(version)
``` 
and 
```
def _run_commands(self, cmds):
    convert = {
      # old command : new command
      'show user-account': 'show user accounts'
    }
    commands = []
    # Convert commands to 4.23+ syntax
    if self.version_number >= distutils.version.LooseVersion("4.23")
        cmds = [i if i not in convert.keys() else convert[i] for i in cmds]
    return self.device.run_commands(cmds)
```

mirceaulinic:
Hey @bewing - have you had a chance to think about the suggestions above? I think your implementation is probably fine either way, just would like to hear your thoughts on a broader use case. Cheers!

bewing:
Sorry for the immense delay on this PR.  Based on feedback, and internal work at $DAY_JOB, I've force-pushed a different changeset done by a co-worker of mine that implements a pyeapi.Node() shim to handle command translation -- take a look and provide some feedback.

(if we hate it, the old code is still in branch fn-0039_old)

napalm/eos/eos.py
```diff
@@ -156,11 +177,15 @@ def open(self):
...
             if self.device is None:
-                self.device = pyeapi.client.Node(connection, enablepwd=self.enablepwd)
+                self.device = Node(connection, enablepwd=self.enablepwd)
             # does not raise an Exception if unusable
 
-            # let's try to run a very simple command
-            self.device.run_commands(["show clock"], encoding="text")
+            # let's try to determine if we need to use new EOS cli syntax
+            cli_version = self._determine_syntax_version(
+                self.device.run_commands(["show version"], encoding="text")[0]["output"]
```
mirceaulinic, line 185: I think the reasoning why we had `self.device.run_commands(["show clock"], encoding="text")` as text, was because `show clock` is only available in that format. That said, I _think_ doing `self.device.run_commands(["show version"])` could simplify the code by getting rid of the regular expression complication and do a simple lookup under the `version` key in the return, i.e.,

```python
sh_ver = self.device.run_commands(["show version"])
cli_version = 2 if EOSVersion(sh_ver[0]['version']) >= EOSVersion("4.23.0") else 1
```

napalm/eos/utils/versions.py
```diff
@@ -0,0 +1,70 @@
+"""Some functions to work with EOS version numbers"""
+import re
+
+
+class EOSVersion:
```
mirceaulinic, line 5: I like this, and I have the feeling we could introduce this sort of classes to any driver (I sort of doubt we can find something generic that works for every platform). That'd be very helpful in many contexts - either NAPALM internal, or users in their environment.

napalm/eos/utils/cli_syntax.py
```diff
@@ -0,0 +1,357 @@
...
+        "vrrp session description": "vrrp description",
+        "vrrp ipv4": "vrrp ip",
+        "vrrp ipv4 secondary": "vrrp ip secondary",
+        "vrrp priority-level": "vrrp priority",
+        "vrrp disabled": "vrrp shutdown",
+        "vrrp advertisement interval": "vrrp timers advertise",
+        "vrrp tracked-object": "vrrp track",
+    },
+    # CLI syntax version after EOS 4.23.0
+    2: {
```
mirceaulinic, line 166: Are all these the mirror of the above? In order to have less stuff to maintain, what if we just flip what's under `1`? Like ``CLI_SYNTAX[2] = {val, key for key, val in CLI_SYNTAX[1].items()}``?

Anyhow, my understanding is that we should only convert, going forward, from the old command to the new command format - and not vice-versa?

napalm/eos/utils/cli_syntax.py
```diff
@@ -0,0 +1,357 @@
...
+
+
+def cli_convert(command, version):
+    """
+    Convert command from CLI version 1 to one from CLI of specified version
+    :param command: str: CLI command
+    :param version: int: EOS CLI version number
+    :return: str: command
+    """
+    return CONVERTER.convert(command, version)
```
mirceaulinic, line 357: This feels a bit redundant. I'd probably just do:

```python
def cli_convert(command, version):
        if version not in CLI_SYNTAX:
            return command

        for c in CLI_SYNTAX[version]:
            if command.startswith(c):
                return CLI_SYNTAX[version][c] + command[len(c) :]

        return command
```
And get rid of the `CliConverter` class.

Not saying to do so, just thinking out loud. :-)

napalm/eos/utils/cli_syntax.py
bewing: I had the same conversation when I first reviewed this, and this was Alexey's feedback:


> Nice idea, however, we have to remove some commands from 2->1 translation dict because of ambiguity. For example, 'description': 'ip msdp description'. It's hard to determine is it related to msdp description or interface description without tracking config mode. And tracking config mode is a bit tough task.

Regarding conversion, I had the following thought:

> Would be nice if the converter supported 2 -> 1 as well, for backwards compatibility with new NAPALM code (someone writes a new getter that uses new commands, will it run on old?)

mirceaulinic: Sounds good

rbcollins123:
@mirceaulinic @bewing So since `load_replace_candidate()` relies upon `send_commands()`, we just had a situation where our rendered configs were being arbitrarily changed back to old syntax now using napalm 3.0.0.  EOS 4.21.4F-4.22.X support old and new formats so we shouldn't arbitrarily swap designed configurations.  It makes sense for things internal to napalm, but if someone is loading a config from Jinja2 template, napalm shouldn't affect the CLI, it should just load it into the config session as-is.  This change will impact people upstream using `napalm-ansible` also if they're on any of those many interim releases depending upon where they are in their configuration migration.  Converting candidate configs magically in the background shouldn't really be the job of napalm when loading configuration files.

mirceaulinic:
Thanks for reporting @rbcollins123.

> just had a situation where our rendered configs were being arbitrarily changed back to old syntax now using napalm 3.0.0

Wondering if we should revisit this discussion: https://github.com/napalm-automation/napalm/pull/1093#discussion_r417872702 (i.e., whether we should offer the flip side, new -> old)?

I'm thinking maybe we'll want to add a new optional argument to toggle if the user wants the conversion in the config or not. Thoughts?

rbcollins123:
> I'm thinking maybe we'll want to add a new optional argument to toggle if the user wants the conversion in the config or not. Thoughts?

I haven't crawled the code yet, but my initial knee-jerk thought when I saw this behavior was to only use the conversion wrapper in an internal `_run_commands` method and change all internal uses to call that and then have the "public" api method keep the old behavior with no auto-conversion;  that way the responsibility for passing valid commands in a call to `run_commands()` falls to the user who would know what they're automating against.

mirceaulinic:
Yep, I understood what you were saying from the first comment @rbcollins123. What I'm suggesting is an optional argument that lets this decision down to the user (i.e., if they want `_run_commands` to transform the config, or not -- which would default to what you were suggesting: won't transform the config, unless explicitly requested). Does this make more sense like that?

rbcollins123:
Yep, makes sense to me

mirceaulinic:
@rbcollins123 would you be able to test my #1212 patch and confirm that'd fix this issue? (@yeled you may be interested in this as well). Feedback welcome from anyone else watching this thread. Thanks!

rbcollins123:
> @rbcollins123 would you be able to test my #1212 patch and confirm that'd fix this issue? (@yeled you may be interested in this as well). Feedback welcome from anyone else watching this thread. Thanks!

FYI I tested and verified that the behavior is now same in v2.4.0, 2.5.0 and 3.0.1 when doing a full config-replace via napalm-ansible.

PR #1106, opened 2020-01-08
Add support for TwoGigabit & TwentyFiveGig interfaces to canonical_map.py

bbmin7b5:
- added support for TwoGigabitEthernet and TwentyFiveGigabitE interfaces found in Cisco 9300 and 9400 series switches.


napalm/base/canonical_map.py
```diff
@@ -99,6 +101,8 @@
     "Serial": "Se",
     "TenGigabitEthernet": "Te",
     "Tunnel": "Tu",
+    "TwoGigabitEthernet": "Tw",
+    "TwentyFiveGigE": "Twe",
```
ktbyers: @bbmin7b5 Wouldn't it be better to be "Two" and "Twe" as "Tw" could be ambiguous with both "TwoGigabitEthernet" and "TwentyFiveGigE"?

napalm/base/canonical_map.py
```diff
@@ -72,6 +72,8 @@
     "Tunnel": "Tunnel",
     "Tun": "Tunnel",
     "Tu": "Tunnel",
+    "Tw": "TwoGigabitEthernet",
+    "Twe": "TwentyFiveGigE",
```
ktbyers: @bbmin7b5 Wouldn't it be better to be "Two" and "Twe" as "Tw" could be ambiguous with both "TwoGigabitEthernet" and "TwentyFiveGigE"?

napalm/base/canonical_map.py
bbmin7b5: @ktbyers I went with "Tw" for the "TwoGigabitEthernet" as that is how Cisco displayed the shortened form of the interface when using "show interface status" and "show ip interface brief". I figured I'd follow that model as that is what the other key/value pairs seemed to conform to.

I hope my reasoning makes sense.

ktbyers:
Made minor update here as I was pretty convinced the reverse map should not have any collisions in it (i.e. should be unique as far as abbreviated names):

https://github.com/napalm-automation/napalm/pull/1107

bbmin7b5:
Thanks @ktbyers

PR #1108, opened 2020-01-13
Prevent nxos.get_interfaces_ip from failing if no ipv6 support

tcaiazza:
Cisco Nexus 3548's still don't support ipv6, so catch the NXAPIPostError
exception when `get_interfaces_ip` tries to run `show ipv6 interfaces`
and return the ipv4 addresses


tcaiazza:
> Thanks for this patch @tcaiazza. Would you please add a test case for this?

@mirceaulinic Test case added.

PR #1121, opened 2020-02-07
Allow username detection on older Cisco ios versions

dlundgren:
This switches to the older `| include` syntax, and allows usernames with `password` to be detected.


dlundgren:
It looks like while this will work for older ios versions, as I have available, it may not work properly for newer versions. I'm evaluating if it's plausible to support detection of section vs include.

ktbyers:
@dlundgren Can you expand a bit on what works on older IOS versions and what doesn't work on newer IOS versions (as it is a bit unclear to me the reference of `this` here). In other words, a bit unclear to me if you are talking about the original problem or the fix.

dlundgren:
Older IOS does not support `section`, but it does support `include`. 

For example, I have a 3750G with  IOS 12.2(44)SE12 and using `| include username` allows napalm to detect the users.

On newer IOS with support for ssh keys, `include` fails to load the `key-hash` for a user, but `section` does include the `key-hash`.

ktbyers:
Yeah, I guess you could do a PR and try the newer form first i.e. `| section` and  then if `Invalid command` comes back try the `| include` solution.

mirceaulinic:
Hey @dlundgren - have you seen Kirk's suggestion?

dlundgren:
I haven't had time to fix this up with his suggestions. I will try to look at this over the next week.

dlundgren:
I've updated to check the command output by looking for `Invalid input detected` in the output of the command using `| section username`, then falling back to trying `| include username`

PR #1132, opened 2020-02-20
No silent pass of errors in base.helpers & iOS-XR - logging

gnubyte:
This is mainly an enrichment of logging in the ios-xr module and base.helpers. The end result is that when logging is enabled, the logging handlers can report back to the ERROR level of logging that an error was encountered when attempting to parse XML responses from **any** of the drivers.

More to come after trying to chase down a timeout issue when parsing large XML responses via netmiko but the addition of logging has allowed for visibility into the bug.


gnubyte:
I'd also like to add that I have added some formalities to logging...


top level debug handler for the library starts at `napalm`
modules following it will use `napalm.<dir>.moduleName` where module is the file

This is in an attempt to create parent and child logging handlers that follow the directory and module structure.

ktbyers:
@gnubyte You are failing the linter/black. You need to run Python `black` on the code.

We are currently using this version of Black:

https://github.com/napalm-automation/napalm/blob/develop/requirements-dev.txt#L1

gnubyte:
I'll take a look and rework - thanks.

gnubyte:
Trimmed back the scope of this PR to just enriching logging. Going to have to open a separate PR and branch to deal with enriching the data being brought back as it breaks multiple drivers' code testing...


The problem I encountered that caused me to look at this in the first place was pyiosxr driver. The XML being sent is receiving a paginated XML response back from the device, which raised an IteratorID error found at the top of pyiosxr driver....However because iosxr driver was passing the error silently the error would not be exposed.

It is now actively logged to the error channel under `napalm.pyiosxr` logging handle off the root napalm logging handler, while also raising it.

napalm/base/helpers.py
```diff
@@ -19,6 +20,11 @@
 from napalm.base.utils.jinja_filters import CustomJinjaFilters
 from napalm.base.canonical_map import base_interfaces, reverse_mapping
 
+# ----------------------------------------------------------------------------------------------------------------------
```
mirceaulinic: I'm surprised pylama didn't complain about this long line :thinking:

napalm/base/helpers.py
```diff
@@ -155,12 +161,19 @@ def regex_find_txt(pattern, text, default=""):
...
+            logger.error("No Regex match found for pattern: %s" % (str(pattern)))
+            raise Exception("No Regex match found for pattern: %s" % (str(pattern)))
         if not isinstance(value, type(default)):
             if isinstance(value, list) and len(value) == 1:
                 value = value[0]
             value = type(default)(value)
-    except Exception:  # in case of any exception, returns default
+    except Exception as regexFindTxtErr01:  # in case of any exception, returns default
+        logger.error(
+            'errorCode="regexFindTxtErr01" in napalm.base.helpers with systemMessage="%s"\
```
mirceaulinic: I think this is a bit superfluous, as `logger.error(str(regexFindTxtErr01), exc_info=True` should be sufficient - a good logger handler should point you to the right file and everything else you need, don't have to code that into the error message. Let's keep it simple :-)

napalm/base/helpers.py
```diff
@@ -232,13 +257,23 @@ def find_txt(xml_tree, path, default=""):
...
+        if xpath_length and xpath_applied[0] is not None:
             xpath_result = xpath_applied[0]
             if isinstance(xpath_result, type(xml_tree)):
                 value = xpath_result.text.strip()
             else:
                 value = xpath_result
-    except Exception:  # in case of any exception, returns default
+        else:
+            if xpath_applied == "":
+                xpath_applied = ""
```
mirceaulinic: Did you intend to have something different here?

napalm/pyIOSXR/iosxr.py
```diff
@@ -46,6 +48,8 @@
 from napalm.pyIOSXR.exceptions import InvalidInputError
 from napalm.pyIOSXR.exceptions import InvalidXMLResponse
 
+logger = logging.getLogger("napalm.pyiosxr")
```
mirceaulinic: `logger = logging.getLogger(__name__)` is usually more flexible. Would be fine to take this in as-is, but I think you'll have a couple more changes, so might as well include this one too. :wink:

napalm/base/helpers.py
```diff
@@ -155,12 +161,19 @@ def regex_find_txt(pattern, text, default=""):
     value = re.findall(pattern, text)
     try:
         if not value:
-            raise Exception
+            logger.error("No Regex match found for pattern: %s" % (str(pattern)))
```
mirceaulinic: This feels a bit redundant.

mirceaulinic:
FYI: we are planning to release NAPALM 3.0.0 soon. If you'd like your patch to be included in this version, please address the comments above. Thanks!

napalm/base/helpers.py
gnubyte: Syntax may be ignored on comments but we can trim it down to PEP8's 79 char line limit

napalm/base/helpers.py
gnubyte: So it is superfluous at face value, but it also gives a lot of intrinsic information to any LEM Log and event manager that a consumer may be using to help that log and event manager automatically extract context from the log/runtime. The `<field>="value"` allows most LEMs to both recognize the data and tag it.

For example:
 -  in ELK you can use a [plugin filter](https://www.elastic.co/guide/en/logstash/current/plugins-filters-kv.html) to auto extract key words from the text 
- in Splunk you the [default key value](https://docs.splunk.com/Documentation/Splunk/8.0.2/Knowledge/Automatickey-valuefieldextractionsatsearch-time) mode is as described `key=value` 

The fields added here, error code: for reference in the library , the systemMessage: from the runtime, and the message: a human readable dictation of the raised error/event - all contribute value adds in a separate context to help enrich or tell that story to both the consumer at REPL then again at a LEM.

From a different perspective, if you are not living in a LEM I can see this being superfluous so I wanted to add some context. I'll defer to your judgement when I hear back from you as to whether this belongs here still. Thank you again for taking the time to review this 🙌👍.

napalm/base/helpers.py
gnubyte: Nice catch. Took me a minute to get what you were driving at.

I originally had a if `xpath_applied == None or xpath_applied== ""` but I think python black didn't like that when the code was being checked. I wanted to do conditionals to express that the variable could be an empty string  or just not be set at all. Next push will feature just the debug message when xpath is applied.

gnubyte:
Added enriched logging around the base iosxr file today where previous iterations were mostly limited to pyiosxr subdriver & base/helper.

Thank you again for taking the time to review this PR!

napalm/base/helpers.py
mirceaulinic: Thanks for this explanation. Yes, I understand your reasoning, though I have mixed feeling about scoping that to a specific application. At the same time, I like having as many details as possible, so I'd say let's go ahead with this! :wink:

napalm/iosxr/iosxr.py
```diff
@@ -2085,6 +2098,15 @@ def traceroute(
         try:
             ipv = IPAddress(destination).version
         except AddrFormatError:
+            try:
+                destination
```
mirceaulinic: What are you trying to test here?

gnubyte: If a user passes destination=None or an implementation populates the variable, I can safely attempt to populate the passed destination parameter then include it in the logging message

napalm/base/helpers.py
gnubyte: This was resolved  in commit [
2130fcc](https://github.com/napalm-automation/napalm/pull/1132/commits/2130fcc1c245c28b423b481a8d019dc5e99395d2) so we should be OK now

napalm/iosxr/iosxr.py
mirceaulinic: Wouldn't just `if destination is None` work? I've never seen this approach with an empty try-except block...

gnubyte: Unfortunately python black doesn't like those so I had to reinvent the wheel here lol.

mirceaulinic: Sorry, that won't do it, please just remove this altogether... that `logger.error` in the form you have now should be fine without this try block.

napalm/base/helpers.py
gnubyte: I hear ya that it feels redundant but the main reason I needed to open the PR was that depending on the driver implementation, these kinds of errors do get passed over when the get raised. For example if a driver implementation passes on a try/except, this will catch the condition globally. While I'd prefer if those kinds of situations were handled at and audited when a driver PR is made & committed, this is a "rather safe than sorry" implementation ;-)

napalm/iosxr/iosxr.py
gnubyte: Hey @mirceaulinic new commit is up with this removed so we should be good.

Again thank you for going through this PR!

PR #1138, opened 2020-03-08
patch privilege behavior of ios driver(eliminate need for priv exec for user exec functions)

remingu:
By default driver('ios').open() will fail if the user does not provide an enable secret or has already privilevel level 15.
Many getter functions inside napalm do not necessarily require priv exec mode and can be already performed in user-exec mode. It is non unsual to have users assigned to limited privilege levels for certain tasks.

The is caused due to invocation of self._netmiko_device.enable() inside 
| napalm:base:networkdriver:_netmiko_open()
This will trigger netmikos enable(). This is perfectly fine when called from netmiko specifically whenever needed.
Without access to priv exec (priv 15) the ios-driver will fail completely but for many 
purposes it would work just fine without priv exec mode(15).

Also enable mode is  device-specific, therefore it should be probably invoked from inside the according driver and not generically from the basedriver,


ktbyers:
enable() is required for many napalm IOS operations. This is a breaking change which I don't think we should do.

You should probably open a GitHub issue and discuss there.

Issue #1145, opened 2020-03-11
Broken get_vlan on NX-OS for VLANs with no interfaces assigned


ggiesen:
### Description of Issue/Question

Using get_vlans for nxos_ssh fails when there are VLANs defined with no interfaces assigned.

Example output captured from vlan_table_raw:
```
[{'vlanshowbr-shutstate': 'noshutdown',
  'vlanshowbr-vlanid': '1',
  'vlanshowbr-vlanid-utf': '1',
  'vlanshowbr-vlanname': 'default',
  'vlanshowbr-vlanstate': 'active',
  'vlanshowplist-ifidx': 'port-channel11,port-channel12,Ethernet1/43,Ethernet1/44,Ethernet1/47,Ethernet1/48'},
 {'vlanshowbr-shutstate': 'noshutdown',
  'vlanshowbr-vlanid': '1530',
  'vlanshowbr-vlanid-utf': '1530',
  'vlanshowbr-vlanname': 'VLAN1530',
  'vlanshowbr-vlanstate': 'active',
  'vlanshowplist-ifidx': 'port-channel11,port-channel12,port-channel100,Ethernet1/1,Ethernet1/2,Ethernet1/3,Ethernet1/4,Ethernet1/5,Ethernet1/6,Ethernet1/7,Ethernet1/8,Ethernet1/43,Ethernet1/44,Ethernet1/47,Ethernet1/48,Ethernet1/53,Ethernet1/54'},
 {'vlanshowbr-shutstate': 'noshutdown',
  'vlanshowbr-vlanid': '1671',
  'vlanshowbr-vlanid-utf': '1671',
  'vlanshowbr-vlanname': 'VLAN1671',
  'vlanshowbr-vlanstate': 'active'},
 {'vlanshowbr-shutstate': 'noshutdown',
  'vlanshowbr-vlanid': '3043',
  'vlanshowbr-vlanid-utf': '3043',
  'vlanshowbr-vlanname': 'VLAN3043',
  'vlanshowbr-vlanstate': 'active',
  'vlanshowplist-ifidx': 'port-channel11,port-channel12,port-channel100,Ethernet1/1,Ethernet1/2,Ethernet1/3,Ethernet1/4,Ethernet1/5,Ethernet1/6,Ethernet1/7,Ethernet1/8,Ethernet1/43,Ethernet1/44,Ethernet1/47,Ethernet1/48,Ethernet1/53,Ethernet1/54'},
[...]

PR #1150, opened 2020-03-13
Fix for napalm-automation/napalm#1145 = get_vlans() broken on NX-OS where no interfaces in VLAN

ggiesen:
Fix for napalm-automation/napalm#1145 = get_vlans() broken on NX-OS where no interfaces in VLAN


test/nxos/mocked_data/test_get_vlans/6000/expected_result.json
```diff
@@ -1,44 +1,7 @@
 {
   "230": {
     "name": "database",
-    "interfaces": [
```
mirceaulinic: Could you please clarify why you're removing these?

I'm not familiar with this `vlanshowplist-ifidx`, from your description from #1145 I understand that `Using get_vlans for nxos_ssh fails when there are VLANs defined with no interfaces assigned.`, however in the mock data associated for this test case I understand that there are: https://github.com/napalm-automation/napalm/blob/8f986aaf4b2c1f7b8cbc5b27ac77ed78d7fdccbd/test/nxos/mocked_data/test_get_vlans/6000/show_vlan_brief.json#L17-L19... what's the secret key here? :smile:

ggiesen: vlanshowplist-ifidx is the list of Interfaces associated to the VLAN. If there are no interfaces associated to the VLAN, vlanshowplist-ifidx doesn't exist (which creates the bug). 

In order to create a test case for this, I just modified one VLAN from the existing test data to remove vlanshowplist-ifidx. You're right ithat I could have added a VLAN instead, let me know if that is your preference.

mirceaulinic: I see, yes, please add a separate test case, without adjusting the existing ones.

mirceaulinic:
Bump @ggiesen

ggiesen:
Added new test cases, however it looks like pylama is failing on code unrelated to my changes

test/nxos/mocked_data/test_get_vlans/interface_without_vlans/show_vlan_brief.json
```diff
@@ -0,0 +1,261 @@
+{
```
mirceaulinic, line 1: It looks like every VLAN in this output has the `vlanshowplist-ifidx`?

test/nxos/mocked_data/test_get_vlans/interface_without_vlans/show_vlan_brief.json
```diff
@@ -0,0 +1,261 @@
...
+        "vlanshowbr-shutstate": "noshutdown",
+        "vlanshowplist-ifidx": "port-channel12,port-channel100,Ethernet1/21,Ethernet1/22,Ethernet1/23,Ethernet1/24,Ethernet1/47,Ethernet1/48"
+      },
+      {
+        "vlanshowbr-vlanid": "1115",
+        "vlanshowbr-vlanid-utf": "1115",
+        "vlanshowbr-vlanname": "Vlan1115",
+        "vlanshowbr-vlanstate": "active",
+        "vlanshowbr-shutstate": "noshutdown"
+      },
```
ggiesen, line 226: VLAN 1115 does not have vlanshowplist-ifidx. I can trim out some of the extraneous VLANs/add a few more VLANs without interfaces to make the test case more apparent if that would be better

mirceaulinic: Na, that's fine, I missed it. Thanks for clarifying!

PR #1176, opened 2020-04-15
Add authentication field to junos_bgp_config_view

mngan:
The authentication key can be defined at the group level, not just at
the neighbor level


napalm/junos/junos.py
```diff
@@ -1068,7 +1069,6 @@ def build_prefix_limit(**args):
         }
 
         _PEER_FIELDS_DATATYPE_MAP_ = {
-            "authentication_key": str,
```
mirceaulinic, line 1071: With this change, this will break the functionality when the key is configured at the neighbor level. So it may fix your use case, but it'll break others'. So please don't remove this line.

The solution would be to pull both (as you did by adding `authentication_key: {authentication-key: unicode}` into the `napalm/junos/utils/junos_views.yml` file), then somewhere under [this](https://github.com/napalm-automation/napalm/blob/develop/napalm/junos/junos.py#L1176-L1205) block have a check like:

```python
if key == 'authentication_key' and not bgp_peer_details['authentication_key']:
    bgp_peer_details['authentication_key'] = value
```

Something along these lines.

This would ensure that the auth key is inherited from the group when not configured at the neighbor level.

mngan: Deleting that field doesn't because all the field in _COMMON_FIELDS_DATATYPE_ are added to _PEER_FIELDS_DATATYPE_MAP

```
        _PEER_FIELDS_DATATYPE_MAP_ = {
            "route_reflector_client": bool,
            "nhs": bool,
        }
        _PEER_FIELDS_DATATYPE_MAP_.update(_COMMON_FIELDS_DATATYPE_)
```

mngan: I am not sure what the additional lines you are suggesting will do. From my tests the authentication key turns up in the neighbor when configured in the neighbor and in the group when it is configured in the group. I don't see any existing inheritance behavior (e.g policies are not inherited from the group). Should this not be reflecting the actual configuration rather than the effective configuration? I looked at the code, and I do not see anything that would inherit values from the group into the neighbor. 

```
group as9050-telekom-romania {
    type external;
    import [ normalize-bgp-routes routes-from-as9050 ];
    authentication-key "SECRET-DATA";
    export [ routes-to-as9050 routes-to-peering ];
    peer-as 9050;
    neighbor 193.231.106.145 {
        family inet {
            unicast {
                prefix-limit {
                    maximum 5000;
                }
            }
        }
    }
}
group as6830-liberty-global {
    type external;
    import [ normalize-bgp-routes routes-from-as6830 ];
    export [ routes-to-as6830 routes-to-peering ];
    peer-as 6830;
    neighbor 213.46.177.45 {
        authentication-key "SECRET-DATA";
    }
    neighbor 2001:730:2d04:6::d52e:b361 {
        authentication-key "SECRET-DATA";
    }
}
```

```
    "as9050-telekom-romania": {
        "type": "external",
        "apply_groups": [],
        "remove_private_as": false,
        "multipath": false,
        "multihop_ttl": 64,
        "description": "",
        "local_address": "",
        "local_as": 0,
        "remote_as": 9050,
        "import_policy": "normalize-bgp-routes routes-from-as9050",
        "export_policy": "routes-to-as9050 routes-to-peering",
        "authentication_key": "SECRET-DATA",
        "cluster": null,
        "prefix_limit": {},
        "neighbors": {
            "193.231.106.145": {
                "route_reflector_client": true,
                "nhs": false,
                "description": "",
                "local_address": "",
                "local_as": 0,
                "remote_as": 0,
                "import_policy": "",
                "export_policy": "",
                "authentication_key": "",
                "inet_unicast_limit_prefix_limit": 5000,
                "prefix_limit": {
                    "inet": {
                        "unicast": {
                            "limit": 5000
                        }
                    }
                }
            }
        }
    },
    "as6830-liberty-global": {
        "type": "external",
        "apply_groups": [],
        "remove_private_as": false,
        "multipath": false,
        "multihop_ttl": 64,
        "description": "",
        "local_address": "",
        "local_as": 0,
        "remote_as": 6830,
        "import_policy": "normalize-bgp-routes routes-from-as6830",
        "export_policy": "routes-to-as6830 routes-to-peering",
        "authentication_key": "",
        "cluster": null,
        "prefix_limit": {},
        "neighbors": {
            "213.46.177.45": {
                "route_reflector_client": true,
                "nhs": false,
                "description": "",
                "local_address": "",
                "local_as": 0,
                "remote_as": 0,
                "import_policy": "",
                "export_policy": "",
                "authentication_key": "<SECRET-DATA>",
                "prefix_limit": {}
            },
            "2001:730:2d04:6::d52e:b361": {
                "route_reflector_client": true,
                "nhs": false,
                "description": "",
                "local_address": "",
                "local_as": 0,
                "remote_as": 0,
                "import_policy": "",
                "export_policy": "",
                "authentication_key": "<SECRET-DATA>",
                "prefix_limit": {}
            }
        }
    },
```

I did notice that there was some asymmetry in the logic between  [here](https://github.com/napalm-automation/napalm/blob/7b704b59303a64b87cb30bc2dc83aa65780c1a4b/napalm/junos/junos.py#L1177-L1178) and [here](https://github.com/napalm-automation/napalm/blob/7b704b59303a64b87cb30bc2dc83aa65780c1a4b/napalm/junos/junos.py#L1211). 

Also, I noticed the route_reflector_client [test](https://github.com/napalm-automation/napalm/blob/7b704b59303a64b87cb30bc2dc83aa65780c1a4b/napalm/junos/junos.py#L1232) is wrong. It only tests for the presence of the key, not the value of the key. So the test is always true.

mirceaulinic: > From my tests the authentication key turns up in the neighbor when configured in the neighbor and in the group when it is configured in the group.

Yes, which is not what we want: check out the structure from https://napalm.readthedocs.io/en/latest/base.html#napalm.base.base.NetworkDriver.get_bgp_config. The `authentication_key` is only at the neighbor level. If we want to make it so it appears at the group level, this will be an API change, so you'll need to submit a PR to make the appropriate changes across _all_ the core drivers implementing `get_bgp_config` (i.e., `junos`, `iosxr`, `eos`, and `ios`). Otherwise, please check again my comment; with my suggestion, you should see the following:

```
    "as9050-telekom-romania": {
        "type": "external",
        "apply_groups": [],
        "remove_private_as": false,
        "multipath": false,
        "multihop_ttl": 64,
        "description": "",
        "local_address": "",
        "local_as": 0,
        "remote_as": 9050,
        "import_policy": "normalize-bgp-routes routes-from-as9050",
        "export_policy": "routes-to-as9050 routes-to-peering",
        "cluster": null,
        "prefix_limit": {},
        "neighbors": {
            "193.231.106.145": {
                "route_reflector_client": true,
                "nhs": false,
                "description": "",
                "local_address": "",
                "local_as": 0,
                "remote_as": 0,
                "import_policy": "",
                "export_policy": "",
                "authentication_key": "SECRET-DATA",
                "inet_unicast_limit_prefix_limit": 5000,
                "prefix_limit": {
                    "inet": {
                        "unicast": {
                            "limit": 5000
                        }
                    }
                }
            }
        }
    },
```

I agree however that it wouldn't reflect the BGP config correctly.

> Also, I noticed the route_reflector_client test is wrong. It only tests for the presence of the key, not the value of the key. So the test is always true.

I'm not following why you're thinking that's wrong? What that code says is: if the `cluster` key is there, let's remove it. :man_shrugging:

PR #1177, opened 2020-04-16
add a few less used canonical mappings

itdependsnetworks:
I had this tied to a PR that never got merged, just going back and cleaning up and adding back.


ktbyers:
@itdependsnetworks CI-CD is failing. It looks like one test needs updated in the test data such that it uses the canonical reults.

```
E           AssertionError: Expected result varies on some keys [{"interface": {"result": "Ethernet1/2", "expected": "eth1/2"}}, {"interface": {"result": "Ethernet1/1", "expected": "eth1/1"}}]
```

napalm/nxos/nxos.py
```diff
@@ -632,9 +632,13 @@ def _parse_vlan_ports(self, vlan_s):
...
                 for i in range(int(find[0][1]), int(find[0][2]) + 1):
-                    vlans.append(find[0][0] + str(i))
+                    vlans.append(
+                        napalm.base.helpers.canonical_interface_name(
+                            find[0][0] + str(i)
+                        )
+                    )
             else:
-                vlans.append(vls.strip())
+                vlans.append(napalm.base.helpers.canonical_interface_name(vls.strip()))
```
mirceaulinic, line 641: I like this, just wanted to mention that on IOS at least there's an optional argument `canonical_int` defaulting to `False` currently - is this something we'd want on NX-OS as well? My preference would probably be to always use the canonical map, but maybe there are some reasons on the existence of `canonical_int` arg I'm not aware of. CC @ktbyers

itdependsnetworks: I do not see this anywhere else in the nxos/nxos_ssh code. I can add it if desired.

ktbyers: Let's open that as a separate issue (so this PR doesn't get bogged down). I would guess for backwards compatibility, but it doesn't look like it was fully/consistently implemented.

I would vote always use canonical interface and just get rid of the argument to toggle that off (especially given that it looks like it is currently inconsistently implemented).

PR #1185, opened 2020-04-28
Fix get_bgp_neighbors_detail IOS-XE

petrmarciniak:
The get_bgp_neighbors_detail function for IOS-XE crashes as it fails to parse the `sh ip bgp all sum` output correctly when:

* neighbor information is spread across two lines (happens often when viewing IPv6 neighbors)
* established, idle, etc. time of the session is in hh:mm:ss format

E.g. the following line would fail to be parsed:
```
2001:559::1
                4         7922 6944479  138054 116434201    0    0 12:22:20       45646
```
As a fix I've updated the corresponding TextFSM templates (as well as the test input and correct result).


test/ios/mocked_data/test_get_bgp_neighbors_detail/normal/show_ip_bgp_all_sum.txt
```diff
@@ -34,7 +34,8 @@ BGP using 39527928 total bytes of memory
 BGP activity 3436532/2710503 prefixes, 19215241/17575537 paths, scan interval 60 secs
 
 Neighbor        V           AS MsgRcvd MsgSent   TblVer  InQ OutQ Up/Down  State/PfxRcd
-2001:559::1     4         7922 6944479  138054 116434201    0    0 12w2d       45646
```
mirceaulinic, line 37: To ensure your changes don't break the existing behaviour, please don't change this file and provide a separate test case with the output you're seeing.

test/ios/mocked_data/test_get_bgp_neighbors_detail/normal/show_ip_bgp_ipv6_unicast_neigh_2001_559__1.txt
```diff
@@ -3,7 +3,7 @@ BGP neighbor is 2001:559::1,  remote AS 7922, external link
  Description: Comcast IPv6 Transit
  Inherits from template PEER-SESSION-AS7922 for session parameters
   BGP version 4, remote router ID 68.86.1.1
-  BGP state = Established, up for 12w2d
+  BGP state = Established, up for 12:22:20
```
mirceaulinic, line 6: Same comment as https://github.com/napalm-automation/napalm/pull/1185/files#r416569119

petrmarciniak: All understood, I've reverted the the file to the previous version

test/ios/mocked_data/test_get_bgp_neighbors_detail/normal/show_ip_bgp_all_sum.txt
petrmarciniak: The `show_ip_bgp_all_sum.txt` is the base for the whole test. Following your suggestion I've added a new neighbor to the list and kept all the original neighbors in place. Please let me know if this makes sense to you?

mirceaulinic: We usually prefer to avoid changing this file when fixing a bug (as probably you won't see both outputs, i.e., in both `12:22:20` and `12w2d` format at the same time -- but never say never Cisco is Cisco lol), and have a separate test case; that means, another set of files under a directory, e.g., ` test/ios/mocked_data/test_get_bgp_neighbors_detail/<some test case name>/`.
But this change is small enough we can leave this pass...

petrmarciniak: All right, it all makes sense now. I wasn't quite sure how the test framework works. Thank you for the explanation, I'll keep it in mind for the next time!

PR #1188, opened 2020-04-29
Fixed pyIOSXR XML mode timeout due to Netmiko cmd_verify

melewitz:
Disabled Netmiko cmd_verify globally in pyIOSXR and now require at least netmiko 3.1.0.

Netmiko now defaults to cmd_verify=True in send_command.  This verifies that the sent text is echoed from the remote.  Devices in XML mode do not echo; therefore it timed out waiting.

Users can still enable global_cmd_verify by passing the option to the constructor or open method.

### DIFF
--- a/napalm/pyIOSXR/iosxr.py
+++ b/napalm/pyIOSXR/iosxr.py
@@ -186,6 +186,7 @@ class IOSXR(object):
                 port=self.port,
                 username=self.username,
                 password=self.password,
+                global_cmd_verify=False,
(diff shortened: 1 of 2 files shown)


ktbyers:
Did you sanity check this works against an IOS-XR device and that it fixes the previous issue you ran into?

melewitz:
Yes, i pulled the latest versions, verified the problem, put the fix in and
verified the resolution.

PR #1193, opened 2020-05-01
IOS-XR: remove unnecessary lines from get_config output

mirceaulinic:


ktbyers:
@mirceaulinic Do we want to get rid of this one also?

```
!! Last configuration change at Wed Aug 17 16:49:00 2016 by vagrant
```

I see it in the test data.

mirceaulinic:
On Fri, 1 May 2020, 17:57 Kirk Byers, <notifications@github.com> wrote:

> @mirceaulinic <https://github.com/mirceaulinic> Do we want to get rid of
> this one also?
>
> !! Last configuration change at Wed Aug 17 16:49:00 2016 by vagrant
>
> I see it in the test data.
>

Yes, I think that is fine to stay - it's commented so it should be ignored
when loaded, while in a backup system that line only changes when there are
other changes (so other lines are likely going to be changed as well) - and
in fact it gives information on who changed the config.

PR #1212, opened 2020-05-13
Add eos_fn0039_config optional arg to toggle FN 0039 on config commands

mirceaulinic:


mirceaulinic:
Travis build: https://travis-ci.org/github/napalm-automation/napalm/builds/686533583

ktbyers:
The argument here seems very obscure `fn0039`.

Is there a more meaningful argument we could use? I would probably call it something like `convert_newstyle_config` and then reference `fn0039` in the docstring.

mirceaulinic:
No particular preference from me, just wanted to have it reference the reasoning of this change, I find "new style" very vague... if Arista will change again, we might end up with new-new-style, who knows :rofl: 
Happy to change if there are any strong opinions on this.

ktbyers:
I don't care about it enough one way or another and it too easily can be bikeshedding (and as you said new_style is very vague).

We'll let other people chime in and now that I have seen `eos_fn0039_config` it is growing on me.

napalm/eos/pyeapi_syntax_wrapper.py
```diff
@@ -32,9 +32,11 @@ def run_commands(self, commands, **kwargs):
         :param kwargs: other args
         :return: list of outputs
         """
-        if isinstance(commands, str):
-            new_commands = [cli_convert(commands, self.cli_version)]
-        else:
-            new_commands = [cli_convert(cmd, self.cli_version) for cmd in commands]
+        fn0039_transform = kwargs.pop("fn0039_transform", True)
```
ktbyers, line 35: Shouldn't this default to False if `fn0039_transform` doesn't exist?

ktbyers:
I am fine with the argument name. I had one other question during the review and besides that I thought everything looked good.

napalm/eos/pyeapi_syntax_wrapper.py
mirceaulinic: Oh, I thought I'd default this to True as we may want to disable this behaviour only for the config-related methods (the getters would always convert the commands under the hood)... so it was mostly a matter of whether we update every getter (everywhere `run_commands` is called) or just `_load_config`. Not sure if I'm explaining this well?

ktbyers: Okay, that makes sense.

So internal to NAPALM for show commands, we would do this automatic conversion. But then for any configuration changes by end users, they would default to `no conversion` (with an option they could specify that would do the conversion).

Is that the correct understanding?

mirceaulinic: Yes, correct, much better worded like this. :smile:

PR #1214, opened 2020-05-13
Fix cmd_verify and null character on NX-OS

ktbyers:


ktbyers:
@xschlef Can you test this fix?

xschlef:
you have a typo in line 509. command_verify should be cmd_verify.
If I change it manually then everything seems to work ok with paramiko 3.1

ktbyers:
@xschlef Ooops, that was dumb...let me fix. Thanks for pointing that out.

ktbyers:
@mirceaulinic You good with this change?

PR #1216, opened 2020-05-14
Raise default max_loops in send_command

ExaneServerTeam:
Raise default max_loops from 500 to 1000 on
nxos_ssh send_command.

We own some very heavy loaded Nexus 5696 Chassis
with ~40 Extensions (HP Enclosure C7000 Enclosures).
Listing the interfaces (get_interfaces) takes quite
a while, and uping the max_loops solve the timeout issue.

### DIFF
--- a/napalm/nxos_ssh/nxos_ssh.py
+++ b/napalm/nxos_ssh/nxos_ssh.py
@@ -444,7 +444,7 @@ class NXOSSSHDriver(NXOSDriverBase):
 
         raw_text argument is not used and is for code sharing with NX-API.
         """
-        return self.device.send_command(command)
+        return self.device.send_command(command, max_loops=1000)
 
     def _send_command_list(self, commands, expect_string=None):
         """Wrapper for Netmiko's send_command method (for list of commands."""


ktbyers:
@ExaneNetworkTeam Does just setting global_delay_factor to 2 in optional_args fix your issue?

The default should allow for ~100 seconds? Your interfaces table doesn't get returned in 100 seconds?

ExaneServerTeam:
@ktbyers Well it's confirmed. running show interface on those switches takes ~3m 10s.
I'll try the global_delay_factor set to 2 and let you know.

ktbyers:
Okay, so just to verify that is a single `show interface` command that is taking ~3m 10s to complete?

It is what it is, but just want to make sure I have accurate data on it.

ExaneServerTeam:
@ktbyers The command *show interface* was run over ssh to perform the measurement.

ExaneServerTeam:
@ktbyers Switching over from *max_loops* to *delay_factor = 2* also works.
Don't know which option we should pick.

ktbyers:
If you can fix it with global_delay_factor, than that is the better fix as it doesn't require any code changes to napalm library i.e. you can do it by adding the following argument to your NAPALM open() call:

```
optional_args = {"global_delay_factor": 2}
```

ExaneServerTeam:
@ktbyers  Setting as requested the following on the driver call does the job ! Thx. I will retire this pull request.
```
optional_args = {"global_delay_factor": 2}
```

PR #1240, opened 2020-06-22
fix: force_no_enable enabled on ios and nx only

tynany:
As per NAPALM documentation, `force_no_enable` is only used for Cisco NXOS and IOS devices, but is currently implemented for all NAPALM drivers that use `_netmiko_open`, which breaks those drivers as it is unlikely the `force_no_enable` method has been implemented. 

This MR only enables `force_no_enable` for the Cisco devices types (`cisco_ios_telnet`, `cisco_ios` and `cisco_nxos`).

Below is an example error output of a NAPALM driver that uses `_netmiko_open` but does not implement the `force_no_enable` method:

```
'$insertNAPALMDriveName' object has no attribute 'force_no_enable'
```


ktbyers:
Hmmm, how about we do a try/except and gracefully handle the AttributeError exception instead?

I am pretty strongly against doing a platform conditional check.

tynany:
Good call -- try/except is a much better idea. I have pushed a new commit

mirceaulinic:
Bump @tynany

tynany:
Apologies for the delay in getting back to you. Travis is now happy -- https://travis-ci.org/github/napalm-automation/napalm/builds/706059614

napalm/base/base.py
```diff
@@ -94,9 +94,14 @@ def _netmiko_open(self, device_type, netmiko_optional_args=None):
         except NetMikoTimeoutException:
             raise ConnectionException("Cannot connect to {}".format(self.hostname))
 
-        # ensure in enable mode if not force disable
-        if not self.force_no_enable:
+        # Disable enable mode if force_no_enable is true (for NAPALM drivers
+        # that support force_no_enable)
+        try:
```
mirceaulinic, line 99: Now that I'm looking at this again, it does look a little strange, wondering if the following would be more readable:

```python
if not getattr(self, 'force_no_enable', False):
    self._netmiko_device.enable()
```

mirceaulinic: Any thoughts on this @tynany / @ktbyers ?

ktbyers: I think the try/except is fine.

ktbyers: This PR looks good to me so I am fine with merging it.

PR #1254, opened 2020-07-10
Add CLI support to configuration methods 

neelimapp:
Define new optional argument (config_encoding) to specify the expected encoding in load_merge_config and load_replace_config. The get_config and compare_config methods now accept an encoding argument that can be set to "xml" or "cli". All encoding arguments default to "cli".


ktbyers:
@neelimapp Linters (black and pylama are failing)

111pontes:
Did black styling based on black==19.10b0 as specified in `requirements-dev.txt`.  However, `tox.ini` references a different version (black==18.9b0) which would still reformat `napalm/iosxr_netconf/iosxr_netconf.py`.   Should we downgrade black or do you plan to update `tox.ini`?

ktbyers:
Let me look at that...I will fix that.

PR #1256, opened 2020-07-15
Add config filter for XR-only module set

111pontes:
Exchange of configuration in XML format is limited to models in the `XR-only` module set.  The namespaces associated with the module set are gathered in the `open` method.  The `load_merge_candidate` and `load_replace_candidate` methods raise an exception (`MergeConfigException` and `ReplaceConfigException`,respectively) if unexpected data is detected.  The `get_config` and `compare_config` methods filter out any configuration associated with models not in the module set.  The models in the module set are determined using the YANG library model. If a device  doesn't support the YANG library model, explicit filtering is performed.


111pontes:
`black` failure is caused by the version inconsistency between `tox.ini` and `requirements-dev.txt`.

ktbyers:
Yeah, I didn't get to that...just update `tox.ini` to use:

black==19.10b0

111pontes:
> Yeah, I didn't get to that...just update `tox.ini` to use:
> 
> black==19.10b0

Done.  Let us know if anything else is blocking this PR.  Separate PR will be sent for missing mock data.

PR #1266, opened 2020-08-10
Commit confirm feature

ktbyers:


ktbyers:
Based on:

https://github.com/napalm-automation/napalm/pull/1179

Extended discussion in earlier PR.

ktbyers:
Behavior Description:

#### commit_config()

```python
def commit_config(self, message="", commit_confirm=False, confirm_timeout=600):
```

With commit_confirm=False will default to current standard commit_config() behavior.

With commit_confirm=True the configuration will be rolled back in 600 seconds (confirm_timeout seconds) unless a `confirm_commit()` is executed.

Will cause `has_pending_commit()` to return True for the duration of confirm_timeout.

The rollback() method can be used to immediately revert the `commit_confirm` and will set `has_pending_commit()` back to False.

Platforms that do not support the `commit_confirm` feature should generate a NotImplementError if `commit_config(commit_confirm=True) is invoked).

confirm_timeout argument will specify how much time to allow for confirm. We will need to pick a reasonable default value here. Probably should look at Juniper default value for this.

commit_config() should raise an exception if commit_config() is executed while a `commit_confirm` is already in progress.



#### confirm_commit

```python
def confirm_commit(self):
```


commit_confirm
Commit the pending commit_confirm

Should cause has_pending_commit to return False when done.

commit_confirm_revert
Rollback to state before the commit

Should cause has_pending_commit to return False when done.

has_pending_commit
Boolean indicating whether a commit_confirm is in process.

Might have platform specific questions on how we track this/know this is in process.

Need to also consider what happens if someone tries to push a second merge/replace operation while a commit operation is underway and what does NAPALM do here.


Questions:
1. Startup config
2. What happens on a reboot.

ktbyers:
Adding @gcotone @rbcollins123

ktbyers:
Another option for naming:

`commit_config(revert_in=600)`

`has_pending_commit()`

`rollback()`

`confirm_commit()` or just a second invocation of `commit_config()` when there is a pending `commit confirm`

ktbyers:
Okay, all feel free to chime-in one last time...and then we will implement and live with whatever we decide.

gcotone:
> 1. Is the commit saved to startup-config upon the initial commit_config(commit_confirm=True)?
> 2. What should be the expected behavior if the device reboots?  

I'd be in favor of not saving and booting from the _pre-commit_confirm_ config if possible. That being said, it would be difficult to have an homogeneous behaviour across vendors since the logic must be implemented on-box and may not be doable on all of them.

In either case, I think we need to document the _preferred_ behaviour as _SHOULD_ and not _MUST_.

> 3. Should we just have the executing of commit_config() a second time cause the full commit and eliminate the confirm_commit() method?

Indeed.

> 4. Still don't really like the naming here especially having both confirm_commit=True and commit_confirm.

It can be simplified by implementing `3.`

ixs:
Okay, my 2 cents.

General impression: The code is fine, it's all stubs after all.
What rubs me the wrong way a little bit is, that this feature seems very implementation specific. Examples are naming, default timeouts and assumptions such as there can only be one commit_confirm() session open at the same time.

I think we shouldn't spend a tad more time making it more generic in order to allow people to use their platform's full capabilities.


### 1. commit_config()

confirm_timeout with a default of 600 seems fine to me.
commit_confirm is named very platform specific, juniper calls it "commit confirmed", Cisco has it as "configure confirm" plus "commit confirm" on XE and arista uses "configure session commit". Just an observation, the naming is fine at the end of the day but I gotta admit, I really like the "session" name from Arista because it makes it very clear what is going on.

The assumption that "commit_config() should raise an exception if commit_config() is executed while a commit_confirm is already in progress" however seems wrong to me. Juniper can do what? 32 configuration sessions in parallel? Reading the docs seems to suggest there could be multiple commit confirmed sessions running, but it is a bit of a mess. Arista can do multiple configure sessions in parallel with the default being 5 but not sure about Cisco.
Arista however does named sessions where each session can have a nice id, is tied to a user and can have it's own commit timer.
That's quite powerful, not sure it's a good idea to limit us to only supporting _one_ session. IETF-styled "should" -> "can" might be a good call?
The Arista driver can easily support an additional argument "session_name=None" or so that would expose that functionality.

### 2. confirm_commit()

Looks good to me.

### 3. has_pending_commit()

Looks good to me.

### 4. questions

> 1. Is the commit saved to startup-config upon the initial commit_config(commit_confirm=True)?

Please don't. This would just confuse people and result in lots of foot-shooting. I think we should stay very close to the non-transaction workflow:
You do your changes, if you reboot without saving them, you end up with the last stored config before your changes.
vs.
You do your changes but do not confirm_commit them yet. On reboot your device comes up with the not confirmed configuration?

That seems wrong to me.

> 2. What should be the expected behavior if the device reboots?

Comes up with the last startup config which ideally is the one before your changes happened. What @gcotone suggested, using the pre-commit_config seems like a nice idea but I can see that going wrong when factoring in people who might log into the devices and do manual configure sessions on the terminal without saving them.
Your running state at the start of the configure session might not be what the startup session has. But automatically copying running-config to startup-config also seems wrong to me.

The approach that seems most surprise-free to me is to only touch the startup config _at_ commit time and not before. If we're happy with the running-config at that time, we commit it.

> 3. Should we just have the executing of commit_config() a second time cause the full commit and eliminate the confirm_commit() method?

Too much magic in my opinion. That way we'd be making a function implicit. Imagine two automation systems talking to the same device. A has a commit confirm session open with 5min still on the clock before rollback. B comes in, starts a regular configure session and thus commits the changes from A? That'd be unexpected I think.

> 4. Still don't really like the naming here especially having both confirm_commit=True and commit_confirm.

I see where you're coming from. Maybe life is easier _not_ trying to overload the existing commit_config() function but instead make it very explicit:
You want session/transaction support? There's different functions for these, separate from the normal commit_config() workflow.

ktbyers:
On the save to startup--what do IOS-XR and Juniper do? 

I would think they would have to save the pending configuration to startup (given the way they operate). We probably need to test these platforms behavior through a commit+confirm+reboot sequence.


Note, the below is the opposite of how NAPALM is intended to operate in general:

`You do your changes, if you reboot without saving them, you end up with the last stored config before your changes.
vs.`

i.e. NAPALM is intending to do ACID-like transactions so the design generally would save things to startup-config (obviously with the caveat that here we are talking about a `commit that might be reverted`). In other words, the current commit_config() without `confirm` must always save the change to startup-config.


```
Too much magic in my opinion. That way we'd be making a function implicit. Imagine two automation
systems talking to the same device. A has a commit confirm session open with 5min still on the
clock before rollback. B comes in, starts a regular configure session and thus commits the changes from A? 
```

Yes, this is my main concern about using `commit_config` as the `confirm` mechanism (i.e. you could have unexpected `confirms` happen).

gcotone:
>What rubs me the wrong way a little bit is, that this feature seems very implementation specific. Examples are naming, default timeouts and assumptions such as there can only be one commit_confirm() session open at the same time.

@ixs Can you name who can do more than one in parallel? From the vendors that support sessions:
 - Arista won't allow you to perform another commit in any other named session, no matter if it's a normal commit or a commit confirm.
 - Juniper will apply the commit confirms in sequence, even while using `private` sessions, effectively confirming the previous commit. Only the last one will be rollback after timer expiration.

> On the save to startup--what do IOS-XR and Juniper do?

On JunOS, startup equals to running config. If there's an ongoing commit confirm operation and the device reboots, it will never be rollback which is not really convenient.

>3. Should we just have the executing of commit_config() a second time cause the full commit and eliminate the confirm_commit() method?
Too much magic in my opinion. That way we'd be making a function implicit. Imagine two automation
systems talking to the same device. A has a commit confirm session open with 5min still on the
clock before rollback. B comes in, starts a regular configure session and thus commits the changes from A? 

It really depends which candidate configuration is being applied with `commit_config()`. If you're unsure of what you're commiting, to me it looks more like automation workflow problem rather than something we should be addressing in the driver.

rbcollins123:
> ```
> Too much magic in my opinion. That way we'd be making a function implicit. Imagine two automation
> systems talking to the same device. A has a commit confirm session open with 5min still on the
> clock before rollback. B comes in, starts a regular configure session and thus commits the changes from A? 
> ```
> 
> Yes, this is my main concern about using `commit_config` as the `confirm` mechanism (i.e. you could have unexpected `confirms` happen).
I agree;  we discussed the notion of overloading previously and already threw that out in favor of being more explicit.

ktbyers:
Okay, let's do this. I like this a bit better as it eliminates the duplicate similar names and simplifies things a bit.

```
commit_config(revert_in=600)

has_pending_commit()

rollback()

confirm_commit()
```

Also commit_config will always save to startup-config regardless of whether there is a confirm or not. I think this is more consistent with current NAPALM behavior and will also avoid inconsistencies between platforms where there are no other options.

I will update the PR to reflect this naming.

ktbyers:
Sorry for flip-flopping on some of this (from earlier discussions)...some of the names I thought were still confusing and hence thought it was better to improve/fix now.

ktbyers:
@mirceaulinic If you get a chance, can you review this and we will merge this into `develop`

ktbyers:
Next-steps:
1. Implement on one platform. @rbcollins123 Do you want to implement this on EOS.
2. Once implemented on one platform, I will implement the test cases here (https://github.com/ktbyers/napalm_custom_test). Others are welcome to help on the tests. 
3. Once we have the tests and implementation done on one platform, then we can merge that into `develop` and others can roll it out to other platforms.

ixs:
@ktbyers I did a quick shot at a super-simple EOS implementation, just to see if the logic is sound and works. PR is at https://github.com/napalm-automation/napalm/pull/1269.

The confirm_commit feature is sound so far and seems to work fine.
I just have some concerns about the following part:

> Also commit_config will always save to startup-config regardless of whether there is a confirm or not. I think this is more consistent with current NAPALM behavior and will also avoid inconsistencies between platforms where there are no other options.

Worst situation of that design is, you get a unreachable machine if the configuration is b0rked and you get a power failure. Machine comes up with the b0rked config and it's game-over.

But even disregarding the worst case, the auto-rollback after X minutes will mean that your startup configuration is different than the running configuration and you now need a manually triggered action to ensure that run == startup.
If you do a manual rollback() call, that is taken care of but not if you wait for the automatic rollback.

That's a bad situation IMHO as automated rollback and manual rollback should be leaving the device in the same state.

Or did I misunderstand your intention somewhere?

Also, looking at the code for EOS I do not see where "save to startup-config" would be in line with current NAPALM behavior. What happens is that startup-config gets copied to a rollback copy and in case the config change gets rolled-back, this config copy is copied to running-config. So any non-committed changes to startup would be dropped in case of a rollback.
Startup-config is only ever touched in case of a commit.
And if we're going for a two-stage commit, the startup-config should only ever be touched in case the confirm_commit happens. Not at the start of the transaction/session/confirm_commit 1st stage.

Again, I suspect I might have misunderstood your design but that's why I wrote the POC code. It's always easier to talk with something concrete on hand. I'd appreciate if you could clarify your design a bit. Thanks.

ktbyers:
@ixs Yes there could be rare edge cases where you could run into serious problems (of course, you are already in this situation regardless of what NAPALM does if you don't have a terminal server and OOB access). In other words, if your reliability requirements are high enough where those rare edge cases matter, then you should already have OOB access.

```
But even disregarding the worst case, the auto-rollback after X minutes will mean that your startup configuration is different than the running configuration and you now need a manually triggered action to ensure that run == startup.
If you do a manual rollback() call, that is taken care of but not if you wait for the automatic rollback.
```

Good point, but note your statements are platform specific i.e. Juniper and IOS-XR will already do this (i.e. the auto-rollback on those two platforms will rollback both the in-memory config and startup config). Note, I am making some assumptions here that I need to test, but given my understanding of both IOS-XR and Junos, this is the behavior I would expect.

Let's do this...let's defer the `what happens with startup-config during an in process commit confirm` until we get to the actual implementation (it will be easier to see then if there are reasonable ways to implement this when we look at the mechanics available on each platform).

Not following you on this?

```
Also, looking at the code for EOS I do not see where "save to startup-config" would be in 
line with current NAPALM behavior.
```

Here is the NAPALM EOS code:

```
    def rollback(self):
        """Implementation of NAPALM method rollback."""
        commands = ["configure replace flash:rollback-0", "write memory"]
        self.device.run_commands(commands)
```

So NAPALM rolls back the configuration to the previous state (before commit_config()) and then also saves that configuration back to startup-config using `write memory`.

My two main concerns on not saving to startup-config are:
1. Inconsistent behavior between core napalm platforms (i.e. Junos/IOS-XR do one thing; IOS and EOS do something else).
2. Inconsistent behavior for this and how NAPALM typically operates on commit_config (of course, one could argue as you are that for a "commit confirm" operation this typical behavior shouldn't apply).

Anyways, let's punt what happens to startup-config to the actual implementation and then I can look at what options EOS and IOS have available for having the auto-rollback get saved to startup-config (and see if it is even practical in a reasonable way...or not).


Regards, Kirk

ixs:
@ktbyers I generally agree on the OOB access. Of course, with my luck, I'm gonna take down the switch that has the OOB stuff connected to it. 😂

With you on the "let's defer decision about startup-config behavior" suggestion. I suspect it might be a good idea to actually make that platform-specific. Because it seems that the different platforms are just too different here.

About the current behavior of EOS code, quoting from https://github.com/napalm-automation/napalm/blob/develop/napalm/eos/eos.py#L333:

```
    def commit_config(self, message=""):
[...]
        commands = [
            "copy startup-config flash:rollback-0",
            "configure session {}".format(self.config_session),
            "commit",
            "write memory",
        ]
```

What I meant: The current EOS rollback procedure does copy the startup-config to the rollback-0 backup. This means that calling rollback() will discard any manual (or automated) changes that had not been written to startup-config.

To be clear: I do not see this as a problem, I am actually totally fine with discarding such changes. I am not suggesting that this should be a blocker for the commit confirm feature. This is just to make clear that platforms will be different and while I appreciate the attempt to generalise them, I prefer a platform to diverge slightly in the implementation rather than trying to (badly) emulate some other platform.

> My two main concerns on not saving to startup-config are:
> 1. Inconsistent behavior between core napalm platforms (i.e. Junos/IOS-XR do one thing; IOS and EOS do something else).

Totally valid concern. But I feel there is only so much we can effectively do considering that there are platform differences. Trying to abstract everything away would make NAPALM not a thin layer anymore with more added complexity. Not something I think we want.

> 2. Inconsistent behavior for this and how NAPALM typically operates on commit_config (of course, one could argue as you are that for a "commit confirm" operation this typical behavior shouldn't apply).

😁 Thank you. This means I've gotten my point across.

But as I said before, have a look at https://github.com/napalm-automation/napalm/pull/1269, it's a possible implementation for EOS. I wanted something I can test out to see if the proposed general feature makes sense. And it does.
I have not touched IOS or Junos in years though, so I would leave that up to someone else...

ktbyers:
On the copying from startup-config instead of running-config (for rollback)...

Yeah, I didn't really like that and possibly/probably could be convinced that is a bug in the EOS driver, but there also is the requirement with NAPALM, that all commit_config operations (currently) save to startup-config.

Consequently, there really is the expectation in NAPALM that running-config and startup-config are identical. In other words, someone would have to have made a manual config change or an automated change by a different system that is not-NAPALM and not saved it. 

And since they didn't save it to startup-config, they have left things in a very ambiguous state (as far as which configuration they want to continue using in a rollback event).

Anyways, let's punt the decision to implementation...as if there is no practical way to actualize the save to startup-config consistently on auto-rollback, then we might have the issue decided for us by the platform (i.e. if there is no reasonable way to do this, then we would be forced into not doing it or in abandoning the entire feature and I don't think this issue is important enough to abandon the feature).


Regards, Kirk

ixs:
@ktbyers Absolutely agree, let's punt that part of the issue. confirm commit is way more important than the intrinsics of some ambiguities during a rollback feature or so...

Issue #958, opened 2019-04-08
newest iosxr (6.5.3) causes pyIOSXR.exceptions.XMLCLIError with get_route_to function


jbotello7381:
During my testing with Cisco NCS5500 and there latest IOSXR (6.5.3) I notice issues with commands that depend on the function **get_route_to** on iosxr.py module.  

Validation

```
>>> from pyIOSXR import IOSXR
>>>
>>> device = IOSXR(hostname='pr04.lab01', username='user', password='password', port=22, timeout=120)
>>> device.open()
>>> route = '''<Get>
  <Operational>
    <RIB>
      <VRFTable>
        <VRF>
          <Naming>
            <VRFName>default</VRFName>
          </Naming>
          <AFTable>
            <AF>
              <Naming>
                <AFName>IPv4</AFName>
              </Naming>
              <SAFTable>
                <SAF>
                  <Naming>
                    <SAFName>Unicast</SAFName>
                  </Naming>
                  <IP_RIBRouteTable>
                    <IP_RIBRoute>
                      <Naming>
                        <RouteTableName>default</RouteTableName>
                      </Naming>
                      <RouteTable>
                        <Route>
                          <Naming>
                            <Address>6.6.6.6</Address>
                            <PrefixLength>32</PrefixLength>
                          </Naming>
                        </Route>
                      </RouteTable>
                    </IP_RIBRoute>
                  </IP_RIBRouteTable>
                </SAF>
              </SAFTable>
            </AF>
          </AFTable>
        </VRF>
      </VRFTable>
    </RIB>
  </Operational>
</Get>'''
>>> device.make_rpc_call(route)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "build/bdist.linux-x86_64/egg/pyIOSXR/iosxr.py", line 151, in make_rpc_call
  File "build/bdist.linux-x86_64/egg/pyIOSXR/iosxr.py", line 416, in _execute_rpc
pyIOSXR.exceptions.XMLCLIError:
[...]

PR #1267, opened 2020-08-11
IOS-XR get_route_to fix pyIOSXR.exceptions.XMLCLIError for some versions

chonty:
See https://github.com/napalm-automation/napalm/issues/958

On some versions of IOS-XR, `get_route_to` fails because the XML path (not sure what else to call it) had an element renamed from `IP_RIBRoute `to `IP_RIBRouteTableName`.

This PR fixes issue #958 and maintains backwards-compatibility using option B from my comment in the same issue.

Global variable `IP_RIBRoute` is used to store the default name 'IP_RIBRoute' and if the RPC call fails, it will suppress the exception (something still pops up and I don't know how to stop it), change the value of var `IP_RIBRoute` to 'IP_RIBRouteTableName' and try the RPC call again.  Doing it this way means that we only need to fail once within the lifetime of a Napalm instance.

    >>> import napalm
    >>> driver = napalm.get_network_driver("iosxr")
    >>> d = driver(
    ...     hostname="10.10.10.9",
    ...     username="napalm",
    ...     password="napalm",
    ... )
    >>> d.open()
    >>> a = d.get_route_to("10.202.0.56/29")
    
    Original call was: <?xml version="1.0" encoding="UTF-8"?><Request MajorVersion="1" MinorVersion="0"><Get><Operational><RIB><VRFTable><VRF><Naming><VRFName>default</VRFName></Naming><AFTable><AF><Naming><AFName>IPv4</AFName></Naming><SAFTable><SAF><Naming><SAFName>Unicast</SAFName></Naming><IP_RIBRouteTable><IP_RIBRoute><Naming><RouteTableName>default</RouteTableName></Naming><RouteTable><Route><Naming><Address>10.202.0.56</Address><PrefixLength>29</PrefixLength></Naming></Route></RouteTable></IP_RIBRoute></IP_RIBRouteTable></SAF></SAFTable></AF></AFTable></VRF></VRFTable></RIB></Operational></Get></Request>
    >>> a
    {'10.202.0.56/29': [{'current_active': True, 'last_active': False, 'age': 2460657, 'next_hop': '10.10.10.10', 'protocol': 'bgp', 'outgoing_interface': '', 'preference': 12, 'selected_next_hop': False, 'inactive_reason': '', 'routing_table': 'default', 'protocol_attributes': {}}]}
    >>> a
    {'10.202.0.56/29': [{'current_active': True, 'last_active': False, 'age': 2460657, 'next_hop': '10.10.10.10', 'protocol': 'bgp', 'outgoing_interface': '', 'preference': 12, 'selected_next_hop': False, 'inactive_reason': '', 'routing_table': 'default', 'protocol_attributes': {}}]}

In the above Python output, the `Original call was:...` will appear on the first get_route_to call if IP_RIBRoute doesn't work.  I couldn't figure out how to silence it.


mirceaulinic (approved):
LGTM. Thanks @chonty!

PR #1272, opened 2020-08-18
Fix EOS get_arp_table for static ARP records

praniq:
For static ARP entries EOS eAPI gives records without "age" key. Therefore NAPALM throws an exception in get_arp_table during:
`age = float(neighbor.get('age'))`

Fixing this bug.


napalm/eos/eos.py
```diff
@@ -966,7 +966,7 @@ def get_arp_table(self, vrf=""):
             interface = str(neighbor.get("interface"))
             mac_raw = neighbor.get("hwAddress")
             ip = str(neighbor.get("address"))
-            age = float(neighbor.get("age"))
+            age = float(neighbor.get("age", 0.0))
```
bewing, line 969: Should we use `-1.0` to mark static entries?  I notice nxos uses this approach, but IOS doesn't.  Perhaps worth standardizing on?

praniq: I would vote for -1.0 for all OSes, as looks like "-1" is a standard value for N/A in Napalm.

mirceaulinic: Agreed. @praniq - please feel free to update this branch, then I can merge it.

praniq: Updated EOS, IOS and IOS XR. Please have a look. Thanks

mirceaulinic (approved):
LGTM. Thanks @praniq

PR #1279, opened 2020-08-26
Add get_vlans getter to EOS network driver

yzguy:
This would add the get_vlans() getter to the EOS network driver, which is currently not supported for the platform

### DIFF
--- a/napalm/eos/eos.py
+++ b/napalm/eos/eos.py
@@ -1990,3 +1990,16 @@ class EOSDriver(NetworkDriver):
                     )
             ping_dict["success"].update({"results": results_array})
         return ping_dict
+
+    def get_vlans(self):
+        command = ["show vlan"]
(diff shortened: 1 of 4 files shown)


yzguy:
Not sure if there are any more tests to add for this new addition, there was a pass on

```
test/eos/test_getters.py::TestGetter::test_get_vlans[normal] PASSED
```

mirceaulinic:
Yep, that's all it needs @yzguy - looking good! Just a small Black formatting issue: https://travis-ci.org/github/napalm-automation/napalm/jobs/721233843#L228, then we should be ready to get this in. (To correct that issue, in your environment, just run `black .` and commit the changes).

mirceaulinic (approved):
LGTM. Thanks for this addition @yzguy! I'll include this in the next release 3.2.0, scheduled to go out early next week.

Issue #1281, opened 2020-08-28
cisco ios get_probes_config: incorrect "ip sla" parsing


javcasalc:
Hi

I've seen this code is pretty old (3 years, https://github.com/napalm-automation/napalm/commit/1bdc4d171baa75a74005ccedf4d7e4ab24c28b97#diff-5a3fb7da3626f2d9b4eed6d1e6011e74R1905), and it turns out that it does not work with newer IOS XE like Version 16.12.3 for example. 

I would change the probes_regex definition from
```
probes_regex = r"ip\s+sla\s+(?P<id>\d+)\n" \
            "\s+(?P<probe_type>\S+)\s+(?P<probe_args>.*\n).*" \
            "\s+tag\s+(?P<name>\S+)\n.*" \
            "\s+history\s+buckets-kept\s+(?P<probe_count>\d+)\n.*" \
            "\s+frequency\s+(?P<interval>\d+)$"
```
to
```
probes_regex = (
            r"ip\s+sla\s+(?P<id>\d+)\n"
            r"\s+(?P<probe_type>\S+)\s+(?P<probe_args>.*)\n.*"
            r"\s+tag\s+(?P<name>\S+)\n.*"
            r"\s+frequency\s+(?P<interval>\d+)\n.*"
            r"\s+history\s+buckets-kept\s+(?P<probe_count>\d+)$"
        )
```

and also I think you have to exec ```sh run | section ip sla [0-9]``` 
```
csr1000v#sh run | section ip sla [0-9]
ip sla 1
 icmp-echo 10.0.0.2 source-interface GigabitEthernet11
 tag napalm_test
 frequency 15
 history buckets-kept 60
```

instead of 
```sh run | include ip sla [0-9]``` 

```
csr1000v#sh run | include ip sla [0-9]
ip sla 1
```
in https://github.com/napalm-automation/napalm/commit/1bdc4d171baa75a74005ccedf4d7e4ab24c28b97#diff-5a3fb7da3626f2d9b4eed6d1e6011e74R1918

Is this ok?

J.

mirceaulinic:
Hi J. Yes, please feel free to submit a PR to correct this, as described above (assuming the current tests would pass -- but another test case would be ideally added for this particular scenario). Thanks.

javcasalc:
Hi

kindly check https://github.com/napalm-automation/napalm/pull/1285

[...]

PR #1285, opened 2020-08-31
Update get probe code to make it work it IOS XE 16.12.x

javcasalc:
This patch makes ios.py get_probes_config() getter able to work with new IOS XE versions:
```
csr1000v#show version | i , Version 
Cisco IOS XE Software, Version 16.12.03
Cisco IOS Software [Gibraltar], Virtual XE Software (X86_64_LINUX_IOSD-UNIVERSALK9-M), Version 16.12.3, RELEASE SOFTWARE (fc5)
```
```
csr1000v#sh run | section ip sla     
ip sla 1
 icmp-echo 10.0.0.2 source-interface GigabitEthernet11
 tag napalm_test
 frequency 15
 history buckets-kept 60
ip sla schedule 1 life forever start-time now
csr1000v#
```

This code is pretty inflexible as it is based on python regex multiline, so the block (ip sla config) must match all lines and options. It should be improved.


javcasalc:
https://github.com/napalm-automation/napalm/issues/1281

napalm/ios/ios.py
```diff
@@ -2697,18 +2697,22 @@ def process_mac_fields(vlan, mac, mac_type, interface):
...
+            r"\s+buckets-kept\s+(?P<probe_count0>\d+))|(\s+history\s+buckets-kept"
+            r"\s+(?P<probe_count1>\d+)\n.*\s+frequency\s+(?P<interval1>\d+)))"
         )
+
         probe_args = {
             "icmp-echo": r"^(?P<target>\S+)\s+source-(?:ip|interface)\s+(?P<source>\S+)$"
         }
         probe_type_map = {"icmp-echo": "icmp-ping"}
-        command = "show run | include ip sla [0-9]"
+        command = "show run | section ip sla [0-9]"
```
mirceaulinic, line 2715: I don't have a lot of experience on IOS platforms, but my understanding from past PRs / conversations is that older versions may not support `| section` (while `| include` should always be there). Could you clarify what's the advantage or why you chose to switch @javcasalc?

mirceaulinic (changes requested):
Thanks for this PR @javcasalc. The regex looks good, but in make it fruitful, we'll need a new test case with the output you're seeing on your platform (to show why the current code is not working, and prove that the new regex would solve it).

napalm/ios/ios.py
javcasalc: According to Cisco doc, the right flag is ```section```: https://www.cisco.com/c/en/us/td/docs/ios/fundamentals/command/reference/cf_book/cf_s1.html

```
Examples

The following examples compare the filtering characteristics of the show running-config | include command with the show running-config | section command. The first example gathers just the lines from the configuration file with "interface" in them.

Router# show running-config | include interface


interface Ethernet0/0 
interface Ethernet1/0 
interface Serial2/0 
interface Serial3/0


The next example uses the show command section command to gather the lines in the configuration file with "interface" in them as well as any lines associated with those entries. In this example, interface configuration information is captured.

Router# show running-config | section include interface


interface Ethernet0/0 
 shutdown 
 no cdp enable

interface Ethernet1/0 
 shutdown 
 no cdp enable 
interface Serial2/0 
 shutdown 
 no cdp enable 
interface Serial3/0 
 shutdown 
 no cdp enable
```

I did some tests in GNS3 with some old cisco routers (cisco 3640 with IOS 12.4; cisco 7200 with IOS 15.2) and I do confirm that ```section``` is the right choice:
imagen


Regarding the test case, I've included an extra "ip sla 3" entry, formatted with the new ios format, in order to test the correct parsing of both styles (old and new styles). Is this ok?

mirceaulinic: Fair enough... any thoughts on this @ktbyers ?

mirceaulinic (approved):

PR #1288, opened 2020-09-07
Fix mapping for Juniper devices with Netmiko

vincentbernat:
`junos` was incorrectly mapped to `juniper_eos` instead of
`juniper_junos`. It shouldn't matter much as, AFAIK, `junos` is not
using Netmiko.


mirceaulinic (approved):
Thanks @vincentbernat

PR #1308, opened 2020-10-21
Add full canonical TwentyFiveGigE to canonical_map

brandomando:


napalm/base/canonical_map.py
```diff
@@ -80,6 +80,7 @@
     "Tunnel": "Tunnel",
     "Tun": "Tunnel",
     "Tu": "Tunnel",
+    "TwentyFiveGigE": "TwentyFiveGigE",
```
mirceaulinic, line 83: Did you want this to be `"TwentyFiveGigE": "TwentyFiveGigEthernet"`?

brandomando: I only have Cisco devices to reference, but those that I have all call 25G interfaces TwentyFiveGigE, and it is the same in the reverse mapping

mirceaulinic: Right. The canonical map has the following logic: on the left side it's what you see on the device (e.g., `Gi`), and on the right hand side it's what it actually means (e.g., `GigabitEthernet`).

brandomando: I get where you are coming from there, but from what I can see, no manufacturers call it `TwentyFiveGigabitEthernet`, and the canonical map, at least in my case, is to map the abbreviated versions of the interface names to the full display names of the interfaces, so if I run a show command that uses brevity, then it would say `Twe1/1/1/` and the canonical map could be used to map it to `TwentyFiveGigE`, which is the full canonical name of the interface. If mapping to `TwentyFiveGigabitEthernet`, and employed to send commands to a device, it would cause failures, so for me, my PR is as I would intend it to be in this case.

mirceaulinic: I'm not sure we're on the same line... you're mapping `TwentyFiveGigE` to `TwentyFiveGigE`, which doesn't make a lot of sense. `TwentyFiveGigE` is also the short version, or in the canonical map, we want to map the short version to the long version. If I do a `show int ?` on a device, I'm seeing:

```
  TwentyFiveGigE   TwentyFiveGigabitEthernet/IEEE 802.3 interface(s) | short name is TF`
```

So the complete, accurate name is `TwentyFiveGigabitEthernet`, which is what should the canonical map point to.

mirceaulinic:
FYI, I've opened #1329 if you want to take a look at what the mapping should be looking like.

PR #1315, opened 2020-10-30
Add ncclient dependency for iosxr_netconf driver

111pontes:
Make ncclient dependency explicit in `requirements.txt`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,3 +13,4 @@ junos-eznc>=2.2.1
 ciscoconfparse
 scp
 lxml>=4.3.0
+ncclient


requirements.txt
```diff
@@ -13,3 +13,4 @@ junos-eznc>=2.2.1
 ciscoconfparse
 scp
 lxml>=4.3.0
+ncclient
```
mirceaulinic, line 16: I think ncclient should already be a dependency of `junos-eznc` referenced above, unless we need to pin to a specific ncclient version. What issues are you seeing @111pontes?

111pontes: You're correct.  `ncclient` is a dependency of `junos-eznc`.  I don't think it make senses to have a driver depend on the dependencies of other drivers.  That could lead to changes in the implementation of a driver breaking other drivers. For instance, what if `junos-eznc` decided to stop using `ncclient` at some point in the future?

mirceaulinic: Well, that's how dependencies work, we can't pull the entire dependency tree here just because. If `junos-eznc` would stop using `ncclient`, would cease to be `junos-eznc`.
But :man_shrugging: will merge the PR if you really want it here, just the CI is currently failing. For the sake of it though, please pin to `ncclient>=0.6.3` (or any newer version than that that satisfies your needs).

ttosttos:
Thanks.  Will send a PR soon setting a floor for the `ncclient version`.

111pontes:
Thanks. Will send a PR soon setting a floor for the `ncclient` version.

requirements.txt
111pontes: Thanks. Will send a PR soon setting a floor for the ncclient version.

ktbyers:
Fixed here:

https://github.com/napalm-automation/napalm/pull/1419

Issue #1320, opened 2020-11-12
IOS XE Traceroute ttl


MatthiasGabriel:
### Description of Issue/Question
The command string created when calling traceroute on IOS device leads to an incomplete command and to an empy `success` in the result.

The problem is that the generated command in the IOS driver is missing the minimum ttl.
The output from the device is then `'^% Invalid input detected at \'^\' marker.`

I noticed that the minimum ttl got removed in the pr #1174. Does IOS XE implement different traceroute commands on different versions?

### Did you follow the steps from https://github.com/napalm-automation/napalm#faq
(Place an ``x`` between the square brackets where applicable)

- [x] Yes
- [] No

### napalm version
(Paste verbatim output from `pip freeze | grep napalm` between quotes below)

```
napalm==3.2.0
```

### Network operating system version
(Paste verbatim output from `show version` - or equivalent - between quotes below)

```
Cisco IOS XE Software, Version 16.11.01a
```

### Steps to Reproduce the Issue
1. Use IOS driver to create a new device connection and call traceroute.

```python
driver = napalm.get_network_driver("ios")
    device = driver(hostname="10.20.0.31", username=<user>, password=<pw>)
    with device as napalm_device:
        napalm_result = napalm_device.traceroute(destination=<reachable_destination>)
```
2. Analyse the command and output variables in the ios driver.
3. Analyse the command variable on the device.
![image](https://user-images.githubusercontent.com/11162099/98903082-27c88580-24b7-11eb-943d-c46a3e522f4b.png)
![image](https://user-images.githubusercontent.com/11162099/98902951-edf77f00-24b6-11eb-8e3c-7f2e90cd50c7.png)

### Error Traceback
(Paste the complete traceback of the exception between quotes below)

```

```

MatthiasGabriel:
During the analysis of this problem I also noticed that there is no possiblity to skip the ttl parameter.
If one calls the method without the ttl parameter, the default value of 255 is taken.
If one calls the method with ttl parameter set to None and therefore the ttl part is not added to the command string, the calculation of the max_loops variable fails.
Maybe for the calculation another default value should be used. (It would be good to use the default ttl of IOS devices which according to a little googling is 30.)
Let me know what you think of this.

MatthiasGabriel:
Maybe the minimum ttl  should be set to 1 instead of 0
![image](https://user-images.githubusercontent.com/11162099/98925213-01fea900-24d6-11eb-8683-c735134ba9ee.png)

[...]

PR #1321, opened 2020-11-12
Set minimum ttl to zero to fix issue #1320

MatthiasGabriel:
Add minimum ttl to ios traceroute command.
Fixes Issue #1320


napalm/ios/ios.py
```diff
@@ -3284,7 +3284,7 @@ def traceroute(
             command += " source {}".format(source)
         if ttl:
             if isinstance(ttl, int) and 0 <= ttl <= 255:
-                command += " ttl {}".format(str(ttl))
+                command += " ttl 0 {}".format(str(ttl))
```
mirceaulinic, line 3287: I'm not sure this syntax is supported on every IOS version. Did you give it a try @MatthiasGabriel?

MatthiasGabriel: Any suggestions which versions I should check?
Also as now mentioned in the issue #1320 I'm not sure about the value 0 either. In my opinion 1 seems like the "default".

mirceaulinic: Gah, not sure what to say, I fortunately don't have a Cisco IOS to check this out. Perhaps @ktbyers would have more input on this matter? I'd also suggest you to engage with the community and see what other IOS users think about it.

MatthiasGabriel: Thanks for your feedback @mirceaulinic 
During the day I added some new findings to the issue #1320

mirceaulinic (approved):

PR #1326, opened 2020-11-23
Add possibility to mock traceroute and ping results

MatthiasGabriel:
Allow to mock traceroute and ping by writing mock files as used by getters etc.


napalm/base/mock.py
```diff
@@ -33,7 +33,7 @@ def raise_exception(result):
 
 
 def is_mocked_method(method):
-    mocked_methods = []
+    mocked_methods = ['traceroute', 'ping']
```
dbarrosop, line 36: black is complaining, I think it is because of the single quotes. You can just run it yourself to double check and/or to have it format it automatically

MatthiasGabriel: sorry about that.

dbarrosop (approved):

dbarrosop:
LGTM but I will let @mirceaulinic merge :)

PR #1328, opened 2020-11-23
Propagate NetMikoTimeoutException error message to ConnectionException

mk-fg:
Hi,

Currently NAPALM reports any NetmikoTimeoutException as "Cannot connect to <host>".

It unfortunately hides potential reason for the error, if netmiko communicates it in its error message.
In my cause, for example, this (unfortunately unhelpful) error meant "permission denied accessing ssh key", which is hard to guess from "cannot connect to" message, so this PR simply adds whole netmiko error message after a dash.

There is related https://github.com/ktbyers/netmiko/pull/2052 issue in my case for hopefully making netmiko error messages contain more information, but regardless of how non-desciptive they might be, don't think there should be any harm propagating info from there to NAPALM users (and any apps using it), and might help to understand what the problem is.
( https://github.com/ktbyers/netmiko/pull/2052#issuecomment-732409571 comment there might also highlight the issue with too-compact error messages)

Thanks.

### DIFF
--- a/napalm/base/base.py
+++ b/napalm/base/base.py
@@ -91,8 +91,8 @@ class NetworkDriver(object):
                 timeout=self.timeout,
                 **netmiko_optional_args
             )
-        except NetMikoTimeoutException:
-            raise ConnectionException("Cannot connect to {}".format(self.hostname))
+        except NetMikoTimeoutException as e:
+            raise ConnectionException("Cannot connect to {}\n\n{}".format(self.hostname, e))
 
         # Disable enable mode if force_no_enable is true (for NAPALM drivers
         # that support force_no_enable)
--- a/napalm/base/base.py
+++ b/napalm/base/base.py
@@ -92,7 +92,9 @@ class NetworkDriver(object):
                 **netmiko_optional_args
             )
         except NetMikoTimeoutException as e:
-            raise ConnectionException("Cannot connect to {}\n\n{}".format(self.hostname, e))
+            raise ConnectionException(
+                "Cannot connect to {}\n\n{}".format(self.hostname, e)
+            )
 
         # Disable enable mode if force_no_enable is true (for NAPALM drivers
         # that support force_no_enable)
@@ -1460,7 +1462,7 @@ class NetworkDriver(object):
             {
                 'error': 'unknown host 8.8.8.8.8'
             }
-            """
+        """
         raise NotImplementedError
 
     def get_users(self):
--- a/napalm/base/helpers.py
+++ b/napalm/base/helpers.py
@@ -147,7 +147,7 @@ def cisco_conf_parse_objects(cfg_section, config):
 
 
 def regex_find_txt(pattern, text, default=""):
-    """""
+    """ ""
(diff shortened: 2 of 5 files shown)


napalm/base/base.py
```diff
@@ -91,8 +91,8 @@ def _netmiko_open(self, device_type, netmiko_optional_args=None):
                 timeout=self.timeout,
                 **netmiko_optional_args
             )
-        except NetMikoTimeoutException:
-            raise ConnectionException("Cannot connect to {}".format(self.hostname))
+        except NetMikoTimeoutException as e:
+            raise ConnectionException("Cannot connect to {} - {}".format(self.hostname, e))
```
ktbyers, line 95: Should probably use newlines between the NAPALM message and the message contained in "e" i.e. something like:

```python
raise ConnectionExcpetion(f"Cannot connect to {self.hostname}\n\n{e}")
```

So it becomes less confusing with the two strings being combined together.

mk-fg: It's somewhat unconventional to use multiline error messages at all in python exceptions.
E.g. don't think stdlib does it anywhere I can remember, and quick google suggests that people can even be surprised by it and don't expect this ( https://github.com/filp/whoops/issues/298 https://stackoverflow.com/questions/51449888/ ... ).

But on the other hand, this specific error message is also known to be multiline in advance, so guess adding more newlines here shouldn't really be a problem.
Updated the PR to put two newlines instead of a dash.

Although without newline, IMO error message will work nicely as a more common `one-line summary<newline><newline>details` format, used in all git commits and many other places, as a middle-ground between single-line and multi-line formats.
I.e. something like this:
```
ConnectionException: Cannot connect to 10.0.0.98 - TCP connection to device failed.

Common causes of this problem are:
1. Incorrect hostname or IP address.
2. Wrong TCP port.
3. Intermediate firewall blocking access.

Device settings: cisco_ios 10.0.0.98:22
```

But again, don't think it matters too much, as long as underlying error is actually reported in some way.

mirceaulinic (changes requested):
Looks good, just run a `black .` to correct this https://travis-ci.org/github/napalm-automation/napalm/jobs/745575209.

mk-fg:
I've installed black and ran `black .` and it made a whole lot of seemingly junk-changes - http://ix.io/2Fi6
I think you might want to check and maybe commit these separately, as it'd otherwise obfuscate the intention of a two-liner commit here.

mk-fg:
> I think you might want to check and maybe commit these separately

Added that commit myself, in case that was the idea behind the suggestion, but again, as it seem to change stuff in random unrelated places, please check if it's not just junk-diffs due to black configuration/version mismatch or something like that.

mirceaulinic:
@mk-fg have you installed `black` 20.8b1 (from https://github.com/napalm-automation/napalm/blob/develop/requirements-dev.txt#L1)?

mk-fg:
Guess I'll just close it in frustration, before this gets any more weird.
Hopefully you might see the value in one-liner fix here and make the change, but if not - oh well, sorry for the noise.

Issue #1312, opened 2020-10-28
IOS XR: AttributeError: 'CiscoXrSSH' object has no attribute 'receive_data_generator'


rlanyi:
### Description of Issue

Calling Cisco ASR9K devices (IOS XR) via SSH from Napalm CLI is unreliable, seemingly only very small responses can be gathered. With most calls I get `AttributeError: 'CiscoXrSSH' object has no attribute 'receive_data_generator'` error, sometimes during the built-in `get_facts` call, sometimes during the call passed on the command line. Even if my command is also `get_facts` it happens that the built-in one runs without issues but the one provided by me fails. Sometimes it fails during the built-in call. It can never show the response of larger calls like `get_bgp_neighbors`.

### Did you follow the steps from https://github.com/napalm-automation/napalm#faq

- [x] Yes
- [ ] No

### napalm version
I'm running napalm from the Dockerized version of [peering-manager](https://github.com/peering-manager/peering-manager/) v1.2.1 .

```
napalm==3.1.0
```

### Network operating system version

```
Cisco IOS XR Software, Version 6.1.2[Default]
```

#### 1st try: built-in `get_facts` call runs properly, `get_facts` passed via the command line fails
<details><summary>Click for details</summary>
<p>

```
root@1dab935bca2b:/opt/peering-manager# napalm --debug --user root --password xxxx --vendor iosxr --optional_args "ssh_config_file='/root/.ssh/config'" router1 call get_facts
2020-10-28 20:41:43,221 - napalm - DEBUG - Starting napalm's debugging tool
2020-10-28 20:41:43,221 - napalm - DEBUG - Gathering napalm packages
2020-10-28 20:41:43,221 - napalm - DEBUG - napalm==3.1.0
2020-10-28 20:41:43,221 - napalm - DEBUG - get_network_driver - Calling with args: ('iosxr',), {}
2020-10-28 20:41:43,221 - napalm - DEBUG - get_network_driver - Successful
2020-10-28 20:41:43,221 - napalm - DEBUG - __init__ - Calling with args: (<class 'napalm.iosxr.iosxr.IOSXRDriver'>, 'router1', 'root'), {'password': '*******', 'timeout': 60, 'optional_args': {'ssh_config_file': '/root/.ssh/config'}}
2020-10-28 20:41:43,222 - napalm - DEBUG - __init__ - Successful
[...]

PR #1332, opened 2020-12-03
Issue #1312: Remove the reference to Netmiko's receive_data_generator

mirceaulinic:
Solves #1312.

### DIFF
--- a/napalm/pyIOSXR/iosxr.py
+++ b/napalm/pyIOSXR/iosxr.py
@@ -348,7 +348,7 @@ class IOSXR(object):
                             delay_factor=delay_factor,
                         )
         else:
-            output += self._netmiko_recv()  # try to read some more
+            output += self.device._read_channel_timing()  # try to read some more
 
         if "0xa3679e00" in output or "0xa367da00" in output:
             # when multiple parallel request are made, the device throws one of the the errors:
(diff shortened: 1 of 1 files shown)


ktbyers (approved):

napalm/pyIOSXR/iosxr.py
```diff
@@ -348,7 +348,7 @@ def _send_command(
                             delay_factor=delay_factor,
                         )
         else:
-            output += self._netmiko_recv()  # try to read some more
+            output += self._read_channel_timing()  # try to read some more
```
ktbyers, line 351: @mirceaulinic Any idea when this `else` clause actually is used? I would be inclined to just use Netmiko's `read_channel()` (but I also don't have a good understanding exactly when this else clause would be used).

Also shouldn't this be `self.device._read_channel_timing()`?

ktbyers (changes requested):
Doesn't it need to be `self.device._read_channel_timing()`?

napalm/pyIOSXR/iosxr.py
mirceaulinic: > Also shouldn't this be self.device._read_channel_timing()?

Ah it should be indeed, great catch! :man_facepalming: 

> Any idea when this else clause actually is used?

I seem to recall this due to the XML agent taking a lot of time to dump the response, and need to wait a little longer, but can't tell for sure, it's been a long time since then. :face_with_head_bandage:

ktbyers: Okay, let me look at it just a little bit...I just want to make sure that we shouldn't be using `read_channel()` instead of `read_channel_timing()` (i.e. that read_channel() would be the more reliable way to do this).

PR #1336, opened 2020-12-16
New NETCONF based IOS-XR driver 

ktbyers:


ktbyers:
get_facts() fails with the following exception:

```                                                                                                                                     
  File "iosxr_netconf.py", line 28, in <module>
    device_facts = device.get_facts()
  File "/home/kbyers/NAPALM/napalm/napalm/iosxr_netconf/iosxr_netconf.py", line 385, in get_facts
    basic_info_tree = facts_rpc_reply_etree.xpath(
IndexError: list index out of range
```

ktbyers:
FYI @neelimapp

neelimapp (approved):

111pontes (approved):
Merging is fine.  We already took a look at the issue with `get_facts` and the fix is simple.  We'll send a PR right after the merge.

111pontes:
Saw some changes related to `get_config (sanitized)` on https://github.com/napalm-automation/napalm/pull/1336/commits/dc530372a5cd51cc47dab0d46d1571c20b91ffb0. Note that #1307 (pending) takes care of the implementation for `get_config (sanitized)`.

ktbyers:
Okay, let me look at #1307. There is still more testing I want to do before I merge. I will create separate issues for things that I find like the `get_facts()` issue.

@111pontes @neelimapp Also what OS is the oldest IOS-XR OS version that we should reasonably expect to work (are designing to support)?

neelimapp:
Okay, Thanks, Kirk. This driver supports IOSXR 6.3.2 and above versions.

ktbyers:
@neelimapp Okay, great...thanks.

111pontes:
All getters and XML config are supported since 6.3.2.  Support for CLI config require 6.7.1 or later.

ktbyers:
Summary of test results on my `Cisco IOS XR Software, Version 6.5.3` test devices:

- get_facts()  FAILS
- get_arp_table() WORKS
- get_bgp_config() WORKS
- get_bgp_neighbors() WORKS
- get_bgp_neighbors_detail() WORKS
- get_config() FAILS (returns empty data for running/startup/candidate)
- get_config: CLI requires 6.7.1 support
- get_environment() FAILS (as it calls get_facts())
- get_firewall_policies Not implemented
- get_interfaces WORKS
- get_interfaces_counters WORKS
- get_interfaces_ip    WORKS
- get_ipv6_neighbors_table     Not implemented
- get_lldp_neighbors   WORKS (returns {})
- get_lldp_neighbors_detail    WORKS (returns {})
- get_mac_address_table        WORKS (returns [])
- get_network_instances        Not implemented
- get_ntp_peers                WORKS (returns {})
- get_ntp_servers              WORKS
- get_ntp_stats                WORKS
- get_optics                   Not implemented
- get_probes_config            WORKS (returns {})
- get_probes_results           WORKS (returns {})
- get_route_to                 WORKS
- get_snmp_information         WORKS
- get_users                    WORKS (returns empty password field though)
- get_vlans                    Not implemented
- is_alive                     WORKS
- ping                         Not implemented (?)
- traceroute                   WORKS

ktbyers:
@neelimapp Was there a reason the `ping` method was not implemented?

neelimapp:
@ktbyers Initially we implemented everything that is implemented in the existing IOS-XR driver, to maintain backward compatibility. Now we completed the implementation of ```ping```, ```get_ipv6_neighbors_table``` and working on the rest of the methods. We would like to commit them once this merge is done. We just wanted to get the lasted code and work on it. 
```get_optics```
```get_network_instances ```
```get_vlans```
```get_firewall_policies```

neelimapp:
get_users WORKS (returns empty password field though--is this a bug?) - Here we are retrieving password and not secret, to maintain backward compatibility with the existing IOS-XR driver.

ktbyers:
Sounds good on the other methods. 

`get_firewall_policies` should be deprecated so I wouldn't worry about implementing it. If you already did it...no worries.

ktbyers:
The next item I want to test is the configuration operations.

111pontes:
> get_users WORKS (returns empty password field though--is this a bug?) - Here we are retrieving password and not secret, to maintain backward compatibility with the existing IOS-XR driver.

@ktbyers, we brought it up in the past.  The device has the notion of a secret and a password.  The former has precedence over the latter.  We could set the password in the dict to the device secret if set on the device.  That'd provide the effective info to the user, but that'd diverge from how other Cisco drivers (`ios`, `iosxr`, `nxos`) are implemented.  We favored backward compatibility, even though I'd consider the behavior obsolete/inadequate.

ktbyers:
@111pontes What you did sounds like the right choice. I couldn't remember the specifics so I just flagged it when I was reviewing, but let's just leave it at what you did.

ktbyers:
@mirceaulinic This should be ready to go. As per previous communication, I think we should merge this into `develop` after the NAPALM 3.3.0 release.

ktbyers:
@mirceaulinic I think we should just go ahead and merge this (regardless of upcoming release) i.e. it is better to just keep moving forward with this and get feedback on it (so improvements can be made).

If you are okay with this, I will go ahead and merge it in.

mirceaulinic (approved):
:tada: :athletic_shoe: :package:

PR #1344, opened 2020-12-22
config merge operation for iOS devices not clear if rollback was successful

network-shark:
When the config merge operation fails napalm quits with  : "Configuration merge failed; automatic rollback attempted " But it does not tell you, if the rollback was successful or failed. Now it does .

I hope you find it useful.

### DIFF
--- a/napalm/ios/ios.py
+++ b/napalm/ios/ios.py
@@ -542,6 +542,12 @@ class IOSDriver(NetworkDriver):
                 raise ReplaceConfigException(msg)
         else:
             # Merge operation
+            filename = self.merge_cfg
+            cfg_file = self._gen_full_path(filename)
+            if not self._check_file_exists(cfg_file):
+                raise MergeConfigException("Merge source config file does not exist")
+            cmd = "copy {} running-config".format(cfg_file)
+            output = self._commit_handler(cmd)
             if "Invalid input detected" in output:
                 rollback = self.rollback()
                 if 'Rollback Done' in rollback:
@@ -552,8 +558,8 @@ class IOSDriver(NetworkDriver):
                             )
                     raise MergeConfigException(merge_error)
                 else:
-                    err_header = "Configuration merge failed; automatic
-                    rollback failed, user intervention required"
+                    err_header = """Configuration merge failed; automatic
+                    rollback failed, user intervention required"""
                     merge_error = "{0}:\n{1}{2}\n{3}".format(err_header,
                             output,rollback)
                     raise MergeConfigException(merge_error)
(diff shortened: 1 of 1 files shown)


napalm/ios/ios.py
```diff
@@ -581,14 +593,16 @@ def rollback(self):
...
+        output = self._commit_handler(cmd)
 
         # After a rollback - we no longer know whether this is configured or not.
         self.prompt_quiet_configured = None
 
         # Save config to startup
         self.device.save_config()
 
+        return output
+
```
ktbyers, line 605: Why was this changed to return `output` from _commit_handler? That didn't seem necessary for the PR?

ktbyers: @network-shark

mirceaulinic: Bump @network-shark

ktbyers: Okay, this was my misunderstanding...I don't think we can change this without changing the behavior of all the drivers, however (i.e. we would be changing what the `rollback` method returns).

Probably an easier solution is just to key off whether `rollback()` generates an exception or not (i.e. if `rollback` does not generate an exception then rollback worked...anything else should be a bug.

mirceaulinic: @network-shark did you want to reply to Kirk's suggestion above?

network-shark (commented):
```
if "Invalid input detected" in output:
              err_header = "Configuration merge failed; automatic rollback attempted"
              merge_error = "{0}:\n{1}".format(err_header, output)

              try:
                  self.rollback()
              except:
                  raise
              finally:
                  raise MergeConfigException(merge_error)
``` 
Like this ?

napalm/ios/ios.py
network-shark:
```
if "Invalid input detected" in output:
              err_header = "Configuration merge failed; automatic rollback attempted"
              merge_error = "{0}:\n{1}".format(err_header, output)

              try:
                  self.rollback()
              except:
                  raise
              else:
                  raise MergeConfigException(merge_error)
``` 
Like this ? @mirceaulinic  @ktbyers

ktbyers: Probably something more like:

```python
    def rollback(self):
        """Rollback configuration to filename or to self.rollback_cfg file."""
        filename = self.rollback_cfg
        cfg_file = self._gen_full_path(filename)
        if not self._check_file_exists(cfg_file):
            raise ReplaceConfigException("Rollback config file does not exist")
        cmd = "configure replace {} force".format(cfg_file)
        self._commit_handler(cmd)
+      output = self._commit_handler(cmd)

        # After a rollback - we no longer know whether this is configured or not.
        self.prompt_quiet_configured = None

+      if "Rollback Done" not in output:
+          raise SomeException(f"Automatic rollback to previous configuration failed:\n\n{output}")

        # Save config to startup
        self.device.save_config()
```

Issue #1069, opened 2019-10-13
naplam.get_route_to() not working on EOS:failed to detect a valid IP address from u'' 


netwninja:
### Description of Issue/Question

*Note*: Please check https://guides.github.com/features/mastering-markdown/
to see how to properly format your request.

### Did you follow the steps from https://github.com/napalm-automation/napalm#faq
(Place an ``x`` between the square brackets where applicable)

- [x] Yes
- [] No

### Setup

napalm-server-------vEOS

### napalm version
(Paste verbatim output from `pip freeze | grep napalm` between quotes below)

```
pip freeze | grep napalm
napalm==2.5.0
napalm-base==1.0.0
napalm-eos==0.6.1
napalm-ios==0.8.1
napalm-panos==0.5.2
```

### Network operating system version
(Paste verbatim output from `show version` - or equivalent - between quotes below)

```
vEOS-A#show version | in image
Software image version: 4.22.2F
```

### Steps to Reproduce the Issue
Use the below script 

```
# cat eos2.py 
import json
from napalm import get_network_driver
#optional_args = {'transport': 'https'}
driver = get_network_driver('eos')
dev = driver(hostname='veos-a', username='ansible', password='ansible123')
dev.open()
dev_info = dev.get_route_to(u'3.3.3.3/32')
dev.close()
print(json.dumps(dev_info, sort_keys=True, indent=4))
```

### Error Traceback
(Paste the complete traceback of the exception between quotes below)

```
Traceback (most recent call last):
  File "eos2.py", line 7, in <module>
    dev_info = dev.get_route_to(u'3.3.3.3/32')
  File "/usr/local/lib/python2.7/dist-packages/napalm/eos/eos.py", line 1258, in get_route_to
    .get("peerAddr", "")
  File "/usr/local/lib/python2.7/dist-packages/napalm/base/helpers.py", line 330, in ip
    addr_obj = IPAddress(addr)
  File "/usr/local/lib/python2.7/dist-packages/netaddr/ip/__init__.py", line 306, in __init__
    'address from %r' % addr)
netaddr.core.AddrFormatError: failed to detect a valid IP address from u''

```

### JSON output from Arista vEOS

```
vEOS-A#show ip route 3.3.3.3 | json
{
    "vrfs": {
        "default": {
            "routes": {
                "3.3.3.3/32": {
                    "kernelProgrammed": true,
                    "directlyConnected": false,
                    "routeAction": "forward",
                    "routeLeaked": false,
[...]

PR #1346, opened 2020-12-29
Fix for issue #1069

malanovo:
Fix for issue #1069 .  The problem was caused when an empty IP address was passed to one of the helpers due to it not being in the expected location.  For certain BGP related routes, peerEntry is moved up a level, so if it isn't found under routeDetail we catch the error thrown by the IP helper and try again in the other location.


napalm/eos/eos.py
```diff
@@ -1385,11 +1385,18 @@ def get_route_to(self, destination="", protocol="", longer=False):
...
-                            .get("peerEntry", {})
-                            .get("peerAddr", "")
-                        )
+                        try:
+                            remote_address = napalm.base.helpers.ip(
+                                bgp_route_details.get("routeDetail", {})
+                                .get("peerEntry", {})
+                                .get("peerAddr", "")
+                            )
+                        except AddrFormatError:
```
mirceaulinic, line 1394: Perhaps not a biggie, just curious:

> For certain BGP related routes, peerEntry is moved up a level, so if it isn't found under routeDetail we catch the error thrown by the IP helper and try again in the other location.

Does this apply to any specific route types, or any pattern we can distinguish at all? I'd be more inclined to a simpler `if 'routeDetail' in bgp_route_details` but perhaps there's some other context I'm missing.

Either way, thanks for this patch as well @malanovo and it'd be great to add a test case for this scenario here as well.

malanovo: Honestly I haven't quite figured out yet all the cases when EOS will return peerEntry in the alternate location.  The devices I have been testing against will trigger the issue for seemingly similar configuration sections, and I even have two different switches in the same network that look like they should both cause it, yet one does while the other does not.  I'm still trying to figure it out, since the values as returned through the CLI seem identical.  I'll circle back on it once the thing I'm building that uses this module is finished.

I fully agree with doing the simpler `if` based method, and originally started doing it that way, but went with the `try/except` method instead as it was easier to see what was going on, plus it seemed more consistent with existing code (I'm referring to how that method was used elsewhere in the code for catching other instances of AddrFormatError).

I'll work on getting a test case put together.  I should be able to capture something from a device in my network to use as a base for it.

malanovo:
Test case has been added to the pull request.

mirceaulinic (approved):

PR #1347, opened 2020-12-29
Set next hop to something valid for vxlan routes

malanovo:
This is a small change that stops Napalm from bombing out on EOS when attempting a get_route_to() on a VXLAN route, and inserts some values that make it useful.  The next hop is set to the IP of the destination VTEP, and the exit interface is set to vxlan1.  Granted we're assuming the interface name will be vxlan1, but at the moment, that's all it can be on EOS.  It might need alteration in the future if Arista changes that.


mirceaulinic (commented):
Looks good @malanovo, I think your changes make sense. Would you also be able to add a test case for this? Thanks.

malanovo:
Sure, I can work on that.  I've got several devices I can pull from for an example.

malanovo:
Test case has been added.  Please note that the test case relies on the fix for issue #1069 having been applied (which was just recently merged into the `develop` branch).

mirceaulinic (approved):
Looks good to me. Thanks @malanovo!

PR #1356, opened 2021-01-14
Fix for BGP description not found by get_bgp_neighbors in some EOS versions

malanovo:
Not all versions of EOS include descriptions in the output of `show ip bgp neighbors vrf all`, which is where `get_bgp_neighbors` is currently looking for it.  This patch adds some logic to pull the value out of the `show ip bgp summary vrf all` output, but retains the original method as well to avoid regressions.  The original code is basically screen scraping instead of relying on JSON output; I'm not sure why that is, but I didn't want to change it in case some older versions of EOS rely on that method.

This patch was definitely needed to get my 4.22.xF switches working, but oddly it wasn't needed for my 4.23.0M vEOS test box.  I have test cases, I'll attach those as soon as I open the pull request so I can get an issue number to keep the test case directory naming conventions intact.


malanovo:
Test case has been added.

mirceaulinic (approved):

mirceaulinic:
Thanks @malanovo

Issue #1359, opened 2021-01-14
key error when BGP peer has the same import/export policy names


lucasalvatore:
### Description of Issue/Question

When running `get_bgp_config()` on a junos device, if a peer as the same peer level import/export policies, a key error is thrown.

### Did you follow the steps from https://github.com/napalm-automation/napalm#faq
(Place an ``x`` between the square brackets where applicable)

- [x] Yes
- [] No

### napalm version
(Paste verbatim output from `pip freeze | grep napalm` between quotes below)

```
~ » pip freeze | grep napalm  
napalm==3.2.0
```

### Network operating system version
(Paste verbatim output from `show version` - or equivalent - between quotes below)

```
Model: qfx10002-72q
Junos: 19.4R2-S1.2
```

### Steps to Reproduce the Issue

have a bgp group configured like this:
```
luca@bsr1.ams1> show configuration protocols bgp group V6-INTERNAL-EDGE
type external;
import [ V6-INTERNAL-EDGE-IN DENY ];
family inet6 {
    unicast;
}
export [ V6-INTERNAL-EDGE-OUT DENY ];
peer-as 65530;
local-as 65520 
multipath;
neighbor x.x.x.x {
    description "switch1";
    }
}
neighbor y.y.y.y {
    description "switch2";
    import DENY;
    export DENY;
}
```

Then try to get the bgp config for that group:

```
def napalm_connect(device):
    username = getpass.getuser()
    driver = get_network_driver("junos")
    optional_args = {"use_keys": True, "allow_agent": True}
    router = driver(str(device), username, "", optional_args=optional_args)
    router.open()
    return router

router = napalm_connect('bsr1.ams1')
print(router.get_bgp_config(group='V6-INTERNAL-EDGE'))
```

### Error Traceback
```
In [214]: router = napalm_connect('bsr1.ams1')
     ...: print(router.get_bgp_config(group='V6-INTERNAL-EDGE'))
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-214-591df75972d8> in <module>
      1 router = napalm_connect('bsr1.ams1')
----> 2 print(router.get_bgp_config(group='V6-INTERNAL-EDGE'))

~/virtualenv/python3.8/lib/python3.8/site-packages/napalm/junos/junos.py in get_bgp_config(self, group, neighbor)
   1219                     if key in ["export_policy"]:
   1220                         # next-hop self is applied on export IBGP sessions
[...]

PR #1363, opened 2021-01-19
Use aply-groups inheritance on get-configuration RPC calls

mirceaulinic:
The behaviour for get-configuration RPC calls has been changed in
2af90cc75, in order to ensure we're gathering from the right database.

The way junos-eznc works is that if there aren't any specific options
passed in, it uses apply-groups by default. But when there are options
passed in, as we do starting with 2af90cc75, it doesn't default the
values of `inherit` and `groups`, and therefore when gathering the
config it won't expand the apply-groups, see
https://github.com/Juniper/py-junos-eznc/blob/ab0e8603eac56fa1d049695c7ef86eb64582ab8f/lib/jnpr/junos/factory/cfgtable.py#L517-L520.

Changing this now, by adding a few new optional arguments, which allow
the user to select the desired behaviour.

### DIFF
--- a/napalm/base/base.py
+++ b/napalm/base/base.py
@@ -1460,7 +1460,7 @@ class NetworkDriver(object):
             {
                 'error': 'unknown host 8.8.8.8.8'
             }
-            """
+        """
         raise NotImplementedError
 
     def get_users(self):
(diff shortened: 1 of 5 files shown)


lucasalvatore:
@mirceaulinic tested this against our devices which had the issue reported in https://github.com/napalm-automation/napalm/issues/1359 and they now work as expected!

PR #1375, opened 2021-02-13
Add cannonical mapping for FE, GE and 10GE interfaces on Cisco SMB switches

qaxi:
Cisco SMB switches (SF3xx, SF5xx, SG3xx, SG5xx) fw v. 1.x and 2.x are using short forms 'fe', 'ge', 'te'

In  napalm-automation-community/napalm-s350 I currently changing napalm's dictionary 'base_interfaces' like that
```
s350_base_interfaces = {**napalm.base.canonical_map.base_interfaces, 'fa':'FastEthernet', 'gi':'GigabitEthernet','te':'TengigabitEthernet'}
local_port = canonical_interface_name(local_port, s350_base_interfaces)
```

All three keys are there witch first letter uppercase with the same values, so this change should not break something.


mirceaulinic (approved):
Nice, thanks @qaxi!

PR #1377, opened 2021-02-14
Switch to GH Actions

ktbyers:
.


ktbyers:
Might break previous process on deployments so we would need to come up with a new process for that. It looks like previous deployment was bound to Travis.

Let me know what you want to do here. For the short-term, we could keep Travis only alive for deployment.

Let me know what you want to do here.

Depending on this, we should also disable the integration to Travis-CI

ktbyers:
I think we would also need to disable the Travis-CI integration in GitHub.

This also looks like it would break the `deployment` process since that was integrated into travis.

ktbyers:
Also need to fix coveralls integration.

Probably should move forward with this PR and then fix deployment process and coveralls separately.

ktbyers:
@mirceaulinic I think this is probably good to go:

The following issues are still outstanding:
1. What is the new deployment process. Previous deployment process looked like it was tied into Travis-CI.
2. I might have to fix coveralls/coverage.io process. I don't understand the current integration here so a bit unclear to me whether this will work after we switch-over or break (but I should be able to fix this).
3. We will need to remove the Travis-CI integration in GitHub (after this is merged)

ktbyers:
@mirceaulinic Are you good with this change? Anything else we need to do here before we merge it?

mirceaulinic (approved):
Already reviewed in #1379, but thanks again for this, Kirk!

PR #1382, opened 2021-02-23
resolve #1293

### LINKED ISSUE #1293: JunOS: Add Support for Logical Interfaces Counters

I noticed that when calling `get_interfaces_counters` for JunOS, only the physical interface counters are returned and not the logical interfaces. Since logical interfaces are returned in a call to `get_interfaces`, I was wondering if it makes sense to also include logical interfaces when calling `get_interfaces_counters`. 

I did some testing and it's definitely possible to do, however at least in my environment, only the `rx_octets` and `tx_octets` have values. Still, they're values nonetheless and it would be nice to access them.

(Place an ``x`` between the square brackets where applicable)

- [x] Yes
- [ ] No

### napalm version
(Paste verbatim output from `pip freeze | grep napalm` between quotes below)

```
napalm==3.2.0
```

(Paste verbatim output from `show version` - or equivalent - between quotes below)

```
JunOS 17.3
```

minefuto:
This is related to #1293.
I added support for logical interface counters of Junos.


napalm/junos/junos.py
```diff
@@ -403,11 +403,18 @@ def get_interfaces_counters(self):
         """Return interfaces counters."""
         query = junos_views.junos_iface_counter_table(self.device)
         query.get()
+        query_logical = junos_views.junos_logical_iface_counter_table(self.device)
```
mirceaulinic, line 406: I don't think we should make a separate request just for this: the logical interfaces are already present in the `junos_iface_counter_table`, we just need to extract them. For this we can use the nested tables junos-eznc offers, see an example [here](https://github.com/napalm-automation/napalm/blob/develop/napalm/junos/utils/junos_views.yml#L64-L75) where a view is able to reference a  sub-table. In the same way, we can have the logical interfaces nested under a sub-table, and avoid an additional RPC request.

mirceaulinic (changes requested):
Good start @minefuto, thanks for having a go at this. I've requested some changes, as I wouldn't want to increase the complexity while the data is already available.

minefuto:
Thanks, @mirceaulinic.
I think I was able to implement it as per your requests.
Sorry, but I have no confidence because I'm not an expert python/junos-eznc.
Could you please review it again?

mirceaulinic (approved):
Hey @minefuto - thanks for making the changes. It is slightly different from what I had in mind, but I like this implementation too. :smile:

PR #1393, opened 2021-03-08
Fix typo in mock.py

Kircheneer:
Fix discard_config counting up commit_config calls instead of discard_config.


mirceaulinic (approved):
:+1:

PR #1395, opened 2021-03-10
Preparation for type hinting

Kircheneer:
As discussed shortly with @mirceaulinic on Slack I have started adding some rudimentary [PEP484](https://www.python.org/dev/peps/pep-0484/) support.

What I did in this PR:
- Created `mypy.ini` with a baseline configuration
  - Ignore any missing imports from dependencies that don't type hint (yet)
  - Ignore type errors from some tests
- Added mypy to `requirements-dev.txt`
- Change the models in `napalm/base/test/models.py` to TypedDict instances
  - May also be used in type checking
  - Fixed test by replacing the `model` parameter for the model testing functions with `model.__annotations__`, which contains the schema that used to be in the model dictionaries. During this I noticed that the code for `TestGettersNetworkDriver._test_model` in `napalm/base/test/base.py` and `test_model` in `napalm/base/test/helpers.py` is duplicated, but I didn't want to pollute this PR with that.

Next steps going forward would be:
- Extend `napalm/base/test/models.py` so that it contains anything a `NetworkDriver` instance might return from a method
- Typecheck the `NetworkDriver` class and their children (otherwise the tests start failing due to differences in the signatures, this might also be something that any community driver has to take into consideration)
- Then at some point add type hints everywhere else and add mypy to the GitHub actions pipeline (if that aligns with the projects vision)


mirceaulinic (approved):
LGTM @Kircheneer. Was there anything else you wanted to push to this branch, or can go ahead and merge?

Kircheneer:
That's alright for now, I will submit another PR at some point.

mirceaulinic:
Great work @Kircheneer, thanks!

PR #1367, opened 2021-01-25
VLAN table information for JunOS

ckozler:

PR #1398, opened 2021-03-19
Add get_vlans for Junos

minefuto:
I add get_vlans for Junos because #1367 is inactivity.


napalm/junos/junos.py
```diff
@@ -2362,3 +2362,36 @@ def get_network_instances(self, name=""):
...
+            vlan = junos_views.junos_vlans_table_switch_l2ng(self.device)
+        elif switch_style == "BRIDGE_DOMAIN":
+            vlan = junos_views.junos_vlans_table(self.device)
+        elif switch_style == "VLAN":
+            vlan = junos_views.junos_vlans_table_switch(self.device)
+        else:  # switch_style == "NONE"
+            return result
+
+        vlan.get()
+        unmatch_pattern = "l2rtb-interface-name|None|l2ng-l2rtb-vlan-member-interface"
```
mirceaulinic, line 2379: I suppose there are different formats to match because of various switch styles? Is `None` a possibility? If so, would you have access to a device to add another test case for that one too?

mirceaulinic (commented):
That's great @minefuto, thanks for taking over. Looks good, just a small detail I'm curious about otherwise this should be good.

napalm/junos/junos.py
minefuto: It returns different values depending on switch_style If there is not binding IF of vlan.
BRIDGE_DOMAIN: l2rtb-interface-name
VLAN: None(string)
VLAN_L2NG: l2ng-l2rtb-vlan-member-interface

I think I have already added the above 3 patterns as a VLAN300 per switch_style.
[BRIDGE_DOMAIN](https://github.com/minefuto/napalm/blob/09fe92f0dfaffdcad003266bad057bc9d109bc85/test/junos/mocked_data/test_get_vlans/bridge_domain/get-bridge-instance-information.xml#L17-L22), [VLAN](https://github.com/minefuto/napalm/blob/09fe92f0dfaffdcad003266bad057bc9d109bc85/test/junos/mocked_data/test_get_vlans/vlan/get-vlan-information.xml#L51-L69), [VLAN_L2NG](https://github.com/minefuto/napalm/blob/09fe92f0dfaffdcad003266bad057bc9d109bc85/test/junos/mocked_data/test_get_vlans/vlan_l2ng/get-vlan-information.xml#L25-L32)

Also, I found an exception on MX(switch_style:BRIDGE_DOMAIN).
It seems it's possible that MX returns None(NoneType) If there is not binding IF of vlan.
I added this pattern only BRIDGE_DOMAIN by f519d24, 09fe92f because I could not found this exception in the others switch_style.

mirceaulinic: Makes sense, thank you @minefuto!

mirceaulinic (approved):

Issue #1410, opened 2021-03-25
Control Plane Interface on 6509 "show interfaces" throws ValueError


rnwolfe:
### Description of Issue/Question

Certain IOS Platforms have begun showing the control plane as an interface in `show interfaces` output and the NAPALM IOS Driver doesn't know how to handle this interface causing an exit as it doesn't match the existing interface patterns in the IOS driver. 

This is news to me and the first time I've ever seen it before on a switch. In this specific case it is a **WS-6509-E** running **Version 12.2(18)SXD4**. Haven't come across it otherwise.

Here's the problem interface:

```
Control Plane Interface is down, line protocol is down
  Hardware is Control-Plane, address is 0000.0000.0000 (bia 0000.0000.0000)
  MTU 0 bytes, BW 10000000 Kbit, DLY 0 usec,
     reliability 0/255, txload 1/255, rxload 1/255
  Encapsulation UNKNOWN, loopback not set
  ARP type: ARPA, ARP Timeout 00:00:00
  Last input never, output never, output hang never
  Last clearing of "show interface" counters never
  Input queue: 0/75/0/0 (size/max/drops/flushes); Total output drops: 0
  5 minute input rate 0 bits/sec, 0 packets/sec
  5 minute output rate 0 bits/sec, 0 packets/sec
  L2 Switched: ucast: 0 pkt, 0 bytes - mcast: 0 pkt, 0 bytes
  L3 in Switched: ucast: 0 pkt, 0 bytes - mcast: 0 pkt, 0 bytes mcast
  L3 out Switched: ucast: 0 pkt, 0 bytes mcast: 0 pkt, 0 bytes
     0 packets input, 0 bytes, 0 no buffer
     Received 0 broadcasts (0 IP multicast)
     0 runts, 0 giants, 0 throttles
     0 input errors, 0 CRC, 0 frame, 0 overrun, 0 ignored
     0 packets output, 0 bytes, 0 underruns
     0 output errors, 0 interface resets
     0 output buffer failures, 0 output buffers swapped out
```

and the platform it showed up on:

### Did you follow the steps from https://github.com/napalm-automation/napalm#faq

- [x] Yes
- [ ] No

### napalm version
(Paste verbatim output from `pip freeze | grep napalm` between quotes below)

```
$ pip freeze | grep napalm
napalm==3.2.0
nornir-napalm==0.1.1
```

### Network operating system version
(Paste verbatim output from `show version` - or equivalent - between quotes below)

```
[...]

PR #1411, opened 2021-03-25
Handle control plane interface

rnwolfe:
See #1410 

I approached the fix by simply adding an additional regex pattern for the interface line. There are multiple approaches that are a matter of opinion, e.g. just modify existing pattern with an or to include the text but I chose this route. 

Also a matter of opinion is whether to handle this case and return the data or just skip it and exclude from the response as it seems to just be a bogus interface with no MAC, MTU, etc; however, I assume let the user handle that decision and just return the data as it is on the device which I think is inline with NAPALM principles.

Tests should also work to cover this case.

### DIFF
--- a/napalm/ios/ios.py
+++ b/napalm/ios/ios.py
@@ -1088,7 +1088,10 @@ class IOSDriver(NetworkDriver):
 
             interface_regex_1 = r"^(\S+?)\s+is\s+(.+?),\s+line\s+protocol\s+is\s+(\S+)"
             interface_regex_2 = r"^(\S+)\s+is\s+(up|down)"
-            interface_regex_3 = r"^(Control Plane Interface)\s+is\s+(.+?),\s+line\s+protocol\s+is\s+(\S+)"
+            interface_regex_3 = (
+                r"^(Control Plane Interface)"
+                r"\s+is\s+(.+?),\s+line\s+protocol\s+is\s+(\S+)"
+            )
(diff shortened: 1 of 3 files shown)


napalm/ios/ios.py
```diff
@@ -1088,7 +1088,8 @@ def get_interfaces(self):
 
             interface_regex_1 = r"^(\S+?)\s+is\s+(.+?),\s+line\s+protocol\s+is\s+(\S+)"
             interface_regex_2 = r"^(\S+)\s+is\s+(up|down)"
-            for pattern in (interface_regex_1, interface_regex_2):
+            interface_regex_3 = r"^(Control Plane Interface)\s+is\s+(.+?),\s+line\s+protocol\s+is\s+(\S+)"
```
mirceaulinic, line 1091: The lint check is failing because of this long line.
```suggestion
            interface_regex_3 = r"^(Control Plane Interface)\s+is\s+(.+?),\s+line\s+protocol\s+is\s+(\S+)"
            interface_regex_3 = (
                r"^(Control Plane Interface)"
                r"\s+is\s+(.+?),\s+line\s+protocol\s+is\s+(\S+)"
            )
```

mirceaulinic (changes requested):
Looks good, just a small lint issue required.

rnwolfe:
Thanks, @mirceaulinic I was wondering what was failing the CI. Looking better now.

mirceaulinic (approved):

Issue #1310, opened 2020-10-23
IOS-XR get_route_to empty outgoing interface


Tristou27:
### Description of Issue/Question

In the IOSXRDriver, the method get_route_to does not check for the outgoingInterface, set by default to an empty string.
On IOS XR, there is a value in the XML output named InterfaceName, located under RoutePath/InterfaceName
```
<RoutePath>
    <Entry>
        <Address>0.0.0.0</Address>
        <InformationSource>0.0.0.0</InformationSource>
        <InterfaceName>GigabitEthernet0/1/1/2</InterfaceName>
(...)
    </Entry>
</RoutePath>
```

Would it be possible to complete this information ? I can open a PR to speed up the addition.

Rgds
Tristou27

### Did you follow the steps from https://github.com/napalm-automation/napalm#faq
(Place an ``x`` between the square brackets where applicable)

- [x] Yes
- [] No

### napalm version
(Paste verbatim output from `pip freeze | grep napalm` between quotes below)

```
(virtenv) {20-10-23 15:28}nietb4:~/Core-Tests-Automation ttr% pip freeze | grep napalm
napalm==3.2.0
```

### Network operating system version
(Paste verbatim output from `show version` - or equivalent - between quotes below)

All

### Steps to Reproduce the Issue
1 / Open a connection to a XR device
2 / Ask for a route in RIB (no matter the route)
3 / In the return dictionary, outgoing Interface is always empty.

### Error Traceback
(Paste the complete traceback of the exception between quotes below)

```
Original call was: <?xml version="1.0" encoding="UTF-8"?><Request MajorVersion="1" MinorVersion="0"><Get><Operational><RIB><VRFTable><VRF><Naming><VRFName>default</VRFName></Naming><AFTable><AF><Naming><AFName>IPv4</AFName></Naming><SAFTable><SAF><Naming><SAFName>Unicast</SAFName></Naming><IP_RIBRouteTable><IP_RIBRoute><Naming><RouteTableName>default</RouteTableName></Naming><RouteTable><Route><Naming><Address>172.18.0.35</Address></Naming></Route></RouteTable></IP_RIBRoute></IP_RIBRouteTable></SAF></SAFTable></AF></AFTable></VRF></VRFTable></RIB></Operational></Get></Request>
[...]

PR #1421, opened 2021-04-09
add outgoing_interface in get_route_to for isoxr

Tristou27:
This PR solves #1310. 

It is just a quick change in order to have the correct outgoing_interface when asking for a route to an IOSXR device.
It should work for all XR versions as long as the XML output does not change (which has not been the case since XR 4).

Hope this helps.


mirceaulinic (approved):
Nice, thanks @Tristou27 !

PR #1433, opened 2021-04-30
Redo of type hints PR

ktbyers:
I had to redo this PR so that we could get the IOS-XR NETCONF pull-request in before this PR (to avoid the merge conflicts in the IOS-XR PR).

PR #1434, opened 2021-04-30
Type hints for napalm.base and napalm.nxos

Kircheneer:
Supersede #1433, #1412 taking into account the newly added `napalm.iosxr_netconf`.


Kircheneer:
As per the NTC slack discussion I could look into utilizing Pydantic for the models. That's probably the proper way to do it anyway. What do you think @ktbyers?

Kircheneer:
Here are some benefits I see:
- Run time schema enforcement with useful error messages for driver implementation
- General schema availability for 3rd party apps using NAPALM (say for example Netbox/Nautobot might be able to gain something from having the actual schemas available)
- Nice IDE support (I've worked with FastAPI in the past and the auto completion and type support is quite useful IMO)

ktbyers:
@Kircheneer Can you open that as a separate Github discussion item? i.e. better to separate it from this PR and then discuss it there.

Kircheneer:
Raise as #1441 - we should delay this PR until that discussion is done.

ktbyers:
I think we should just move forward with this and not hold off for discussions on #1441 (i.e. once conflicts are fixed, then we should just merge this PR).

Kircheneer:
Alright. I will get the checks passing later and let you know.

Kircheneer:
I must admit that I don't have a clue as to what exactly is failing here. JunOS tests are reporting a missing `transform` function... Any clues?

Kircheneer:
Re-iterated upon in #1476. Something weird happened with this one, I don't know. The failing tests are running properly in that PR. Sorry for the mess!

PR #1455, opened 2021-05-26
Add source_interface argument to ping

chadell:
Some vendors allow the definition of a `source interface` instead of the `source address` for the ping command with different keyword. For instance, [Junos](https://www.juniper.net/documentation/us/en/software/junos/icmp/topics/ref/command/ping.html).
Others don't use different keyword, so both arguments (`source` and `source_interface`) will end up using the same one.

### DIFF
--- a/napalm/base/base.py
+++ b/napalm/base/base.py
@@ -1273,6 +1273,7 @@ class NetworkDriver(object):
         size=c.PING_SIZE,
         count=c.PING_COUNT,
         vrf=c.PING_VRF,
+        source_interface=c.PING_SOURCE_INTERFACE,
     ):
         """
         Executes ping on the device and returns a dictionary with the result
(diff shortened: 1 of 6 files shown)


mirceaulinic (commented):
Looks good. Any chance you could add the argument for the remaining platforms? (Even without actually implementing, just logging a warning when the argument is being used it would be just fine - see for example like this: https://github.com/napalm-automation/napalm/blob/develop/napalm/junos/junos.py#L1548-L1550)

chadell:
> Looks good. Any chance you could add the argument for the remaining platforms? (Even without actually implementing, just logging a warning when the argument is being used it would be just fine - see for example like this: https://github.com/napalm-automation/napalm/blob/develop/napalm/junos/junos.py#L1548-L1550)

When you say "other platforms" which ones are you thinking on?
Maybe I have skipped some but I tried to add support to all the `ping` methods in the repository, where `junos` and `eos` have a differentiated argument for the `source_interface` case and for `ios` and `nxos` it overloads the already available `source` argument as it can take or a `source_ipaddress` or `source_interface`.

mirceaulinic (approved):
You're right @chadell, I assumed the XR drivers implement `ping` too... Thanks!

PR #1470, opened 2021-06-18
Junos get_lldp_neighbor port returns the port-id, not the port-decription

chadell:
While working with `get_lldp_neighbors` for `junos` I noticed that the return value for `port` was misleading (at least for me), being actually the `port-description` instead the expected `port-id`.

I understand this could be a breaking change, so opening this PR to get feedback on this proposed change.


chadell:
closing this, as I will use the key as the port-id, but maybe the field should ideally be renamed to `description` :)

napalm/junos/utils/junos_views.yml
```diff
@@ -104,7 +104,7 @@ junos_lldp_table:
 junos_lldp_view:
   fields:
     hostname: lldp-remote-system-name
-    port: lldp-remote-port-description | lldp-remote-port-id
```
mirceaulinic, line 107: `lldp-remote-port-description` needs to stay as that's what older Junos versions provide.

mirceaulinic (changes requested):

napalm/junos/utils/junos_views.yml
```diff
@@ -104,7 +104,7 @@ junos_lldp_table:
...
+    port: lldp-remote-port-id
```
mirceaulinic, line 107:
```suggestion
    port: lldp-remote-port-id | lldp-remote-port-description
```

mirceaulinic:
I left a suggested change, in order to avoid breaking the behaviour for older Junos. I'll also ask you to provide an additional test case (or rename the existing on to something like `old_junos` and provide a new one containing the `<lldp-remote-port-id>27</lldp-remote-port-id>` line).

napalm/junos/utils/junos_views.yml
chadell: thanks for your point (I was not aware of this)

chadell:
@mirceaulinic , after reviewing the logic, I decided to implement a small change on `get_lldp_neighbors`, in order to get only the first element of a list, as it's actually how I understand the `|` operator.
Following your recommendation, I created an old version (without the `<lldp-remote-port-id>` tag), and kept the normal one, with both tags.

PR #1476, opened 2021-07-25
Type hinting

Kircheneer:
Something weird happened with #1434 so I redid the PR - again. This should be mostly free of any non-sense commits and should reflect a proper and useful implementation of type hinting for the NXOS parts of NAPALM.

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -35,7 +35,7 @@ jobs:
 
       - name: Run type checker
         run: |
-          mypy -p napalm || true
+          mypy -p napalm --config-file mypy.ini || true
 
       - name: Run Tests
         run: |
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -35,7 +35,7 @@ jobs:
 
       - name: Run type checker
         run: |
-          mypy -p napalm --config-file mypy.ini || true
+          mypy -p napalm --config-file mypy.ini
 
       - name: Run Tests
         run: |
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -1,4 +1,4 @@
---- 
+---
 name: build
 on: [push, pull_request]
 jobs:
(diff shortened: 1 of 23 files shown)


Kircheneer:
Anything else I can do here to help get this merged?

ktbyers:
Yeah, don't hold this up for me...so if it is good with @mirceaulinic (then it is fine with me).

mirceaulinic:
Hi @Kircheneer. Yes, just let me release version 3.4.0 (hopefully next week), then we can merge this and start preparing release 4.0.0. Sorry for the wait!

Kircheneer:
Thanks @mirceaulinic. Just let me know whenever we are ready for merging, then I'll clean up the conflicting files again.

Kircheneer:
Left with the following errors:

```
napalm/base/helpers.py:305: error: Incompatible default for argument "default" (default has type "str", argument has type "R")
napalm/base/mock.py:201: error: Module has no attribute "CommitError"
```

For the first one: The docs state explicitly that a `str` will always be returned, even though in general I though this would be a generic function. Any clues as to what degree this is true?

For the second one: I'm unsure how this got introduced - any pointers?

Kircheneer:
Regarding the `CommitError` thing, I'm thinking that this is probably a bug Mypy uncovered. I just moved the `raise napalm.CommitError("Pending commit confirm already in process!")` line to the front of the commit_config method:
```python
    def commit_config(self, message: str = "", revert_in: Optional[int] = None) -> None:

        raise napalm.CommitError("Pending commit confirm already in process!")
        count = self._count_calls("commit_config")
        self._raise_if_closed()
        if revert_in is not None:
            if self.has_pending_commit():
                raise napalm.CommitError("Pending commit confirm already in process!")
            else:
                self._pending_commits = True
        self.merge = None
        self.filename = None
        self.config = None
        mocked_data(self.path, "commit_config", count)
```

This lead to the following unit test errors:
```
FAILED test/base/test_mock_driver.py::TestMockDriver::test_configuration_merge - AttributeError: module 'napalm' has no attribute 'CommitError'
FAILED test/base/test_mock_driver.py::TestMockDriver::test_configuration_replace - AttributeError: module 'napalm' has no attribute 'CommitError'
FAILED test/base/test_mock_driver.py::TestMockDriver::test_configuration_replace_confirm - AttributeError: module 'napalm' has no attribute 'CommitError'
```

I'll put in a fix for that into this branch if you don't mind.

About the other issue I think we need to have a synchronous discussion because that touches lots of places at once.

mirceaulinic (approved):
Awesome work, @Kircheneer! 🙇🏼

Issue #1459, opened 2021-06-01
Cisco IOS get_interfaces retuns wrong bandwith


network-shark:
### Description of Issue/Question

get_interfaces func returns the speed in megabit/s 

configuration
```
interface gi0/0
bandwith 2500  
```

napalm return
```
 {'speed': 2 }
```
expected return
```
 {'speed': 2.5 }
```

```
router#show interfaces gigabitEthernet 0/0
GigabitEthernet0/0 is up, line protocol is up
  Hardware is CN Gigabit Ethernet, address is 0000.0000.0000 (bia 0000.0000.0000)
  Description: Internet 
  Internet address is 192.168.1.1/30
  MTU 1500 bytes, BW 2500 Kbit/sec, DLY 100 usec
```

### Steps to Reproduce the Issue
run get_interfaces against a router 

### 

Could this be improved without breaking the existing code ?

mirceaulinic:
Yes, I think this should be corrected.

network-shark:
@mirceaulinic I think I can create a PR for this , just for clarification. We have to change interface speed to float in every os , right ? iOS , junnos ....

network-shark:
Hmm , 

I tried to implement a PR , but after my first change I got .


```test/ios/test_getters.py::TestGetter::test_get_interfaces[alternate] key: speed
model_class: <class 'int'>
data_class: <class 'float'>
key: mac_address
model_class: <class 'str'>
data_class: <class 'str'>
FAILED
test/ios/test_getters.py::TestGetter::test_get_interfaces[alternate2] key: speed
model_class: <class 'int'>
data_class: <class 'float'>
key: mac_address
model_class: <class 'str'>
data_class: <class 'str'>
FAILED
test/ios/test_getters.py::TestGetter::test_get_interfaces[control-plane-interface] key: speed
model_class: <class 'int'>
data_class: <class 'float'>
key: mac_address
model_class: <class 'str'>
data_class: <class 'str'>
FAILED
test/ios/test_getters.py::TestGetter::test_get_interfaces[pseudowire] key: speed
model_class: <class 'int'>
data_class: <class 'float'>
key: mac_address
model_class: <class 'str'>
data_class: <class 'str'>
FAILED
test/ios/test_getters.py::TestGetter::test_get_interfaces[normal] key: speed
model_class: <class 'int'>
data_class: <class 'float'>
key: mac_address
model_class: <class 'str'>
data_class: <class 'str'>
FAILED

```

PR #1478, opened 2021-08-01
Fixing #1459 bandwith parameter now returns float

network-shark:
https://github.com/napalm-automation/napalm/issues/1459


mirceaulinic (approved):
Nice work @network-shark - thanks!

Issue #1489, opened 2021-09-02
nxos_ssh - get_vlans returns 16777216 as the vlan id.


thomasbridge74:
### Description of Issue/Question

I just used the get_vlans device using the nxos_ssh driver - to get a dict back with a keys that don't use the VLAN id, but have strings such as '16777216'.

### Did you follow the steps from https://github.com/napalm-automation/napalm#faq
(Place an ``x`` between the square brackets where applicable)

- [X] Yes
- [] No

### napalm version

```
napalm-3.3.1-py2.py3-none-any.whl
```

### Network operating system version
(Paste verbatim output from `show version` - or equivalent - between quotes below)

```
DEFRA1SWRDFA01A# sh ver
Cisco Nexus Operating System (NX-OS) Software
TAC support: http://www.cisco.com/tac
Documents: http://www.cisco.com/en/US/products/ps9372/tsd_products_support_series_home.html
Copyright (c) 2002-2016, Cisco Systems, Inc. All rights reserved.
The copyrights to certain works contained in this software are
owned by other third parties and used and distributed under
license. Certain components of this software are licensed under
the GNU General Public License (GPL) version 2.0 or the GNU
Lesser General Public License (LGPL) Version 2.1. A copy of each
such license is available at
http://www.opensource.org/licenses/gpl-2.0.php and
http://www.opensource.org/licenses/lgpl-2.1.php

Software
  BIOS:      version 3.1.0
  kickstart: version 7.3(1)D1(1)
  system:    version 7.3(1)D1(1)
  BIOS compile time:       02/27/2013
  kickstart image file is: bootflash:///n7700-s2-kickstart.7.3.1.D1.1.bin
  kickstart compile time:  9/9/2016 23:00:00 [09/10/2016 02:54:38]
  system image file is:    bootflash:///n7700-s2-dk9.7.3.1.D1.1.bin
  system compile time:     9/9/2016 23:00:00 [09/10/2016 04:28:12]
```

### Steps to Reproduce the Issue

1.  Connect to device with nxos_ssh driver.
2. Call `device.get_vlans()`
3. Get unexpected feedback.

### Error Traceback

Problem appears to be based that napalm is running `show vlan brief | json` and the output has changed.    The version of NXOS on the box being used is giving this as part of the response (trimmed for brevity)
```
      {
        "vlanshowbr-vlanid": "3507421184", 
        "vlanshowbr-vlanid-utf": "4049",
        ....
      }
```

[...]

PR #1490, opened 2021-09-02
Fixing the issue with the updated json output in newer versions of NXOS.

thomasbridge74:
This addresses the issue found in #1489 .


mirceaulinic (approved):
Code looks good. Could you also add a test case for this issue please?

thomasbridge74:
@mirceaulinic I've added two tests as requested.    One covers the scenario where the `vlanshowbr-vlanid` string is not the same as the `vlanshowbr-vlanid-utf` string and one covers a scenario where the `vlanshowbr-vlanid-utf` is not received (which may be the case in older versions of Arista, I'm not sure).

PR #1494, opened 2021-09-10
Fix pyIOSXR handling of devices with XML agent issues

dkaplan1:
This PR addresses two issues causing `_send_command` to potentially infinitely recurse in cases where the XML agent on a device is having issues and locking is turned on (the default).

Issues explained
1) When `_send_command` encounters an `IOError` and the criteria is met for an XML agent issue, it calls `_enter_xml_mode` without checking whether the timeout has expired https://github.com/napalm-automation/napalm/blob/68a7c70f3c928db40749bf76d10944bd347b7f43/napalm/pyIOSXR/iosxr.py#L318-L342
`enter_xml_mode` may call `lock` depending on whether locking is turned on (the default) https://github.com/napalm-automation/napalm/blob/68a7c70f3c928db40749bf76d10944bd347b7f43/napalm/pyIOSXR/iosxr.py#L274-L275 `lock` will call `_execute_rpc` if it's not already locked https://github.com/napalm-automation/napalm/blob/68a7c70f3c928db40749bf76d10944bd347b7f43/napalm/pyIOSXR/iosxr.py#L532-L535 
2) `_send_command` is called without a start arg and without checking whether timeout specified has been exceeded (it no longer has access to the original start time). https://github.com/napalm-automation/napalm/blob/68a7c70f3c928db40749bf76d10944bd347b7f43/napalm/pyIOSXR/iosxr.py#L411-L419 since start isn't set, it's reassigned in `_send_command ` to the current time and the cycle continues https://github.com/napalm-automation/napalm/blob/68a7c70f3c928db40749bf76d10944bd347b7f43/napalm/pyIOSXR/iosxr.py#L296-L297

To fix the issues noted above:
1) `_send_command` checks `_timeout_exceeded` prior to calling `_enter_xml_mode`
2) `start_time` is now an `IOSXR` instance attribute that is set on first entrance to `_execute_rpc` and unset on exit
3) All `start` references/args have been replaced with `start_time`

Note - There was once instance I found where `_timeout_exceeded` is called outside the context of `_execute_rpc` https://github.com/napalm-automation/napalm/blob/68a7c70f3c928db40749bf76d10944bd347b7f43/napalm/pyIOSXR/iosxr.py#L204 however `start` wasn't passed prior here, so this shouldn't change any existing behavior here.


dkaplan1:
@ktbyers thanks for merging. I saw that tests were failing on black formatting, so I added a commit fixing that to this branch, hopefully I did that right :)

Issue #1014, opened 2019-07-06
New driver: eos_ssh for Arista devices without API


TheRealBecks:
For an internal project I needed to create eos_ssh as a new driver as we have the EOS API not licensed (yet) and also not configured on our Arista devices. eos_ssh is a work in progress and I really like to contribute this driver to NAPALM, but first I wanted to know if you would integrate that driver?

It's not yet finished, but working great so far. If I get a GO from you I will work further on that project. So what do you think about that driver? :)

mirceaulinic:
Hey @TheRealBecks. I'd give that a :+1: as I'm well aware of the challenge with the eAPI licensing. I'm not sure what approach you've been working on, but I'd suggest the following to minimize the amount of code we're maintaining (and, implicitly, your work :wink:) while re-using the existing code.
The core idea is that from the CLI you can get the JSON output you'd normally get over the eAPI, by simply adding `| json` to the usual command. AFAIK, the support for this is going sufficiently back in time for this.

In brief:

- Have another choice for the `transport` optional argument, `ssh` which users can select, yet `https` should always remain the default value.
- Define an internal helper method that depending on the transport used, invokes either `self.device.run_commands(` or adds `| json` to the command and loads the JSON document, e.g.,

```python
def _run_commands(self, commands):
    if self.transport == 'ssh':
        ret = []
        for command in commands:
            cmd_pipe = command + ' | json'
            cmd_txt = self._netmiko_device.send_command(cmd_pipe)
            cmd_json = json.loads(cmd_txt)
            ret.append(cmd_json)
        return ret
    else:
        return self.device.run_commands(commands)
```

Something along these lines. the `self._netmiko_object` would be initialized during `open()` when the transport of choice is `ssh`.

- In every of the existing methods, instead of executing `self.device.run_commands`, you'd invoke that newly defined method, e.g., this line: https://github.com/napalm-automation/napalm/blob/develop/napalm/eos/eos.py#L326 would become:

```python
result = self._run_commands(commands)
```

I think that'd work well. Let me know if you have any questions or other thoughts!

ktbyers:
I would be okay with doing this as long as we do this in the way Mirceau describes. In other words, there is no parsing and everything is done via `| json` and all the commands/mechanisms are essentially or very close to essentially the same.

Or worded differently, we would basically be saying that SSH is just another transport and that we won't be doing any screen-scraping in the driver.

Sound reasonable to everyone?

thomasbridge74:
[...]

PR #1502, opened 2021-10-03
arista ssh access.

thomasbridge74:
As discussed in https://github.com/napalm-automation/napalm/issues/1014 I'm starting work on allowing the EOS module to use ssh as a transport rather than HTTP/HTTPS.    This does necessitate some additional if statements within the code base.

So far I've managed to get this working to the extent we can connect to an Arista device using the ssh as the underlying transport.   I've updated a number of getter functions (not all yet) as well as the is_alive() function to cover the case where we're using ssh as the transport.   This PR is very much a WIP and (in my view) is not ready for committing to the develop branch but given there's still an amount of effort involved (the rest of the getters, ping and traceroutes, not to mention the configuration support options I've not even started to look at), I wanted to submit something for feedback!


thomasbridge74:
I've now updated as per https://github.com/napalm-automation/napalm/pull/1502/commits/2ff4e212139b8d20f3083e8192b536fabcc30516 the get functions that do not try to parse the `show running-config` output.

This still excludes the following functions on Arista that use them:
- "get_bgp_config"
- "get_bgp_neighbors",
- "get_bgp_neighbors_detail",
- "get_config",
- "get_ntp_severs",
- "get_snmp_information",

I also need to look at the following functions for further testing:
- "get_route_to",
- "get_users"

In `get_users` it appears to be that the newer version of vEOS I'm testing against (4.26.1F) does support JSON in the output for the command but is using a newer command syntax.

napalm/eos/eos.py
```diff
@@ -131,58 +135,119 @@ def _process_optional_args(self, optional_args):
...
+                return {"is_alive": self.device.remote_conn.transport.is_active()}
+            except (socket.error, EOFError):
+                # If unable to send, we can tell for sure that the connection is unusable
+                return {"is_alive": False}
+
         return {"is_alive": True}  # always true as eAPI is HTTP-based
 
+    def _run_commands(self, commands, **kwargs):
+        if self.transport == "ssh":
+            fn0039_transform = kwargs.pop("fn0039_transform", True)
```
mirceaulinic, line 225: We already have an optional argument, `eos_fn0039_config`, whose value you can access it from `self.fn0039_config` - doesn't have to be per-method. Or could you explain what are you trying to do here?

mirceaulinic (commented):
Nice work @thomasbridge74! This looks good to me, @ktbyers any opinion?

napalm/eos/eos.py
thomasbridge74: So obviously the main goal is to be able to convert commands between the old and new syntax.

Digging into how this was done for EAPI connections, I found inside the `run_commands` method of the `Node` inside the `pyeapi_syntax_wrapper.py` file the above code, and basically copied it to maintain the logic.

I assume your recommendation here is drop the line
```python
fn0039_transform = kwargs.pop("fn0039_transform", True)
```
and replace the following `if` statement with
```python
if self.fn0039_config
```

Happy to do that - it looks like I missed the significance of that setting as I poked around the code.

thomasbridge74:
While I've done the get functions, I still need to do the following:

1. ping.
2. traceroute
3. load_template

as well as update the neccessary functions for the configuration support matrix

thomasbridge74:
Just an update from me - I'm still experimenting with the functions to manage the device configuration.   I'm currently testing against a vEOS instance running 4.26.1F.

The challenge I'm facing is that the internal helper function `_get_pending_commits()` is returning false even when there is a commit waiting on the device.    Looking at the code, it appears to be looking for a key for `"pendingCommitTimer"` in the return output from `show configuration sessions detail`.     This isn't appearing in the dict returned either via EAPI or SSH:

```python
api_s = api_device._run_commands(["show configuration sessions detail"])
pprint(api_s)
[{'maxOpenSessions': 5,
  'maxSavedSessions': 1,
  'sessions': {'napalm_263187': {'description': '',
                                 'instances': {},
                                 'state': 'pending'},
               'napalm_350978': {'description': '',
                                 'instances': {},
                                 'state': 'pending'}}}]
```

This appears to be a bug that is more general than the transport type in use.

thomasbridge74:
I've completed the commands for configuration management.

One caveat: in order to get the SSH version of the session management to work, I've changed the logic in `commit config` from:

```
configuration session <sessionid>
commit [timer <revert_in>]
```

to 

```
configuration session <sessionid> commit [timer <revert_in>]
```

(ie a oneline command instead of two lines).   This is because it's a little messy using napalm with two lines considering that effectively enters configuration mode - which would also need to be quit but changes the prompt as well.      This may have issues.   I'll raise a question with my companies Arista Sales Architect.

thomasbridge74:
> I've completed the commands for configuration management.
> 
> One caveat: in order to get the SSH version of the session management to work, I've changed the logic in `commit config` from:
> 
> ```
> configuration session <sessionid>
> commit [timer <revert_in>]
> ```
> 
> to
> 
> ```
> configuration session <sessionid> commit [timer <revert_in>]
> ```
> 
> (ie a oneline command instead of two lines). This is because it's a little messy using napalm with two lines considering that effectively enters configuration mode - which would also need to be quit but changes the prompt as well. This may have issues. I'll raise a question with my companies Arista Sales Architect.

thomasbridge74:
I had confirmation from the Arista account contact who advised that 

> There shouldn't be any issues as the single line approach was implemented with config session timers from day 1 in 4.18.0F

So I'm happy leaving this modified approach if that's okay?

mirceaulinic:
> There shouldn't be any issues as the single line approach was implemented with config session timers from day 1 in 4.18.0F

I'm not a big fan of this, as this means only people running EOS 4.18.0F or newer will be able to use this. But otoh, I can't see a better workaround, so I'm fine to proceed. I'd like to request however to document this caveat. Thanks @thomasbridge74!

thomasbridge74:
just checking in - is there anything you're waiting on from me to merge this?

Two things with regard to the issue around 4.18.0F or newer 

1. Arista are only supporting 4.20F and above now.
2. The config session calls won't work with earlier versions regardless of EAPI or SSH connection as it didn't exist.   Nothing in my patch should therefore be different depending on transport.

mirceaulinic:
Hi @thomasbridge74 - yes, I think we want a documented caveat for this.

> 1. Arista are only supporting 4.20F and above now.

True, but I wouldn't assume there aren't people still EOS older than 4.20F (in fact, I'm absolutely sure there are). Here in NAPALM, is not our job to limit this.

> 2. The config session calls won't work with earlier versions regardless of EAPI or SSH connection as it didn't exist. Nothing in my patch should therefore be different depending on transport.

No so sure about that: in the past, I've personally used NAPALM with EOS versions earlier than 4.18 which made use of the configue sessions (this behaviour didn't change over the last 6 years at least).

scetron:
Hello @thomasbridge74! I have an interest in getting napalm to work with arista over ssh as well. I wanted to try out your extensions. I've noticed that I have a problem with opening connections to devices with optional args, specifically an alternative port. I believe the issue is the transport_class being a string doesn't allow all of the optional args to make it to the netmiko_optional_args dict that is used. 

I can share the full traceback if necessary and happy to have a chat or take a look my self if necessary. Thanks!

napalm/eos/eos.py
```diff
@@ -131,60 +136,120 @@ def _process_optional_args(self, optional_args):
             "transport", optional_args.get("eos_transport", "https")
         )
         self.fn0039_config = optional_args.pop("eos_fn0039_config", False)
-        try:
-            self.transport_class = pyeapi.client.TRANSPORTS[transport]
-        except KeyError:
-            raise ConnectionException("Unknown transport: {}".format(self.transport))
+        self.transport = transport
+        if transport == "ssh":
+            self.transport_class = "ssh"
```
scetron, line 141: transport_class as a string runs into an issue on line 150 and again in line 155 where a string only has self, so no optional args get passed into netmiko_optional_args.

napalm/eos/eos.py
```diff
@@ -131,60 +136,120 @@ def _process_optional_args(self, optional_args):
...
+            ret = []
+            for command in commands:
+                if kwargs.get("encoding") == "text":
+                    cmd_output = self._netmiko_device.send_command(command).replace(
+                        "% Invalid input", ""
+                    )
+                    ret.append({"output": cmd_output})
+                    continue
+
+                cmd_pipe = command + " | json"
```
bewing, line 240: This might have to be ` | json version 2`.  I can't remember exactly which version/reversion of JSON output EAPI uses.  Might have to ask Arista for exact clarification.

napalm/eos/eos.py
thomasbridge74: I've merged the changes to the napalm/eos/eos.py file.

thomasbridge74:
@mirceaulinic I've updated the pull request with the update to the driver file from @scetron but not including his (very substantial) suite of tests.    I've also made a couple of minor updates to the documentation.

With regard to the comment about 4.18.0F versus earlier versions, it would appear it was specifically the config timer that was introduced with 4.18.0F - I've included a comment about that in the documentation updates.        I can also be pinged on the NTC Slack if you think that might be a quicker path to resolving the issues here.

mirceaulinic (approved):
This looks good to me.

Thanks for your work @thomasbridge74! This will be included in release 4.0.0, which I target to release very soon.

napalm/eos/eos.py
mirceaulinic: CC @thomasbridge74 - I let you investigate this, feel free to come up with a subsequent PR if needed.

AshikKuppili:
@mirceaulinic, with this can we use SSH for EOS like NXOS_SSH , what should be driver name for this 
I tried today , but some how it failed for EOS , what should driver name 
Error I am getting is ""SSL Handshake not properly done"" something like that

ktbyers:
@AshikKuppili Try setting the transport in optional_args to be "ssh":

https://github.com/napalm-automation/napalm/blob/develop/napalm/eos/eos.py#L144

SkyFields:
> @AshikKuppili Try setting the transport in optional_args to be "ssh":
> 
> https://github.com/napalm-automation/napalm/blob/develop/napalm/eos/eos.py#L144

Hello, I am trying to use transport SSH for Arista with napalm-ansible, passing optional_args ssh, it looks like working, but I keep receiving an error
'AristaSSH' object has no attribute 'connection'

example of my task

`- name: get facts from device
      napalm_get_facts:
        hostname: '{{ ansible_host }}'
        username: '{{ ansible_user }}'
        dev_os: '{{ os }}'
        password: 
        optional_args:
          port: '{{ ansible_port }}'
          transport: 'ssh'
        filter: 'facts,interfaces,bgp_neighbors'
      register: result`

Issue #1508, opened 2021-10-12
JunOS: Caching of interface_list in get_facts()


kzorba:
### Caching of interface_list in get_facts

In junos driver when get_facts() is called a `get-interface-information` rpc call is issued that can be a very expensive operation in some cases (eg BNG devices). Caching of interface_list in facts can greatly benefit performance.

When NAPALM is used in a Salt(stack) environment during proxy minion startup `get_facts()` is called multiple times. On some devices, like BNG boxes, the `get-interface-information` is a very expensive operation and can cause the minion to wait multiple minutes to be able to start receiving and servicing requests from the salt master after its boot. While a lot of other facts are cached smartly in the underlying PyEZ library, NAPALM does not cache interface_lists. A PR will follow with a fix to enable further discussion. This caching mechanism could also benefit `get_interfaces()`.

### Setup
Install Salt v3002.7 and NAPALM 3.3.1, configure a proxy minion to target a BNG box with thousands on sessions terminated on it.

### napalm version

```
root@proxy-mxlar1-mq:~# pip freeze | grep napalm
napalm==3.3.1

```

### Network operating system version

Tested with: 

```
Junos: 17.3R3-S3.3

```

Problem is most probably there with other (or all) recent JunOS versions.

### Steps to Reproduce the Issue

Configure a Salt proxy minion targeting a BNG box with many sessions of users. Watch the startup process of the minion and observe the many times `get-interfaces-information` rpc is called via `get_facts()` that cause the minion to wait even for 10s of minutes.

network-shark:
@kzorba If you don’t need grains you could disable grains on your minion.

network-shark:
@kzorba

From the Saltstack docs , there is a blacklist grains option . 

```
 grains_blacklist:
  - cpu_flags
  - zmq*
  - ipv[46]
```

What do you think ?

kzorba:
Hello @network-shark, I had seen this option and I just tested it. It works as I expected, the grains collections happen like before but the specific items you put in this option are filtered out afterwards. No effect in my problem, the router is overwhelmed with get-interface-information netconf requests. I believe the way to go is to use 'provider' option in napalm proxy minion, overriding napalm_base with my implementation.

network-shark:
@kzorba Ok , maybe the naming here is a bit odd. For me blacklisting is that it should not be executed anyway. But maybe this is more a salt issue / feature request. 

I think you are right , you need to patch it yourself.

PR #1509, opened 2021-10-12
Cache interface list in get_facts for junos.

kzorba:
In junos driver when get_facts() is called a `get-interface-information` rpc call is issued that can be a very expensive operation in some cases (eg BNG devices). Caching of interface_list in facts can greatly benefit performance. This fix can be applied to other places too, like `get_interfaces()`. Open to discussion.

Relevant issue: napalm-automation/napalm#1508

### DIFF
--- a/napalm/junos/junos.py
+++ b/napalm/junos/junos.py
@@ -132,6 +132,8 @@ class JunOSDriver(NetworkDriver):
             )
 
         self.platform = "junos"
+        # caching of interface list on device
+        self.interface_list = []
         self.profile = [self.platform]
 
     def open(self):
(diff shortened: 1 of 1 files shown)


mirceaulinic:
I appreciate the intent (and your work), but this can have unwanted side effects when you run consecutive `get_facts` over the same session, i.e., we don't want a stale list of interfaces, rather put more workload than having incorrect data.
Happy to discuss further though! 😄

kzorba:
Hello @mirceaulinic,

thanks for your input, however in my setup I really have to find a solution, minions targeting BNG devices take ~15min to respond after a start-up or restart. I cannot estimate the possible side-effects (you are the expert in this code base of course). I imagine we could apply the caching in get_facts() and a get_interfaces() could always refresh the cache. Would that be an option? 

From the little I saw, the underlying PyEZ library does a lot of caching internally.

Is there any other option for speeding-up the start-up of a minion? The problem in Salt is that get_facts() seems to be called many times...

Best regards,
Kostas

PR #1511, opened 2021-10-13
Return mtu -1 and speed -1.0 if regex does not find the data in nxos_ssh

ubaumann:
I have a Nexus with a display bug. This PR should make the NXOS driver a little more stable but I tottaly understand if you don't want to merge it.

If the regex search for MTU and speed fails, MTU -1 and speed -1.0 will be returned instead. 

The cisco bug I am facing:

On some interfaces it does not show the MTU in the `show interface`

```
Ethernet1/2/3 is down (Link not connected)
  Hardware: 10/100/1000 Ethernet, address: 0cd9.9608.0d6f (bia 0cd9.9608.0d6f)
, BW 23 Kbit, DLY 10 usec
```

Correct would be:
```
Ethernet1/2/3 is down (Link not connected)
  Hardware: 10/100/1000 Ethernet, address: 0cd9.9608.0d6f (bia 0cd9.9608.0d6f)
  MTU 9612 bytes, BW 23 Kbit, DLY 10 usec
```

### DIFF
--- a/test/nxos_ssh/mocked_data/test_get_interfaces/missing_mtu/expected_result.json
+++ b/test/nxos_ssh/mocked_data/test_get_interfaces/missing_mtu/expected_result.json
@@ -1,0 +1,56 @@
+{
+    "Management0": {
+        "is_enabled": true, 
+        "description": "", 
+        "last_flapped": -1.0,
+        "is_up": true,
+        "mac_address": "2C:C2:60:12:EC:74",
+        "mtu": 1500,
+        "speed": 1000
+    },
+    "Ethernet2/1": {
+        "is_enabled": true,
+        "description": "Testing port descriptions",
+        "last_flapped": -1.0,
+        "is_up": true,
+        "mac_address": "2C:C2:60:4F:FE:B2",
+        "mtu": -1,
+        "speed": 1000
+    },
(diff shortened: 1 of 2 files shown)


napalm/nxos_ssh/nxos_ssh.py
```diff
@@ -164,17 +164,21 @@ def parse_intf_section(interface):
...
+            # This was alway in Kbit (in the data I saw)
+            if speed_unit != "Kbit":
+                msg = "Unexpected speed unit in show interfaces parsing:\n\n{}".format(
+                    interface
+                )
+                raise ValueError(msg)
+            speed = float(speed / 1000.0)
+        else:
+            speed = -1.0
+            mtu = -1
```
mirceaulinic, line 181: For the MTU I agree it should be -1 when this information is missing, but the speed is there, we should just try to match a different regex. Perhaps wrap the MTU section from `re_speed` into a `()?` block? (i.e., `r"(\s+MTU (?P<mtu>\S+)\s+bytes)?,\s+BW\s+(?P<speed>\S+)\s+(?P<speed_unit>\S+).*$"`)

mirceaulinic (changes requested):

PR #1515, opened 2021-11-04
fix flaky test_unit.py::test__compare_getter_list

shenganzhang:
This PR aims to fix the flaky test `test_unit.py::test__compare_getter_list`. In previous versions, the test will run into failure when running for multiple times. And the reason is that the parameters `src, dst, result` got changed because of the `pop()` operations in `napalm/base/validate.py`. A deep copy can fix this issue. The test failure can be reproduced by 
`pip install pytest-flakefinder`
`pytest --flake-finder --flake-runs=2 test/base/validate/test_unit.py`
Notice that the PR is modifying the test to make it more robust without changing the source code.

### DIFF
--- a/test/base/validate/test_unit.py
+++ b/test/base/validate/test_unit.py
@@ -1,5 +1,6 @@
 """Tests for the validate methods."""
 import pytest
+import copy
 
 from napalm.base import validate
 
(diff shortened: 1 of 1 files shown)


mirceaulinic:
Hey @shenganzhang, thanks for your PR! The tests are currently failing due to a formatting error, could you check?

shenganzhang:
> Hey @shenganzhang, thanks for your PR! The tests are currently failing due to a formatting error, could you check?

Thanks. I ran `python -m black` on the modified file and updated it.

mirceaulinic (approved):

PR #1522, opened 2021-11-23
if we are ignoring warnings on commit, also ignore them on discard

ekoyle:
Make a similar change to #1480 to allow more graceful recovery from expected configuration warnings (see [this comment](https://github.com/napalm-automation/napalm/pull/1480#issuecomment-905961019)).

The changes in [py-junos-eznc pull 1131](https://github.com/Juniper/py-junos-eznc/pull/1131) are included in junos_eznc-2.6.3 (I'm not sure which version first included the change).


mirceaulinic (approved):
Makes sense to me.

PR #1545, opened 2022-01-23
Escape vlan_name before building regex string

icovada:
Sanitise input before building the regex with it

Current code fails when vlan_name begins with a regex special character such as "*"


icovada:
Wrong commit sent. Will close and re-open

Issue #451, opened 2017-10-24
Inconsistent MAC and IPv6 format for several IOS getters


mirceaulinic:
### Description of Issue/Question

https://github.com/napalm-automation/napalm-ios/pull/110 shows that as of napalm-ios 0.6.0, the following methods need a bit of work and use the `ip` and `mac` napalm-base helpers:

- [x] get_mac_address_table, fixed in https://github.com/napalm-automation/napalm-ios/pull/110
- [x] get_arp_table, fixed in https://github.com/napalm-automation/napalm-ios/pull/110
- [x] get_interfaces
- [ ] get_lldp_neighbors
- [ ] get_lldp_neighbors_detail
- [ ] get_bgp_neighbors
- [ ] get_interfaces_ip
- [ ] get_ntp_peers
- [ ] traceroute
- [ ] ping

### napalm-ios version
(Paste verbatim output from `pip freeze | grep napalm-ios` between quotes below)

```
0.6.0
```

---

Issue moved from https://github.com/napalm-automation/napalm-ios/issues/111

mundruid:
I am interested in pushing this issue through the finish line. Let me know if this is still valid or it has already been fixed. Thanks!

mirceaulinic:
Resolved via #1560.

PR #1560, opened 2022-02-12
Adds mac, ip checks for getters

mundruid:
Adds mac / ip checks for:
```
get_lldp_neighbors
get_lldp_neighbors_detail
get_interfaces_ip
get_ntp_peers
traceroute
ping
``` 

Covers https://github.com/napalm-automation/napalm/issues/451

### DIFF
--- a/napalm/ios/ios.py
+++ b/napalm/ios/ios.py
@@ -906,7 +906,9 @@ class IOSDriver(NetworkDriver):
         # format chassis_id for consistency
         for entry in lldp_entries:
             entry["remote_chassis_id"] = napalm.base.helpers.convert(
-                napalm.base.helpers.mac, entry["remote_chassis_id"]
+                napalm.base.helpers.mac,
+                entry["remote_chassis_id"],
+                entry["remote_chassis_id"],
             )
 
         # Older IOS versions don't have 'Local Intf' defined in LLDP detail.
--- a/napalm/ios/ios.py
+++ b/napalm/ios/ios.py
@@ -911,7 +911,6 @@ class IOSDriver(NetworkDriver):
                 entry["remote_chassis_id"] = napalm.base.helpers.mac(
                     entry["remote_chassis_id"]
                 )
-                print(entry["remote_chassis_id"])
 
         # Older IOS versions don't have 'Local Intf' defined in LLDP detail.
         # We need to get them from the non-detailed command
--- a/napalm/ios/ios.py
+++ b/napalm/ios/ios.py
@@ -903,14 +903,11 @@ class IOSDriver(NetworkDriver):
         if len(lldp_entries) == 0:
             return {}
 
-        # format chassis_id for consistency, if it is a mac address
+        # format chassis_id for consistency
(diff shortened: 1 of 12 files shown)


napalm/ios/ios.py
```diff
@@ -901,6 +903,16 @@ def get_lldp_neighbors_detail(self, interface=""):
...
 
+        # format chassis_id for consistency, if it is a mac address
+        for entry in lldp_entries:
+            if re.search(
+                r"^(\d|\w){4}.(\d|\w){4}.(\d|\w){4}$", entry["remote_chassis_id"]
+            ):
+                entry["remote_chassis_id"] = napalm.base.helpers.mac(
+                    entry["remote_chassis_id"]
+                )
+                print(entry["remote_chassis_id"])
```
mirceaulinic, line 914:
```suggestion
```

napalm/ios/ios.py
```diff
@@ -901,6 +903,16 @@ def get_lldp_neighbors_detail(self, interface=""):
         if len(lldp_entries) == 0:
             return {}
 
+        # format chassis_id for consistency, if it is a mac address
+        for entry in lldp_entries:
+            if re.search(
+                r"^(\d|\w){4}.(\d|\w){4}.(\d|\w){4}$", entry["remote_chassis_id"]
```
mirceaulinic, line 909: Instead of doing this regex check, I'd just apply the `convert` function instead, e.g.,

```suggestion
entry["remote_chassis_id"] = napalm.base.helpers.conver(napalm.base.helpers.mac, entry["remote_chassis_id"], entry["remote_chassis_id"])
```

mirceaulinic (changes requested):
It feels great align everything! Two changes requested, then should be able to get this in.

napalm/ios/ios.py
mundruid: Thanks for the quick review and all the suggestions. Just checking: I agree with your suggestion, but if the chassis id cannot be converted, an empty chassis id will be returned. This is an example value from a chassis id from one of the tests:
```
In [75]: id = "hmbbkku-la"

In [76]: id = napalm.base.helpers.convert(napalm.base.helpers.mac, id)

In [77]: id
Out[77]: ''
```

Is that OK?

mundruid: I have addressed the comment with the latest commit. Let me know what you think @mirceaulinic

mirceaulinic: Yes, there should be a third argument, with the default, notice in the suggestion above:

```python
entry["remote_chassis_id"] = napalm.base.helpers.conver(napalm.base.helpers.mac, entry["remote_chassis_id"], entry["remote_chassis_id"])
```

When unable to convert to the normalised MAC format, it'd default to the extracted value.

mundruid: Thank you for the help! Added.

mirceaulinic (approved):

mirceaulinic:
Thank you @mundruid!

PR #1563, opened 2022-02-13
Add ttp_parse helper function to codebase

dmulyalin:
# What this PR does

This PR developed as part of NANOG84 Hackathon efforts.

This PR add new helper function `ttp_parse`.

# Motivation

Debugging and developing regular expressions can be tedious process, for some use cases it is possible to use TTP as a drop in to produce same results but with less efforts. Besides parsing, TTP also might bring other use cases like results data validation, formatting, testing or postprocessing.

# How this was tested

Updated `test/base/test_helpers.py` file with `test_ttp_parse` test that does 5 sub-tests:

* check if raises TemplateRenderException when template does not exist
* check if raises TemplateRenderException for /utils/ttp_templates/bad_template.txt
* check if returns expected result as output for ./utils/ttp_templates/good_template.txt
* check if returns expected result as output for inline template
* check if returns expected result as output for template from ttp templates

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -47,5 +47,8 @@ ignore_missing_imports = True
 [mypy-textfsm]
 ignore_missing_imports = True
 
+[mypy-ttp]
+ignore_missing_imports = True
+
 [mypy-pytest]
 ignore_missing_imports = True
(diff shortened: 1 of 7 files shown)


mirceaulinic (commented):
Looks good, just please fix the lint errors, then we should be able to take this in.

mirceaulinic (approved):

PR #1564, opened 2022-02-13
Replace confparse w netutils

ktbyers:


jvanderaa:
I see what is going on here. Even though the variable name has `text` in it, it is really a list. The result from the previous function was a list, but it also accepted a list. The netutils library expects that the configuration incoming is a string. In working through one aspect of it I resolved 2 of the errored tests by updating this behavior. Going to try to take this into account a little better yet.

PR #1573, opened 2022-02-21
iss-1416

emfusion:
- fix get_bgp_neighbors_detail() when using  multi-agent mode
- fix issue where multihop var was wrongly interpreted as it was comparing 0 with ‘0’
- Add IPv6 prefixes to the advertised and received prefix count
- Add more test for get_bgp_neighbors_detail


napalm/eos/eos.py
```diff
@@ -1682,13 +1682,26 @@ def _parse_per_peer_bgp_detail(peer_output):
                 "configured_keepalive",
                 "advertised_prefix_count",
                 "received_prefix_count",
+                "advertised_ipv6_prefix_count",
+                "received_ipv6_prefix_count",
             ]
 
             peer_details = []
 
+            # determine if in multi-agent mode to get correct extrator_type
```
fdomain, line 1691: typo: extractor

napalm/eos/eos.py
```diff
@@ -1740,6 +1753,12 @@ def _parse_per_peer_bgp_detail(peer_output):
                 item["local_address"] = napalm.base.helpers.convert(
                     napalm.base.helpers.ip, item["local_address"]
                 )
+                # Get all the advertised prefixes
+                item["advertised_prefix_count"] += item["advertised_ipv6_prefix_count"]
+                item["received_prefix_count"] += item["received_ipv6_prefix_count"]
+                # Remove the ipv6_prefix for comformity with test_model
```
fdomain, line 1759: typo: conformity

napalm/eos/eos.py
```diff
@@ -1740,6 +1753,12 @@ def _parse_per_peer_bgp_detail(peer_output):
                 item["local_address"] = napalm.base.helpers.convert(
                     napalm.base.helpers.ip, item["local_address"]
                 )
+                # Get all the advertised prefixes
+                item["advertised_prefix_count"] += item["advertised_ipv6_prefix_count"]
+                item["received_prefix_count"] += item["received_ipv6_prefix_count"]
```
fdomain, line 1758: Why do you mix ipv4 and ipv6 prefixes in the same counter ?
Shouldn't be ipv6 prefixes counter added to its own counter ?

napalm/eos/eos.py
```diff
@@ -1774,7 +1793,7 @@ def _append(bgp_dict, peer_info):
                 commands.append("show ip bgp neighbors %s vrf all" % neighbor_address)
                 summary_commands.append("show ip bgp summary vrf all")
             elif peer_ver == 6:
-                commands.append("show ipv6 bgp neighbors %s vrf all" % neighbor_address)
+                commands.append("show ipv6 bgp %s vrf all" % neighbor_address)
```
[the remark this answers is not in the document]
emfusion: no. this was an error did not mean to remove it

napalm/eos/eos.py
emfusion: the other devices share the naming so adding a separate counter for EOS will also mean adding for the other devices.

fdomain: Ok fair enough.

napalm/eos/eos.py
```diff
@@ -1682,13 +1682,26 @@ def _parse_per_peer_bgp_detail(peer_output):
...
                 "advertised_prefix_count",
                 "received_prefix_count",
+                "advertised_ipv6_prefix_count",
+                "received_ipv6_prefix_count",
             ]
 
             peer_details = []
 
+            # determine if in multi-agent mode to get correct extrator_type
+            commands = [
```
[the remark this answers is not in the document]
emfusion: I did that at start but adding correct testing was a pain. 
If more discrepancies are found, this is something that could be done/added later. For now, I think this is the best way to do

napalm/eos/eos.py
```diff
@@ -1699,7 +1712,7 @@ def _parse_per_peer_bgp_detail(peer_output):
                     True if item["local_address"] else False
                 )
                 item["multihop"] = (
-                    False if item["multihop"] == 0 or item["multihop"] == "" else True
+                    False if item["multihop"] == "0" or item["multihop"] == "" else True
```
mirceaulinic, line 1715: In the TexFSM template, the value would seem to be extracted as a number (`Value multihop (\d+)`). Could you clarify why comparing against string (`"0"`) here?

napalm/eos/eos.py
```diff
@@ -1682,13 +1682,26 @@ def _parse_per_peer_bgp_detail(peer_output):
...
+                "received_ipv6_prefix_count",
             ]
 
             peer_details = []
 
+            # determine if in multi-agent mode to get correct extractor_type
+            commands = [
+                "show running-config | include service routing protocols model multi-agent"
+            ]
+            is_multi_agent = self.device.run_commands(commands, encoding="text")[0].get(
```
mirceaulinic, line 1695: Can probably get rid of the `commands` variable:

```suggestion
            is_multi_agent = self.device.run_commands(
                [
                    "show running-config | include service routing protocols model multi-agent"
                ],
                encoding="text",
            )[0].get()
```

mirceaulinic (approved):
Looks good overall, only had one question and a minor nit.

napalm/eos/eos.py
emfusion: Yes, so the TexFSM template is expecting a regular expression (\d+) for the multihop variable but the output remains a string : [example](https://pyneng.readthedocs.io/en/latest/book/21_textfsm/README.html#:~:text=The%20result%20of%20script%20execution)
that is also why the eos.py file still uses the "napalm.base.helpers.convert()" function to convert some of the extracted variables to int.

bewing:
Little late the the party here, apologies.  

Are the two output formats so different we can't use a single template?  Recognizing the start of the next record should allow one to close out the previous record -- this approach is used in the VRF template: https://github.com/napalm-automation/napalm/blob/develop/napalm/eos/utils/textfsm_templates/vrf.tpl#L7

mirceaulinic:
Merged as-is, but feel free to follow up with another PR and we shall review! 😄

emfusion:
> @emfusion have you had a chance to look at Brandon's suggestion above? This still LGTM otherwise...

Hi, I unfortunately havent had the time to look into it (forgot how time consuming newborns are 😆). It's on my To Do list and will definitely jump into it soon

PR #1575, opened 2022-02-23
Use regex module if available

DavidVentura:
The [regex module](https://pypi.org/project/regex/) is a drop-in replacement for `re`, with better support for concurrency, as it drops the GIL while parsing

When running concurrently on hundreds of devices, this allows significantly better responsiveness for the process

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -26,6 +26,10 @@ ignore_errors = True
 [mypy-napalm.base.test.*]
 ignore_errors = True
 
+[mypy-regex.*]
+ignore_errors = True
+ignore_missing_imports = True
+
 [mypy-netaddr.*]
 ignore_missing_imports = True
 
--- a/napalm/base/helpers.py
+++ b/napalm/base/helpers.py
@@ -4,7 +4,11 @@ import logging
 
 # std libs
 import os
-import re
+
+try:
+    import regex as re
+except ModuleNotFoundError:
+    import re  # type:ignore
 import sys
 from typing import Optional, Dict, Any, List, Union, Tuple, TypeVar, Callable
 from collections.abc import Iterable
--- a/napalm/base/mock.py
+++ b/napalm/base/mock.py
@@ -20,7 +20,11 @@ import napalm.base.exceptions
 import inspect
 import json
 import os
-import re
+
+try:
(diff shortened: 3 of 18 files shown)


DavidVentura:
I wouldn't agree with calling napalm "hugely IO bound". I'll let you decide whether my use-case is too niche to be merged.

I am running against ~600 devices from an asyncio application, which very frequently runs to 100% cpu usage due to regex parsing of device config. 
Not all of it is directly stemming from napalm (see [pyeapi PR](https://github.com/arista-eosplus/pyeapi/pull/220) for one such case). My ios devices' config are between 405KB and 1.7MB, and this change shows significantly better responsiveness.
I have not precisely measured the total execution time yet between both versions, they look similar.

ktbyers:
Knock yourself out i.e. create a fork and do what you want, but I don't think we want it as part of NAPALM itself (i.e. the percent that benefit from this is very small and the risk of a lot of extra debugging cost/time is high).

You also can just monkey patch re to use your regex and not this...so you can probably do it without even creating a fork.

mirceaulinic:
Yes, I agree with what Kirk mentioned, the vast majority of the time is spent on interacting with the network device, rather than what NAPALM computes. I'm absolutely convinced improvements can and should be done in various places whenever possible... But I don't feel like I would trade off a standard library for a couple of miliseconds speed up when the execution times is typically measured in seconds.
That said however, I'd love to see some numbers @DavidVentura - could you do a little research and compare the execution times of NAPALM methods when used with `re` vs. `regex`?

DavidVentura:
Sure, will get you some numbers next week

network-shark:
There is an lwn.net article on this topic it is mainly about that re is really old an regex could be a drop in replacement for the re module like the requests lib for urllib ? . I am not able to post the link , because it's  for members... only 

One benchmark from the article :) 
```
A test data set of 700MB was processed in 77 seconds with re but it took 92 seconds with regex. 

```

DavidVentura:
In the end the performance was not great -- the time to execute was slightly larger, and I am not entirely sure how to measure how long the GIL is held during execution, I'd rather not merge this without concrete benchmarks either way

PR #1576, opened 2022-03-02
Move models out of tests

itdependsnetworks:
I was going to reference these models in napalm-panos, but then I saw them in a test folder. To me that indicates that there is no expected semver for them. I think this makes sense to move the models out of a test folder.


ktbyers:
Are we going to break the community drivers because of this change?

Since the `base/test` folder is already inside of the NAPALM code directory, I think it is already under semantic versioning (or we can just declare it as being so).

Here I am making a distinction between:

```
{ napalm-repo }/test
```

And

```
{ napalm-repo }/napalm          [and all things under it]
```

@itdependsnetworks Does it work for you if we just declare that anything inside `{ napalm-repo }/napalm` is covered by sematic versioning?

itdependsnetworks:
> Are we going to break the community drivers because of this change? 

No, there hasn't been a release since June, this was pushed in Feb. 

imho keeping it as is does not pass principal of least astonishment. As a developer, I did not think to look in tests directory, and was confused why it would be there once I found it. Further, there was actually an issue found here, that I would think would be likely to happen again. Specifically, the mypy check was turned off, as presumably other future linting may as well. 

Not going to lose sleep over it, but I think it makes sense to move. Ev

ktbyers:
I didn't follow your `this was pushed in Feb` statement, models.py has been in that base/test dir for a long time?

Here is NAPALM 3.2.0:

https://github.com/napalm-automation/napalm/blob/3.2.0/napalm/base/test/models.py


```
Further, there was actually an issue found here, that I would think would be likely to happen again. Specifically, the mypy check was turned off, as presumably other future linting may as well.
```

I was out of the loop on this, what was the `issue found here`. I agree though that we should not be disabling linting/mypy on it (so that should be fixed regardless).

I don't care one way or the other besides the breaking existing code and that I definitely care about :-)

itdependsnetworks:
Take another look at that file, I don’t think there is a single thing that is the same between 3.3 and develop branch.

ktbyers:
Okay, that makes sense.

So basically it has all changed between NAPALM 3.3.1 and `develop` branch so it is already all backwards incompatible.

It is fine with me to move this.

@mirceaulinic Are you okay with doing this? We should probably do it before the next release (if we are going to do it).

itdependsnetworks:
@mirceaulinic does this look good?

ktbyers (approved):

ktbyers:
I think this is probably okay so merging.

If not, I will fix it.

napalm/base/models.py
```diff
@@ -400,31 +400,6 @@
...
-)
-
-OpticsPerChannelDict = TypedDict(
-    "OpticsPerChannelDict", {"index": int, "state": OpticsStatePerChannelDict}
-)
-
-OpticsPhysicalChannelsDict = TypedDict(
-    "OpticsPhysicalChannelsDict", {"channels": OpticsPerChannelDict}
-)
-
```
ktbyers, line 427: @itdependsnetworks Why were these ones removed?

ktbyers (changes requested):

ktbyers:
Okay, there was one outstanding question that I had when doing a final review (see above).  @itdependsnetworks

napalm/base/models.py
Kircheneer: I introduced these as duplicates within #1476 - when Ken enabled type hinting for that file Mypy caught it and it had to be removed.

ktbyers (approved):

PR #1599, opened 2022-03-25
Remove PY3.6 support, fix type hint bug in py3.10 and add py3.10 test support

ktbyers:


ktbyers:
@mirceaulinic The first commit here is an error that breaks NAPALM on PY3.10

The other changes are to enable testing on PY3.10. I tried to use even newer versions of both py.test and pylama, but run into other issues.

Note, I removed PY3.6 support as the pylama library I needed to use for PY3.10 had dropped support for it (so I ran into a test issue). I thought this was okay since PY3.6 is EOL.

mirceaulinic (approved):
That's great - thanks for your work and details on this, Kirk!

PR #1607, opened 2022-04-05
init variables used for total and used mem

tealflipper:
Fixes error given in python 3.8 where total_mem cannot be calculated due to proc_total_mem not being initialized with a value


ktbyers:
@tealflipper Do you have an example where this is happening?

i.e. what does `show memory statistics` show for output on the failing device?

I would say that we probably shouldn't need to initialize the values i.e. that the parsing should always yield values (if it doesn't something else has gone wrong).

ktbyers:
Any more details @tealflipper ?

tealflipper:
Searching for the error, I found I was not using the correct driver. Thank
you!

PR #1608, opened 2022-04-06
ignore_warnings argument requires Juniper pyez >= 2.6.3

ktbyers:


mirceaulinic (approved):

mirceaulinic:
To avoid this annoying error, was wondering if you'd agree to port this commit into the `master` branch and release 3.4.1 before 4.0.0? Ideally we should release 4.0.0 soon(-ish) but there are some features I'd like to include there, that aren't ready yet...

PR #1611, opened 2022-04-07
Roll to pyez 2.6.3 to fix ignore_warnings (push to master)

ktbyers:


requirements-dev.txt
```diff
@@ -10,3 +10,6 @@ pylama==7.7.1
 mock==4.0.3
 tox==3.24.5
 mypy==0.931
+# Force click to previous release so black/PY3.6 is not broken
+# Can be removed in NAPALM >= 4.0.0
+click==8.0.4
```
mirceaulinic, line 15: I think `black` release 22.3.0 has fixed this? Or you've run into other issues with it?

ktbyers: The underlying issue is that Click > 8.0.4 removed PY3.6 support.

Black fixed this in 22.3.0 by going with a downrev Click, see:

https://github.com/psf/black/pull/2966/files

I was trying to keep changes minimal and didn't want to redo the fix that we did for dropping PY3.6 support.

This is also just a `dev` dependency so pinning it to click==8.0.4 should be fine (it is fixed already in `develop` with the PY3.6 removal). Alternatively, we could say click<=8.0.4, but I don't think it matters much (since it is a `dev/tools` dependency and not a part of the napalm release itself).

PR #1613, opened 2022-04-11
Convert NAPALM over to use Netmiko >= 4.0.0

ktbyers:


ktbyers:
Also tested using `napalm_custom_test` repository where I tested against real/virtual IOS, NX-OS, and IOS-XR device.

PR #1627, opened 2022-04-28
Fix get_lldp_neighbors_detail getter for nxos_ssh

gneville-ot:
- show_lldp_neighbors_detail.tpl was missing as a TextFSM template for nxos_ssh


mirceaulinic:
@gneville-ot `get_lldp_neighbors_detail` isn't currently implemented for the NX-OS SSH driver, and just adding the TextFSM template isn't sufficient. You can use the `get_environment` method as an example: very briefly, you need, at minumum the following steps: execute a show command, pass the output to the TextFSM template, then map out the values to the method output.

PR #1634, opened 2022-04-29
Nxos ssh interface parsing fixes

jpsutton:
I had some issues calling get_interfaces() on one of my Nexus 6k units. These changes fix the parsing issues.


jpsutton:
The current form of this change seems to fail as assignment expressions (PEP 572) aren't supported until Python 3.8. Boo.

mirceaulinic (commented):
Code looks good. Could you also add a test case for the output you had?

PR #1636, opened 2022-05-05
Fix pyIOSXR netmiko exception import path

dkaplan1:
https://github.com/ktbyers/netmiko/commit/4782157c920bb7f17f61b6ea9897ff2aa3bd78d6#diff-d624fe8122a167b02cbacae5860d4ecafb637c8837c02d285c122844a94667bc changed the path of these exceptions.


ktbyers:
Thanks for finding this. I had a weird environment/packaging issue where those errors were not occurring for me (which is a different netmiko issue/bug I need to fix).

ktbyers (approved):

Issue #1500, opened 2021-09-24
Junos "| display json" in cli method


seismiccollision:
### Description of Issue/Question
I would like to be able to run arbitrary show commands on junos and get the JSON or XML data back instead of stdout
It seems like junos.py isn't setup for handling "| display" and just drops it
Is that something you'd consider adding support for?

The way NX-OS SSH works I am able to do the above equivalent

*Note*: Please check https://guides.github.com/features/mastering-markdown/
to see how to properly format your request.

### Did you follow the steps from https://github.com/napalm-automation/napalm#faq
(Place an ``x`` between the square brackets where applicable)

- [x] Yes
- [] No

### napalm version
(Paste verbatim output from `pip freeze | grep napalm` between quotes below)

```
napalm==3.3.1
```

### Network operating system version
(Paste verbatim output from `show version` - or equivalent - between quotes below)

```
Junos: 20.2R3-S2.5
```

### Steps to Reproduce the Issue
If I run "show lacp interfaces | display json" it doesn't honor the format request

```
>>> device.cli(['show lacp interfaces | display json'])
[...]

PR #1637, opened 2022-05-12
EOS: Add a JSON option when running a cli command

dasturiasArista:
pyeapi supports JSON as an option for the cli commands. However right
now the JSON option isn't exposed to users of this API. This exposes the
JSON option as well.

I manually tested this with an Arista switch and verified the changes work.


ktbyers:
Tests are failing on this change.

dasturiasArista:
> Tests are failing on this change.

I was noticing that. I'm wondering the best way to handle adding functionality would be. It looks like tests are failing because I've added an extra argument to this function that it's not expecting. Would it ok to modify the test to allow the extra args in the EOS case?

ktbyers:
So the test issue is that all of the NAPALM drivers (junos, eos, ios, etc) must have the same signature for the method:

```
    def cli(self, commands, encoding="text"):
```

So if we add the `encoding` argument, then we have to add it for all of the drivers.

You can always access pyeapi via napalm using:

```
napalm_object.device
```

I guess I wonder if we really need to be able to execute the `cli` method but with JSON output. Note, I basically never use the `cli` method...so probably would be better if @mirceaulinic or other users of CLI commented on this and whether it is worth doing this.

dasturiasArista:
> So the test issue is that all of the NAPALM drivers (junos, eos, ios, etc) must have the same signature for the method:
> 
> ```
>     def cli(self, commands, encoding="text"):
> ```
> 
> So if we add the `encoding` argument, then we have to add it for all of the drivers.
> 
> You can always access pyeapi via napalm using:
> 
> ```
> napalm_object.device
> ```
> 
> I guess I wonder if we really need to be able to execute the `cli` method but with JSON output. Note, I basically never use the `cli` method...so probably would be better if @mirceaulinic or other users of CLI commented on this and whether it is worth doing this.

Ok. that seems reasonable. I was unaware that the "device" attribute was considered a stable part of the api. I have proposed this solution for the customer trying to use this to confirm if this works for them or not.

ktbyers:
Yeah, I would think we would `not` change `napalm_object.device` (at least not without really good reason and without following semantic versioning i.e. a major release) as that would break lots of things.

It is possible this `napalm_object.device` convention is not followed in napalm-community drivers, but it should be followed in napalm core.

ktbyers:
@dasturiasArista Yeah, let us know what you find and if that solution works.

dasturiasArista:
> @dasturiasArista Yeah, let us know what you find and if that solution works.

Yes will do! Not to keep you hanging, still waiting from the response from the customer.

mirceaulinic:
Just to add to Kirk's comment further:

> So if we add the encoding argument, then we have to add it for all of the drivers.

We have a similar request for Junos (#1500) so it'd be great if you could go ahead with this and I can take care of the Junos implementation.

dasturiasArista:
> Just to add to Kirk's comment further:
> 
> > So if we add the encoding argument, then we have to add it for all of the drivers.
> 
> We have a similar request for Junos (#1500) so it'd be great if you could go ahead with this and I can take care of the Junos implementation.

Ok. I can go ahead and add the argument to all of the drivers, but for those that don't support that encoding i'll just have it raise an error. How do you feel about raising a `ValueError`?

dasturiasArista:
> > Just to add to Kirk's comment further:
> > > So if we add the encoding argument, then we have to add it for all of the drivers.
> > 
> > 
> > We have a similar request for Junos (#1500) so it'd be great if you could go ahead with this and I can take care of the Junos implementation.
> 
> Ok. I can go ahead and add the argument to all of the drivers, but for those that don't support that encoding i'll just have it raise an error. How do you feel about raising a `ValueError`?

Btw I have proactively made the changes we talked about here. Also, each method will check that it's valid and throw a `ValueError` if it's not a valid encoding.

napalm/eos/eos.py
```diff
@@ -885,17 +885,21 @@ def get_lldp_neighbors_detail(self, interface=""):
                 )
         return lldp_neighbors_out
 
-    def cli(self, commands):
+    def cli(self, commands, encoding="text"):
+        if encoding not in ("text", "json"):
+            raise ValueError("%s is not a supported encoding" % encoding)
```
mirceaulinic, line 890:
```suggestion
            raise NotImplementedError("%s is not a supported encoding" % encoding)
```

napalm/ios/ios.py
```diff
@@ -2375,6 +2375,8 @@ def cli(self, commands):
             'show clock': u'*22:01:51.165 UTC Thu Feb 18 2016'}
 
         """
+        if encoding not in ("text",):
+            raise ValueError("%s is not a supported encoding" % encoding)
```
mirceaulinic, line 2379:
```suggestion
            raise NotImplementedError("%s is not a supported encoding" % encoding)
```

napalm/iosxr/iosxr.py
```diff
@@ -885,7 +885,9 @@ def get_lldp_neighbors_detail(self, interface=""):
 
         return lldp_neighbors
 
-    def cli(self, commands):
+    def cli(self, commands, encoding="text"):
+        if encoding not in ("text",):
+            raise ValueError("%s is not a supported encoding" % encoding)
```
mirceaulinic, line 890:
```suggestion
            raise NotImplementedError("%s is not a supported encoding" % encoding)
```

napalm/junos/junos.py
```diff
@@ -1012,8 +1012,10 @@ def get_lldp_neighbors_detail(self, interface=""):
 
         return lldp_neighbors
 
-    def cli(self, commands):
+    def cli(self, commands, encoding="text"):
         """Execute raw CLI commands and returns their output."""
+        if encoding not in ("text",):
+            raise ValueError("%s is not a supported encoding" % encoding)
```
mirceaulinic, line 1018:
```suggestion
            raise NotImplementedError("%s is not a supported encoding" % encoding)
```

napalm/nxos/nxos.py
```diff
@@ -1127,7 +1127,11 @@ def get_bgp_neighbors(self) -> Dict[str, models.BGPStateNeighborsPerVRFDict]:
             results[vrf_name] = result_vrf_dict
         return results
 
-    def cli(self, commands: List[str]) -> Dict[str, Union[str, Dict[str, Any]]]:
+    def cli(
+        self, commands: List[str], encoding: str = "text"
+    ) -> Dict[str, Union[str, Dict[str, Any]]]:
+        if encoding not in ("text",):
+            raise ValueError("%s is not a supported encoding" % encoding)
```
mirceaulinic, line 1134:
```suggestion
            raise NotImplementedError("%s is not a supported encoding" % encoding)
```

napalm/nxos_ssh/nxos_ssh.py
```diff
@@ -787,7 +787,9 @@ def get_bgp_neighbors(self):
         # FIX -- need to merge IPv6 and IPv4 AFI for same neighbor
         return bgp_dict
 
-    def cli(self, commands):
+    def cli(self, commands, encoding="text"):
+        if encoding not in ("text",):
+            raise ValueError("%s is not a supported encoding" % encoding)
```
mirceaulinic, line 792:
```suggestion
            raise NotImplementedError("%s is not a supported encoding" % encoding)
```

mirceaulinic (commented):
Looks good to me, just requesting to raise `NotImplementedError` instead of `ValueError`.

mirceaulinic (approved):
Thanks @dasturiasArista!

PR #1643, opened 2022-05-20
eos: support arbitrary transport class

bewing:
Allow passing of a subclass of EapiConnection as the transport, either
as a class object or a string that can be imported as a class.


napalm/eos/eos.py
```diff
@@ -166,6 +162,26 @@ def _process_optional_args(self, optional_args):
                 if k in init_args and k not in filter_args
             }
 
+    def _parse_transport(self, transport):
```
mirceaulinic, line 165: Could you add some comments pls? This section has become so complex, I'm having a hard time to follow what & how we're trying to accomplish here... 😃

bewing: Rebased and added comments.  New changeset pushed.

mirceaulinic (approved):

PR #1644, opened 2022-05-20
remove tox in favor of gh actions/act

bewing:


mirceaulinic:
Is this what you're proposing to use for local testing? https://github.com/nektos/act I'm not familiar with it yet, but not opposed. We'll need to update the docs for this change though.

ktbyers:
With switch to GH actions, we ceased using Tox so we should just go ahead and remove it.

We can decide separately whether we want a local way to test GH actions behaviors.

mirceaulinic:
> With switch to GH actions, we ceased using Tox so we should just go ahead and remove it.
We can decide separately whether we want a local way to test GH actions behaviors.

I think I disagree with that. I'm looking at it from a user / contributor perspective: having tox around doesn't hurt, while silently changing (or dropping) the current way of testing locally without offering a (documented) perspective does hurt.

ktbyers:
Does Tox even work? At a minimum it is all out of date and doesn't actually correspond to our testing in GitHub actions.

For example, it is still testing Python 3.6 which will fail.

bewing:
> Is this what you're proposing to use for local testing? https://github.com/nektos/act I'm not familiar with it yet, but not opposed. We'll need to update the docs for this change though.

I think we can generally recommend `pytest` for local, single interpreter testing, and `act` for those who wish to execute the full GH actions suite prior to pushing.

mirceaulinic:
Fair enough. Would you be willing to add some docs around this?

bewing:
also, might want to update https://github.com/napalm-automation/napalm/blob/develop/.github/workflows/commit.yaml#L11-L13 to have 5 in parallel to run all the python tests in a single go (instead of 4 workers for 5 python versions)

ktbyers (approved):

ktbyers:
@bewing Don't know on the preview of doc rendering, but I would just say we push it and check it after the fact and then fix it if broke.

Sounds good on the workers increase.

PR #1647, opened 2022-05-24
Eos stateful

bewing:
In conjunction with #1643, if an arbitrary transport class is stateful, we should be trying to maintain state (exiting config mode after entering, closing when closing, etc).


bewing:
FWIW, this is related to me looking into EAPI Websockets to reduce TCP setup/teardown overhead when running large amounts of NAPALM commands against a single device.  Being able to maintain a single websocket connection while sending commands may be useful for others as well.

mirceaulinic (approved):

PR #1667, opened 2022-06-09
Improving iosxr get_facts behavior to handle XML changes in newer OS versions

ktbyers:


ktbyers:
@Kani999 So this only fixed `get_facts()` it doesn't fix anything else.

Can you see if this works on your failing device?

Kani999:
> @Kani999 What issue did you run into with the `get_interfaces` call and are you still running into it with this updated code here?

There are changes for `get_interfaces()` method, which works with IOS-XR 7.3.3
I have to look at the rest of the methods and find out all the problems.

ktbyers:
Okay, I will try to incorporate the get_interfaces() fixes (since get_facts() calls get_interfaces()).

Have you tried using the `iosxr_netconf` driver in NAPALM (i.e. does that driver work properly with your 7.3.3 code)?

Kani999:
> Okay, I will try to incorporate the get_interfaces() fixes (since get_facts() calls get_interfaces()).
> 
> Have you tried using the `iosxr_netconf` driver in NAPALM (i.e. does that driver work properly with your 7.3.3 code)?

`iosxr_netconf` works fine except `get_environment`. Which does not work with XML too.

I'll look at the `get_environment` method to get it to work via XML calls. Other [supported](https://napalm.readthedocs.io/en/latest/support/index.html) methods, should be working with the changes.

Kani999:
I've tested `get_environment` with `IOS XR Version 6.6.3` and it raises an exception too: 
```
napalm.pyIOSXR.exceptions.XMLCLIError:
Original call was: 
<?xml version="1.0" encoding="UTF-8"?><Request MajorVersion="1" MinorVersion="0">
<Get><AdminOperational><MemorySummary>                </MemorySummary></AdminOperational></Get></Request>
```

mirceaulinic (approved):

PR #1673, opened 2022-06-15
Handle modern memory output from Arista EOS 4.27+

### LINKED ISSUE #1671: get_environment on recent EOS crashes due to output changes

- [x] Yes
- [] No

### napalm version

```
napalm==3.4.1
nornir-napalm==0.1.2
```

```
Arista DCS-7050SX-64-R
Hardware version: 02.24
Serial number: XXXXXXXX
Hardware MAC address: XXXXXXXX
System MAC address: XXXXXXXX

Software image version: 4.27.4M
Architecture: i686
Internal build version: 4.27.4M-27115127.4274M
Internal build ID: 5c4f3083-0f5b-4afc-b983-d6fa00f6db6e
Image format version: 2.0
Image optimization: DEFAULT

Uptime: 1 week, 0 days, 18 hours and 54 minutes
Total memory: 3982552 kB
Free memory: 2473260 kB
```

### Error Traceback
(Paste the complete traceback of the exception between quotes below)

```
Host 'example.com': task 'napalm_get' failed with traceback:
  File "/home/brobert/.local/share/virtualenvs/netops-2cm1J9EX/lib/python3.6/site-packages/nornir/core/task.py", line 99, in start
    r = self.task(self, **self.params)
  File "/home/brobert/.local/share/virtualenvs/netops-2cm1J9EX/lib/python3.6/site-packages/nornir_napalm/plugins/tasks/napalm_get.py", line 44, in napalm_get
    result[g] = method(**options)
  File "/home/brobert/.local/share/virtualenvs/netops-2cm1J9EX/lib/python3.6/site-packages/napalm/eos/eos.py", line 741, in get_environment
    "available_ram": int(m.group("total")),
AttributeError: 'NoneType' object has no attribute 'group'
```

### Cause

The `show process top once` output has changed format slightly. Now instead of being an integer number of kilobytes, it's a decimal number of megabytes. The regex doesn't handle this well, and the code assumes the regex will match, and crashes when it doesn't.

```
#show processes top once
top - 11:53:56 up 7 days, 18:53,  2 users,  load average: 0.29, 0.16, 0.14
Tasks: 206 total,   1 running, 204 sleeping,   0 stopped,   1 zombie
%Cpu(s): 14.3 us,  8.6 sy,  0.0 ni, 77.1 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st
MiB Mem :   3889.2 total,    150.3 free,   1104.5 used,   2634.4 buff/cache
MiB Swap:      0.0 total,      0.0 free,      0.0 used.   2416.3 avail Mem 
```

### Potential fix

The regexes can be made to match by replacing `\d+` with `[0-9.]+` to match the decimal number. However in this case, where the line starts with `MiB Mem`, the numbers obtained will need to be multiplied by 1024 to convert from megabytes to kilobytes, and due to rounding errors, the conversion likely won't be completely accurate.

optiz0r:
Arista EOS 4.27+ (and possibly earlier versions, untested on
4.24-4.26) changes the output from `show processes top once`
command, rendering values decimal numbers of MB instead of integer
numbers of KB, and also reorders the fields.

This commit extends the `get_environment` function with additional
logic to handle the output from both older and newer OS versions.

The memory expressed in MB is converted into KB for backward
compatibility, but because the switch has done some rounding and
displays only a limited number of decimal places, the conversion
into KB will not be exact. There is nothing that Napalm can do
about this.

Fixes #1671

### DIFF
--- a/napalm/eos/eos.py
+++ b/napalm/eos/eos.py
@@ -846,16 +846,32 @@ class EOSDriver(NetworkDriver):
         # Matches either of
         # Mem:   3844356k total,  3763184k used,    81172k free,    16732k buffers ( 4.16 > )
         # KiB Mem:  32472080 total,  5697604 used, 26774476 free,   372052 buffers ( 4.16 < )
+        # MiB Mem :   3889.2 total,    150.3 free,   1104.5 used,   2634.4 buff/cache (4.27 < )
         mem_regex = (
-            r"[^\d]*(?P<total>\d+)[k\s]+total,"
-            r"\s+(?P<used>\d+)[k\s]+used,"
-            r"\s+(?P<free>\d+)[k\s]+free,.*"
+            r"^(?:(?P<unit>\S+)\s+)?Mem\s*:"
+            r"\s+(?P<total>[0-9.]+)[k\s]+total,"
+            r"(?:\s+(?P<used1>[0-9.]+)[k\s]+used,)?"
+            r"\s+(?P<free>[0-9.]+)[k\s]+free,"
+            r"(?:\s+(?P<used2>[0-9.]+)[k\s]+used,)?"
+            r".*"
         )
         m = re.match(mem_regex, cpu_lines[3])
-        environment_counters["memory"] = {
-            "available_ram": int(m.group("total")),
-            "used_ram": int(m.group("used")),
-        }
+
+        def _parse_memory(unit, total, used):
+            if unit == "MiB":
+                return {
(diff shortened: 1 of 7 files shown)


bewing:
Can you add a new test case to `test/eos/mocked_data/test_get_environment` as well?

optiz0r:
Test case added, and commit updated

PR #1701, opened 2022-07-12
IOSDriver should catch "Rollback aborted" case when using `configure replace`

rbcollins123:
Invalid candidate configs may silently fail installation currently because IOS-XE may attempt to rollback upon error but the rollback itself may be invalid.

Here is an example output from a Cat9300 switch on IOS-XE v17.3.3 when the intended config file attempts to make impossible changes to the running-config

```ncl-sw-03#configure replace flash:ncl-sw-03_render.cfg list force revert trigger error timer 60
Rollback Confirmed Change: Backing up current running config to flash:ncl-sw-03-Jul-12-23-02-40.132-UTC-25

!Pass 1
!List of Rollback Commands:
interface GigabitEthernet0/0
 no negotiation auto
interface GigabitEthernet0/0
 shutdown
  negotiation auto
end


!Pass 2
!List of Rollback Commands:
interface GigabitEthernet0/0
 no negotiation auto
interface GigabitEthernet0/0
 shutdown
  negotiation auto
end


!Pass 3
!List of Rollback Commands:
interface GigabitEthernet0/0
 no negotiation auto
interface GigabitEthernet0/0
 shutdown
  negotiation auto
end


!Pass 4
!List of Rollback Commands:
interface GigabitEthernet0/0
 no negotiation auto
interface GigabitEthernet0/0
 shutdown
  negotiation auto
end


!Pass 5
!List of Rollback Commands:
interface GigabitEthernet0/0
 no negotiation auto
interface GigabitEthernet0/0
 shutdown
  negotiation auto
end


The rollback configlet from the last pass is listed below:
********
!List of Rollback Commands:
interface GigabitEthernet0/0
 no negotiation auto
interface GigabitEthernet0/0
 shutdown
  negotiation auto
end
********


Rollback aborted after 5 passes```

The above should generate an exception within IOSDriver.commit_config

### CODE REVIEW

**Comment by ktbyers** [comment]
Do we need a different process here?

In this case applying the candidate config failed and the `on-box rollback` failed. 

I think we should automatically invoke a NAPALM rollback in this case as we are now in an invalid state (as far as the config). By `invalid` I mean it was neither the original configuration nor the intended candidate config.

Thoughts?

**Comment by ncsurfus** [comment]
> Do we need a different process here?
> 
> In this case applying the candidate config failed and the `on-box rollback` failed.
> 
> I think we should automatically invoke a NAPALM rollback in this case as we are now in an invalid state (as far as the config). By `invalid` I mean it was neither the original configuration nor the intended candidate config.
> 
> Thoughts?

Would the rollback process just end up issuing another `configure replace`, but back to the original configuration? How much different is that compared to what `configure replace` is doing during its rollback?

**Comment by ktbyers** [comment]
@ncsurfus Yes, it would try to do a `configure replace` again, but back to original configuration (which has been saved to flash).

I would expect this should work (otherwise, I would say it is an IOS/IOS-XE bug)...since the config that we are rolling back to is the running configuration before the original configure replace started.

We would have to make sure we can't get into a loop (i.e. if the NAPALM rollback also fails, then we raise an exception and tell the user).  We would also need to make sure here that if we fail again that we raise a clear message to the end-user.

**Comment by ncsurfus** [comment]
I did some manual testing for when this occurs, and yeah running a `configure replace bootflash:rollback_config.txt force` seemed to work pretty well to get it back to the original state (at least for my scenario).

**Comment by rbcollins123** [comment]
@ktbyers @ncsurfus I pulled the match for this case out into its own block and added a `self.rollback()` for now.  This isn't something I can readily test on a device at the moment, but if you want to drop this iteration of `ios.py` into your lab and try it @ncsurfus, let me know if it doesn't behave as planned.  It should net the same result as what you were doing by hand.

**Comment by ncsurfus** [comment]
@rbcollins123 @ktbyers I pulled in these changes and was able to test it. It worked great and I was able to get successful rollbacks across a couple different devices. I also verified with the command below and only got the expected difference.

```
xxxxxx#show archive config differences flash:rollback_config.txt
!Contextual Config Diffs:
+file prompt quiet

```

**Comment by rbcollins123** [comment]
Ok, let me know if there’s anything else I can do to help get this merged
in @ktbyers

**Comment by jifox** [comment]
Hi, is there a napalm 4 compatible "procurve" driver available?

When trying "napalm-procurve" I get the following error on `ndriver = get_network_driver("procurve")`.

```
raceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/ansible/.vscode-server/extensions/ms-python.python-2022.14.0/pythonFiles/lib/python/debugpy/adapter/../../debugpy/launcher/../../debugpy/__main__.py", line 39, in <module>
    cli.main()
  File "/home/ansible/.vscode-server/extensions/ms-python.python-2022.14.0/pythonFiles/lib/python/debugpy/adapter/../../debugpy/launcher/../../debugpy/../debugpy/server/cli.py", line 430, in main
    run()
  File "/home/ansible/.vscode-server/extensions/ms-python.python-2022.14.0/pythonFiles/lib/python/debugpy/adapter/../../debugpy/launcher/../../debugpy/../debugpy/server/cli.py", line 284, in run_file
    runpy.run_path(target, run_name="__main__")
  File "/home/ansible/.vscode-server/extensions/ms-python.python-2022.14.0/pythonFiles/lib/python/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_runpy.py", line 321, in run_path
    return _run_module_code(code, init_globals, run_name,
  File "/home/ansible/.vscode-server/extensions/ms-python.python-2022.14.0/pythonFiles/lib/python/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_runpy.py", line 135, in _run_module_code
    _run_code(code, mod_globals, init_globals,
  File "/home/ansible/.vscode-server/extensions/ms-python.python-2022.14.0/pythonFiles/lib/python/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_runpy.py", line 124, in _run_code
    exec(code, run_globals)
  File "/home/ansible/netbot/jupyter-notebooks/napalm_procurve.py", line 18, in <module>
    driver_pc = get_network_driver("procurve")
  File "/home/ansible/.cache/pypoetry/virtualenvs/jupyter-notebooks-gpTIvI8W-py3.8/lib/python3.8/site-packages/napalm/base/__init__.py", line 88, in get_network_driver
    module = importlib.import_module(module_name)
  File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 848, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/ansible/netbot/jupyter-notebooks/napalm_procurve.py", line 18, in <module>
    driver_pc = get_network_driver("procurve")
  File "/home/ansible/.cache/pypoetry/virtualenvs/jupyter-notebooks-gpTIvI8W-py3.8/lib/python3.8/site-packages/napalm/base/__init__.py", line 110, in get_network_driver
    raise ModuleImportError(
napalm.base.exceptions.ModuleImportError: No class inheriting "napalm.base.base.NetworkDriver" found in "napalm.procurve".
```

**Comment by rbcollins123** [comment]
Bump! Anything else I can do to help get this merged?

**Comment by ncsurfus** [comment]
I'm not sure this CR should be merged. "configure replace" uses the term rollback in a confusing way. "configure replace" was intended to _rollback_ to a previously saved running configuration, not deploy an entirely new configuration. For this reason "configure replace" uses the term rollback to indicate that it's applying the desired configuration. For instance, when "configure replace" is successful, it'll end with "Rollback Done". The terminology is confusing, but kind of makes sense if you understand the original use case.

"Rollback Aborted" also doesn't mean "configure replace" attempted to revert back to the original configuration and failed, but rather it appears to indicate that it's just aborting the deployment. The most common reason, from my testing, is that some of the desired configuration isn't showing up in the running configuration. For instance, if you include "no shutdown" on an interface, then "configure replace" will attempt to apply "no shutdown", and when it doesn't show up in the running configuration after 5 passes, then it "aborts". Some things like SNMPv3 users don't show up in the running configuration and can also cause this output. From my testing, this is generally harmless, but I'm sure there are scenarios where this could break something.

**Comment by mirceaulinic** [comment]
Code-wise, this would kinda make sense to me. But I can't claim I understand what `rollback aborted` actually means, and based on the comment above, it sounds like the message is rather counter-intuitive.
@ktbyers would you please advise?

**Comment by ktbyers** [comment]
@mirceaulinic @rbcollins123 @ncsurfus I think we should leave the current solution and just close this PR (as per the comments from @ncsurfus). Basically, it sounds like we might be creating larger problems by attempting to auto-rollback on this.

I am definitely open to discussions on this, however.

**Comment by rbcollins123** [comment]
@ktbyers I don't care if the napalm-level-rollback happens or not, that was only added at your request.  My only concern in raising the PR originally was to ensure an exception is raised when this occurs since the config-replace has failed in this scenario.  If the `self.rollback()` is the only point of contention, we can just revert back to [the original commit on the PR](https://github.com/napalm-automation/napalm/pull/1701/commits/41c652f3e752aa8ba0404c2759c22c6aab2380a0) to generate the exception and merge that.  It would be nice to not have to locally patch our installs to add the exception at some point.

**Comment by ktbyers** [comment]
@rbcollins123 (@ncsurfus) is pointing out that `rollback aborted` is relatively normal behavior (meaning it is just a dumb string parser and it is looking for certain text in the output and if the text doesn't show up then we will get this `rollback aborted`). This could easily happen due to default values (like shutdown/no shutdown).

So I would have the same issue with raising an exception (i.e. we would be causing working config files to fail).

@rbcollins123 Was this your test merge config to demonstrate the issue:

```
********
!List of Rollback Commands:
interface GigabitEthernet0/0
 no negotiation auto
interface GigabitEthernet0/0
 shutdown
  negotiation auto
end
********
```

If not, can you provide a test case on IOS/IOS-XE?

rbcollins123:
Yes that was an IOS-XE example and real device test case to force and show
the behavior.  Yes that example is a benign but not all cases are. We have
seen similar happen on NXOS cases in real world too.

PR #1806, opened 2022-11-28
Fix issue in lldp neighbors discovery for IOS devices.

CarloMara:
Hello everyone,

The parsing logic incorrectly assumes that `remote_chassis_id` and `remote_port` will always contain MAC addresses. According to the spec this is not always the case.

This pr adds the ability to handle uncommon cases, and a testcase to show the issue.
The data in the testcase was taken from a production device and anonymized (mac addresses and hostnames changed).

Thanks again for making Napalm, it's a godsend.

Cheers,
Carlo

### DIFF
--- a/napalm/ios/ios.py
+++ b/napalm/ios/ios.py
@@ -44,6 +44,7 @@ from napalm.base.helpers import (
     generate_regex_or,
     sanitize_configs,
 )
+from netaddr.core import AddrFormatError
 from napalm.base.netmiko_helpers import netmiko_args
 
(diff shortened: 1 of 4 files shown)


CarloMara:
Hola @mirceaulinic,

sorry for the direct tag, is there something I can do/missing to get this merged?

Thanks a lot, 
Carlo

ktbyers:
Superseded by: https://github.com/napalm-automation/napalm/pull/1849

Issue #1372, opened 2021-02-11
nxos get_environment does not parse power supply output with alternate key name


kaibabbob:
### Description of Issue/Question

Some Nexus models or NXOS versions have a different key for the power load.  The function is expecting 'watts' but some switches return the key 'actual_out' instead.

I propose the following update to the  _process_pdus() function in nxos.py

'''
<                 normalized[psinfo["psnum"]]["output"] = float(psinfo.get("watts", -1.0))
---
>                 # Some models or versions return 'watts' as the key and others 'actual_out'
>                 if psinfo.get("watts") != None:
>                     normalized[psinfo["psnum"]]["output"] = float(psinfo.get("watts", -1.0))
>                 else:
>                     normalized[psinfo["psnum"]]["output"] = float(psinfo.get("actual_out", -1.0).split()[0])
'''

### Did you follow the steps from https://github.com/napalm-automation/napalm#faq
(Place an ``x`` between the square brackets where applicable)

- [x] Yes
- [] No

### napalm version
(Paste verbatim output from `pip freeze | grep napalm` between quotes below)

```
napalm==3.2.0
napalm-panos==0.5.2

```

### Network operating system version
(Paste verbatim output from `show version` - or equivalent - between quotes below)

```
Cisco Nexus Operating System (NX-OS) Software
TAC support: http://www.cisco.com/tac
Documents: http://www.cisco.com/en/US/products/ps9372/tsd_products_support_series_home.html
Copyright (c) 2002-2019, Cisco Systems, Inc. All rights reserved.
The copyrights to certain works contained in this software are
owned by other third parties and used and distributed under
license. Certain components of this software are licensed under
the GNU General Public License (GPL) version 2.0 or the GNU
Lesser General Public License (LGPL) Version 2.1. A copy of each
such license is available at
http://www.opensource.org/licenses/gpl-2.0.php and
http://www.opensource.org/licenses/lgpl-2.1.php

Software
  BIOS:      version 2.12.0
  kickstart: version 7.3(5)D1(1)
  system:    version 7.3(5)D1(1)
  BIOS compile time:       05/29/2013
  kickstart image file is: bootflash:///n7000-s2-kickstart.7.3.5.D1.1.bin
  kickstart compile time:  9/30/2019 21:00:00 [11/09/2019 05:55:30]
  system image file is:    bootflash:///n7000-s2-dk9.7.3.5.D1.1.bin
[...]

PR #1809, opened 2022-12-06
Handle non standard output for NXOS PSU's in get_environment

tcaiazza:
These minor changes are to address the issues that were reported in https://github.com/napalm-automation/napalm/issues/1371 and https://github.com/napalm-automation/napalm/issues/1372


mirceaulinic (approved):
Thanks @tcaiazza! 🙇🏼

PR #1813, opened 2022-12-13
Add tolerance percentage to validate

sjhloco:
This pr adds the ability to validate based on a percentage tolerance either side of the specified value (useful for BGP prefixes). I was trying to keep it as close to existing operators (`"<d"`, `">d"`, `"d<->d"`), in the end came up with `"t%d"`. The logic behind it is that a tolerance of *"t percent of d"* is allowed either side of the value of *d*. For example `"10%100"` would be 10% of 100, so a value between 90 and 110 is accepted. I didnt really want to use a % sign but couldn't think of a better way to represent this. Have done tests and docs that I could find, let me know if need anything else.


mirceaulinic (approved):
Very cool!

PR #1849, opened 2023-02-16
Fix ios lldp discovery (rebase)

ktbyers:


ktbyers:
FYI @CarloMara

ktbyers:
@CarloMara Let me know if you see any issues:

FYI, I moved this port normalization to always happen:

```python
                # If port is a mac-address, normalize it.
                # The MAC helper library will normalize "15" to "00:00:00:00:00:0F"
                if port.count(":") == 5 or port.count("-") == 5 or port.count(".") == 2:
                    try:
                        port = napalm.base.helpers.mac(port)
                    except AddrFormatError:
                        pass
```

I also had to put the conditional as some port identifiers like "15" would incorrectly try to get converted to MAC addresses.

I didn't see a good reason why the port MAC normalization should only happen in the `if not hostname` case. 

Realistically, the NAPALM `mac` helper function should probably have more logic in it (to avoid incorrect MAC conversions) and my quick check shouldn't be needed.

PR #1850, opened 2023-02-16
Pylama was broke and not actually executing pycodestyle

ktbyers:


napalm/eos/eos.py
```diff
@@ -452,7 +452,7 @@ def _load_config(self, filename=None, config=None, replace=True):
             return None
 
         try:
-            if not any(l == "end" for l in commands):
+            if not any(cmd == "end" for cmd in commands):
```
ktbyers, line 455: This was a linter failure...it didn't like the single-letter variable.

PR #1879, opened 2023-03-23
Update get_bgp_config to clarify return values

### LINKED ISSUE #1878: `get_bgp_config` should return a canonical BGP ASN for the router
Per discussion in Slack:

`get_bgp_config` does not explicitly tell you what the configured BGP ASN is for the main BGP process.  The different device drivers have conflicting methods for setting the ASN in the "no-group" neighbor group `_`.

We should clarify in the documentation that the neighbor group `_` should ALWAYS exist if BGP is running on a router, and that the configured ASN for the BGP process be stored in that group's `local_as` property.

bewing:
Closes #1878

This changeset also includes logic to set the `local_as` in the group and neighbor dictionaries in the return value to the actual local AS instead of using `0` as a value to indicate it was not overridden


test/junos/mocked_data/test_get_bgp_config/normal/expected_result.json
```diff
@@ -1 +1,86 @@
...
+      "F",
+      "I",
+      "X",
+      "-",
+      "L",
+      "I",
+      "M",
+      "I",
+      "T"
+    ],
```
bewing, line 79: this absolutely looks like a bug, but I am just reformatting/refactoring the file here.

ktbyers: Yeah that is a bug...looks like we have a case where we expected it to be a list or a dict and looped over it (and it ended up being a string instead).

ktbyers: Do we need to create a separate issue for this?

ktbyers (approved):
LGTM. I tested it on a Juniper device with a simple configuration and it looked correct. I didn't test it on any of the others, but looked through the changes for all of them.

PR #1887, opened 2023-03-28
Fixing parsing error when there is no BGP configured on Junos

ktbyers:
Currently if there is no BGP configuration on a Junos device you will get an exception when calling `get_bgp_config`


bewing:
LGTM

PR #1890, opened 2023-03-29
Implement no BGP config behavior for Junos (get_bgp_config)

ktbyers:
I tested empty behavior `get_bgp_config` returns {}

I also tested `get_bgp_config(group="_") returns {} when no BGP is configured.

I still need to test behavior of `group="_")` when BGP is configured.

And `group="something"` behavior (both when doesn't exist and when configured).


ktbyers:
Expected behavior (per communication with @bewing).

Call to `get_bgp_config()` by default will return all BGP peer groups and the "_" (no peer group configuration).

If BGP is not configured on the device, then {} will be returned.

If `get_bgp_config(group="x")` is called then only that peer group will be returned. You can also call `group="_"` and then only the no-peer-group will be returned. If a given group is not configured or if there is no BGP config, then {} will be returned for `get_bgp_config(group="x"` calls.

bewing (approved):
LGTM

PR #1910, opened 2023-04-28
get_bgp_config for Cisco when no BGP is configured

ktbyers:
Tested with no BGP config.

Tested with group="_" both with and without any BGP config.

Tested with group="bogus" with no BGP config.


bewing (approved):
LGTM

PR #1913, opened 2023-05-06
Fix documentation of getter support matrix

ogenstad:
Replace pytest-json with pytest-json-report and make modifications to support new format.


ktbyers:
LGTM

PR #1921, opened 2023-05-15
fix: eos config_session not cleared if discard failed

kpetremann:
If the abort command is refused, for instance because of a TACACS issue, the config_session is never cleared because an exception is raised by self._run_commands.

This is an issue in Saltstack case:
- a pending session could contain incomplete changes
- this pending session is resumed on the next load_config call causing messed up configuration

This change ensure the config_session is always cleared.

### DIFF
--- a/docs/support/index.rst
+++ b/docs/support/index.rst
@@ -141,6 +141,7 @@ ____________________________________
 * :code:`eos_fn0039_config` (eos) - Transform old style configuration to the new style, available beginning with EOS release 4.23.0, as per FN 0039. Beware
   that enabling this option will change the configuration you're loading through NAPALM. Default: ``False`` (won't change your configuration commands).
   .. versionadded:: 3.0.1
+* :code:`force_cfg_session_invalid` (eos) - Force the config_session to be cleared in case of issues, like `discard_config` failure. (default: ``False``)
 
 The transport argument
 ______________________
--- a/napalm/eos/eos.py
+++ b/napalm/eos/eos.py
@@ -543,9 +543,10 @@ class EOSDriver(NetworkDriver):
     def discard_config(self):
         """Implementation of NAPALM method discard_config."""
         if self.config_session is not None:
-            commands = [f"configure session {self.config_session} abort"]
-            self._run_commands(commands, encoding="text")
+            config_session = self.config_session
             self.config_session = None
+            commands = [f"configure session {config_session} abort"]
+            self._run_commands(commands, encoding="text")
 
     def rollback(self):
         """Implementation of NAPALM method rollback."""
--- a/napalm/eos/eos.py
+++ b/napalm/eos/eos.py
@@ -96,6 +96,8 @@ class EOSDriver(NetworkDriver):
         Optional args:
             * lock_disable (True/False): force configuration lock to be disabled (for external lock
                 management).
+            * force_cfg_session_invalid (True/False): force invalidation of the config session
+                in case of failure.
             * enable_password (True/False): Enable password for privilege elevation
             * eos_autoComplete (True/False): Allow for shortening of cli commands
             * transport (string): transport, eos_transport is a fallback for compatibility.
@@ -134,6 +136,10 @@ class EOSDriver(NetworkDriver):
         # Define locking method
         self.lock_disable = self.optional_args.pop("lock_disable", False)
 
+        self.force_cfg_session_invalid = self.optional_args.pop(
+            "force_cfg_session_invalid", False
+        )
+
         # eos_transport is there for backwards compatibility, transport is the preferred method
         transport = self.optional_args.get(
             "transport", self.optional_args.get("eos_transport", "https")
@@ -150,7 +156,6 @@ class EOSDriver(NetworkDriver):
         self.netmiko_optional_args = netmiko_args(optional_args)
 
     def _process_optional_args_eapi(self, optional_args):
-
         # Parse pyeapi transport class
         self.transport_class = self._parse_transport(self.transport)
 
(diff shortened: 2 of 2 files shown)


bewing:
Wouldn't the session still be in use, causing future issues due to the pending session preventing a config lock?

kpetremann:
hello @bewing,
Thanks for the quick response.

Yes totally, but "only" for 24 hours as EOS will cancel any pending changes older than that.
It would make it closer to the lock implementation for JunOS:
https://github.com/napalm-automation/napalm/blob/cb4845c40017a3e8e3c2b4e57838de2fd31a1348/napalm/junos/junos.py#L159-L166

It seems better to lock for 24 hours rather than having unexpected pending changes which can be applied by any future Salt state in uncontrolled manner. Knowing that such fails should not occur regularly and should be considered as incident. In our case, it was a TACACS issue which occurred during the changes in the config session (EOS checks permissions progressively for each line typed in the config session). We had multiple occurrences of this problem causing a mess in the configuration.

We could make this behavior optional adding a parameter at the init of the EOSDriver.

The other alternatives I see:
* calling `discard_commit` each time we are about to do a change => not scalable as one could easily forget to do that in a Salt state and ending up with the same issue
* add an optional alternative of the lock itself, depending on a new EOSDriver attribute, to have a kind of "private configuration" mode.

WDYT?

kpetremann:
Hello @ktbyers,

> Why doesn't the exception get detected by Salt?

When Salt proxy-minion are setup with Napalm as a driver, it uses the Napalm library to be vendor agnostic.
Even if Salt was catching the exception when it tried to discard the pending config, the only thing that Salt could do would be very specific to EOS, for instance by clearing manually the `device.config_session`.

The only generic alternative I see would be to do change in Salt to make it completely drop the connection when there is an error, which seems overkill.

> Also from you stated above, this implies Salt is reusing a NAPALM connection over very long periods of time--is that correct?

Yes, proxy-minions are running as a "daemon" and keep the connection active (by default, but there is an option reconnect at each request).
The proxy-minion keeps the Napalm Device object to reuse connection. It also keeps the config_session until the change is actually committed. This is to be able to load configuration without committing, then complete the config and then commit.
At least from what we can deduce from their documentation:
* [net.load_config](https://docs.saltproject.io/en/latest/ref/modules/all/salt.modules.napalm_network.html#salt.modules.napalm_network.load_config) can have commit=False
* [net.commit](https://docs.saltproject.io/en/latest/ref/modules/all/salt.modules.napalm_network.html#salt.modules.napalm_network.commit) permits to commit pending changes

Note: we are not using commit=False, this is only our understanding of Saltstack implementation of Napalm based proxy-minions.

ktbyers:
I don't like the solution proposed as it is an unexpected thing to do: `exception happens on discard() and then hide the configuration session` (even though the config session still exists and is active).

Given Salt's very long running sessions (which are atypical) does it make more sense for Salt's logic to bake more checks in here (for example, call discard() and then ensure there are no config sessions at the end of that call, or call discard() and then do a compare_config() after that to ensure there are no pending changes)?

kpetremann:
From user perspective, Napalm have inconsistent behavior at least when comparing EOS and JunOS.
In JunOS, if the discard does not work, the lock is not unlocked. This is the opposite with EOS implementation.

> exception happens on discard() and then hide the configuration session (even though the config session still exists and is active).

It is not really hidden as the lock would block this. Also, even if the session still exists, it is more pending than active. I mean, if you create another session, the pending changes are not resumed. It is like private config session on JunOS.

> I don't like the solution proposed

What about the other propositions I made:
* make this change an option => something like always_clear_session_on_discard (needs a better name)
* or, add an optional alternative of the lock itself, depending on a new EOSDriver attribute, to have a kind of "private configuration" mode. Meaning having the exact same lock mechanism than JunOS: if there is a pending session => lock, even if the config_session is the one from self.config_session.

kpetremann:
>  does it make more sense for Salt's logic to bake more checks in here (for example, call discard() and then ensure there are no config sessions at the end of that call, or call discard() and then do a compare_config() after that to ensure there are no pending changes)?

In the TACACS issue we had, the has_pending_commit/compare_config would trigger an exception as it requires permissions as well so it cannot be done at the end.
The only solution would be to add an option in Salt to check if there a pending session before calling load_config. But again, this is only for EOS...

ktbyers:
@kpetremann So on Juniper we have two locking modes: a device lock (which locks/unlocks at the commit/discard level) and a session lock (session_config_lock).

On Arista we really only have a session_config_lock (i.e. the config_session is the lock). If you have this session id, then you possess the lock.

So your issue is, you are constantly reusing the config_session so you constantly have the lock. Consequently, when something fails like discard_config() you can get into potentially a bad state (i.e. there are still stale configuration changes hanging around in the candidate_config).

Is this an accurate description of the issue/problem? I am just trying to make sure I am properly understanding this.

kpetremann:
yes, exactly.

ktbyers:
Yeah, maybe this is the best middle ground:

```
make this change an option => something like always_clear_session_on_discard (needs a better name)
```

Something like `cfg_session_invalidate` or `force_cfg_session_invalid`?

We probably should not tie it to `discard` (in case there are other scenarios where we need to do this invalidation)?

kpetremann:
sounds great :) I'll update the PR
thanks

kpetremann:
@ktbyers, PR updated, please let me know if it needs further changes

ktbyers (approved):

ktbyers:
Some slight changes here, but should have the same net-effect:

https://github.com/napalm-automation/napalm/pull/1927

PR #1926, opened 2023-05-19
Implement No BGP for IOSXR-netconf and get_bgp_config()

ktbyers:


ktbyers:
I also cleaned up a Python / LXML warning where it didn't like how the code was checking for a boolean status (and switched it to the pattern the warning recommended)

PR #1927, opened 2023-05-19
EOS add force_cfg_session_invalid argument

ktbyers:
This allows the `self.config_session` to be invalidated on certain failures (right now a `discard_config()` failure).


ktbyers:
CC: @kpetremann @bewing 

Slight changes to the previous solution.

Do you see any issues?

kpetremann:
@ktbyers sure :) actually I was hesitating with try except approach. I just tend to choose the solution with less, but that is only code style :grin:

kpetremann:
with less indentation*

ktbyers:
Yeah, it is six of one half dozen of the other...we thought this showed in the code when this case would happen a bit clearer (i.e. it is when the discard fails that we need to potentially do this override).

kpetremann:
> we thought this showed in the code when this case would happen a bit clearer (i.e. it is when the discard fails that we need to potentially do this override).

you are totally right, it is more obvious that way.
Thanks

PR #1931, opened 2023-05-23
Fix `Can't convert with callable None` - for junos routers

teresalves:
The code keeps breaking on this line: https://github.com/napalm-automation/napalm/blob/develop/napalm/junos/junos.py#L1339 
when there is a key that is not included in the [_GROUP_FIELDS_DATATYPE_MAP_](https://github.com/napalm-automation/napalm/blob/develop/napalm/junos/junos.py#L1247)
In particular, for the group names "multihop" and "cluster" this is a problem because it will inevitably show up in the router configs, but the datatype will always be None, which will provoke this error:
ValueError: Can't convert with callable None - no default is defined for this type.
originating from [here](https://github.com/napalm-automation/napalm/blob/develop/napalm/base/helpers.py#L471)

In previous versions, this `convert` function was not doing the type checks, so if it was just returning None. The value for this case is irrelevant because both are deleted from the config further ([here](https://github.com/napalm-automation/napalm/blob/develop/napalm/junos/junos.py#L1358) and [here](https://github.com/napalm-automation/napalm/blob/develop/napalm/junos/junos.py#L1438)), respectively. Their existence in the config is temporary and not necessary - it is only used to produce/update other values ([here](https://github.com/napalm-automation/napalm/blob/develop/napalm/junos/junos.py#L1359) and [here](https://github.com/napalm-automation/napalm/blob/develop/napalm/junos/junos.py#L1414))

Therefore, I added here a conditional clause that never adds them to the final bgp config, so there is no need to delete it further ahead. Instead, a separate object is created for validation.

Issue: https://github.com/napalm-automation/napalm/issues/1932

### DIFF
--- a/napalm/junos/junos.py
+++ b/napalm/junos/junos.py
@@ -1246,6 +1246,8 @@ class JunOSDriver(NetworkDriver):
         }
         _GROUP_FIELDS_DATATYPE_MAP_.update(_COMMON_FIELDS_DATATYPE_)
 
+        _UNWANTED_GROUP_FIELDS = ["multihop", "cluster"]
+
         _DATATYPE_DEFAULT_ = {str: "", int: 0, bool: False, list: []}
 
         bgp_config = {}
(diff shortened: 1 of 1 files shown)


mirceaulinic:
@teresalves sorry for late review, would you mind updating your branch? Thanks.

teresalves:
Hey @mirceaulinic ! Thanks for taking a look
Updated now, could you recheck please?

mirceaulinic (approved):

PR #1940, opened 2023-05-23
Update black to a newer version

ktbyers:


ktbyers:
I reviewed all of these Python black changes. They all looked fine.

PR #1942, opened 2023-05-24
Optimize Dockerfile

duhow:
- Properly mount git repository from a `bind mount`, this ensures content is deleted after image creation.
- Update Python from 3.6 -> 3.9
- Pull `python3-dev` instead of `python-dev` (2.7)
- Add `ENTRYPOINT` to run command `napalm` directly.

NOTE: This requires `docker buildx` to build as per bind mount.


Dockerfile
```diff
@@ -1,10 +1,16 @@
...
+FROM scratch AS installer
 COPY ./ /var/cache/napalm/
 
-RUN apt-get update \
- && apt-get install -y python-dev python-cffi libxslt1-dev libssl-dev libffi-dev \
- && apt-get autoremove \
- && rm -rf /var/lib/apt/lists/* \
- && pip --no-cache-dir install -U cffi cryptography /var/cache/napalm/ \
- && rm -rf /var/cache/napalm/
+FROM python:3.9-slim
```
mirceaulinic, line 6:
```suggestion
FROM python:3.12-slim-bookworm
```

The reason why it's a good idea to specify the distro (as above) is because `3.9-slim` currently defaults to Bookworm and therefore the build fails because `python3.9-dev isn't available`:

```
#11 2.845 Building dependency tree...
#11 2.954 Reading state information...
#11 3.051 E: Unable to locate package python3.9-dev
#11 3.051 E: Couldn't find any package by glob 'python3.9-dev'
#11 3.051 E: Couldn't find any package by regex 'python3.9-dev'
------
```

We can probably bump to Python 3.12 which NAPALM is now compatible with.

Dockerfile
```diff
@@ -1,10 +1,16 @@
...
+
+RUN --mount=type=bind,from=installer,source=/var/cache/napalm,target=/var/cache/napalm,rw \
+    apt-get update && \
+    apt-get install -y \
+        python3.9-dev libxslt1-dev libssl-dev libffi-dev && \
```
mirceaulinic, line 11:
```suggestion
        python3-dev libxslt1-dev libssl-dev libffi-dev && \
```

I'd just leave the Python version to be decided by the base image.

mirceaulinic (commented):
Looks good overall, thansk for this PR @duhow!

I left some comments and I've also suggest adding a Dockerignore file on this occasion to avoid sending all the massive build contex (but only what's necessary, `napalm/` dir, setup.cfg or requirements files).

Issue #1952, opened 2023-05-28
eos: Rollback should revert to the running config, not the startup config.


ichilton:
### Description of Issue/Question

The code in the commit_config method of the eos driver uses the following commands:

```
            commands = [
                "copy startup-config flash:rollback-0",
                "configure session {} commit".format(self.config_session),
                "write memory",
            ]
```

src: https://github.com/napalm-automation/napalm/blob/50ab9f73a2afd8c84c430e5d844e570f28adc917/napalm/eos/eos.py#L518

I would argue this is incorrect. It should save the running configuration, not the startup configuration.

There is no guarantee that the configuration was recently saved/written and therefore the running config could be completely different from the startup config.

I believe it should either save the running config for rollback or do a write before the change, like:

```
            commands = [
                "write memory",
                "copy startup-config flash:rollback-0",
                "configure session {} commit".format(self.config_session),
                "write memory",
            ]
```

### napalm version

```
napalm==4.1.0
```

### Network operating system version

```
4.29.3M
```

ktbyers:
NAPALM on commit_config() should always save the running-config to the `startup-config` and the startup-config is the permanent config (i.e. is the "durable" state of things wrt ACID transactions i.e. it is what the system will revert to in a reboot event).

So I think NAPALM is doing this correctly.

```
There is no guarantee that the configuration was recently saved/written and therefore the running config could be completely different from the startup config.
```

You can always separately call `write mem` (if you are not confident in your sync state between running-config and startup-config).

If changes came from outside NAPALM, I don't think it is correct for NAPALM to just sync them i.e. the person or process that made the changes should ensure they are synced.

PR #1953, opened 2023-05-28
eos: Save running-config for rollback instead of startup-config (fixes issue #1952).

ichilton:
As per [issue](https://github.com/napalm-automation/napalm/issues/1952), I believe it should save the running config for rollback, not the startup config, which could be different to the running config if changes have been made since the last `write memory`.


ktbyers:
See comments on issue #1952

PR #1966, opened 2023-06-16
Use `LastStateTransitionTime` to populate `last_flapped` for `get_interfaces()` in IOS-XR

jlixfeld:


jlixfeld:
Currently, `last_flapped` for `get_interfaces()` is hard coded to `-1`.

In at least IOS-XR 7.1.2, `Last link flapped` from `show interface` is exposed by `<Get><Operational><Interfaces></Interfaces></Operational></Get>` as `LastStateTransitionTime`.

This can be used to populate `last_flapped`, with `-1` being a default for backwards compatibility where `LastStateTransitionTime` is not exposed.

mirceaulinic (approved):

PR #2006, opened 2023-09-06
Closes #2005: Change rollback behavior

### LINKED ISSUE #2005: Change rollback behavior
Based on [this discussion](https://github.com/napalm-automation/napalm/discussions/2001).

Currently, calling `rollback()` on NXOS does nothing unless napalm is aware of making changes. However, in some cases it makes sense to call `rollback()` in a new napalm session.

Proposed behavior is that `rollback()` will become agnostic to any changes napalm is aware of, and will have the net effect of reverting to the most recent state. For platforms with a native rollback feature, this will revert to the most recent snapshot. Platforms that don't, such as IOS, will attempt to revert to the rollback state that napalm created in a previous session. If such a rollback state does not exist, an exception will raise.

decoupca:
Changes rollback behavior so that calling `rollback()` attempts to roll back to the most recent config snapshot, regardless of whether napalm is aware of making any changes in the current session. For platforms that use file-based snapshots, ensures the rollback file exists on the remote system before attempting rollback.

JUNOS, IOSXR and EOS all behaved this way before this PR -- no changes made.

~~IOS: added check to ensure rollback file exists before attempting.~~ IOS already has a file exists check in place for rollbacks

NXOS and NXOS-SSH: Removed check that prevented rollbacks if napalm is not aware of changes made. Also implemented `_check_file_exists` and added checks to ensure `rollback_cfg` exists prior to rollback attempt.


ktbyers:
Related issue and discussion:

https://github.com/napalm-automation/napalm/issues/2005

ktbyers:
@decoupca Is there a separate change for Cisco IOS?

```
IOS: added check to ensure rollback file exists before attempting.
```

decoupca:
> @decoupca Is there a separate change for Cisco IOS?
> 
> ```
> IOS: added check to ensure rollback file exists before attempting.
> ```

No, sounded like we wanted to make the changes in one pass but I can separate it out if you like.

napalm/nxos/nxos.py
```diff
@@ -918,10 +942,12 @@ def _load_cfg_from_checkpoint(self) -> None:
 
     def rollback(self) -> None:
         assert isinstance(self.device, NXOSDevice)
-        if self.changed:
-            self.device.rollback(self.rollback_cfg)
-            self._copy_run_start()
-            self.changed = False
+        if not self._check_file_exists(cfg_file=self.rollback_cfg):
+            msg = f"Rollback file '{self.rollback_cfg}' does not exist on device."
+            raise ReplaceConfigException(msg)
```
ktbyers, line 947: This might not be the right exception i.e. what are the cases where `rollback` gets called. We probably need to look at that a bit more. I generally view a `ReplaceConfigException` as an exception you get when you try to do the replace config operation.

I guess we could raise a generic NapalmException

ktbyers:
@decoupca There aren't any IOS changes in this pull-request though? Was there another file/change that needed included?

decoupca:
> @decoupca There aren't any IOS changes in this pull-request though? Was there another file/change that needed included?

Oh my mistake -- I didn't change anything, that check was already built in to the IOS driver. Only NXOS and NXOS_SSH have changed to match the IOS/JUNOS behavior.

napalm/nxos/nxos.py
decoupca: I copied `ReplaceConfigException` from other code that raises that exception during a rollback. I thought about raising `NapalmException` or creating a new `RollbackConfigException`, but I don't have strong preferences here.

napalm/nxos/nxos.py
```diff
@@ -191,6 +191,30 @@ def _send_command(
...
+        116526366720 bytes total
+        """
+        cmd = f"dir {cfg_file}"
+        output = self._send_command(command=cmd, raw_text=True)
+        if "No such file or directory" in output:
+            return False
+        elif cfg_file in output:
+            return True
+        return False
+
```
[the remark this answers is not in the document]
decoupca: Good catch, pushed an update. Based on all my tests, if `self._send_command()` doesn't raise an exception and 'No such file or directory' is not in the output, then the file exists (i.e., there's no need to check for anything in the output)

ktbyers:
No worries...I just misunderstood (didn't check) what you had mentioned above.

napalm/nxos/nxos.py
ktbyers: Okay, let me look some more. I only looked at it quickly (it could make the most sense given what we have previously done).

ktbyers (approved):
LGTM

PR #2015, opened 2023-10-16
eos: Require EOS 4.23+

bewing:


ktbyers (approved):
LGTM

ktbyers:
@bewing Good job.

We should probably roll the NAPALM major version when we release this--since it is a pretty big change. What do you think?

bewing:
> We should probably roll the NAPALM major version when we release this--since it is a pretty big change. What do you think?

Agreed.  Do we need to hold merging this until we think we're ready for a major release?  I am in no particular rush.

PR #2016, opened 2023-10-18
NXOS_SSH/IOS XE IPv6 Regex update

ubaumann:
- The regex did not match for IPv6 addresses ending or starting with `::` (IOS XE and NXOS_SSH)
- On NXOS, the `show bgp all summary vrf all` with IPv6 addresses and long AS numbers has two newlines.


ubaumann:
I tested the long AS number with IPv4 and the same problem occurs. Therefore and for readability I have changed the code to make two `re.sub`

ktbyers:
@ubaumann Are there tests for the long as-number case?

ubaumann:
@ktbyers The long as-number is combined with the IPv6 test case

```
2001:db8:e0:df::
                4    12345678
                         327465  327268      145    0    0     3w1d 4
```
https://github.com/napalm-automation/napalm/pull/2016/files#diff-cea37f4186bd9fbcbb95c704fda5ea3b17ecad55d1d50ee605d55aab70ef84fbR29

I did not write test for IPv4 with long AS number but because the two `re.sub` the case with ipv4 and ipv6 is the same.
I also did not write tests in the IOS XE for the IPv6 addresses ending or starting with `::`

Should I add some more testing?

ktbyers:
That is probably fine...I missed the long-number AS earlier in the test output.

ktbyers (approved):

PR #2112, opened 2024-07-08
EOS bugfix - newer EOS reports RPKI validation info

jmaslak:
In newer versions of EOS, when `show ip bgp neighbor` (and similar IPv6 command) additional information is reported for RPKI.

For instance:
```
BGP neighbor is 192.0.2.1, remote AS 65001, external link
 Description: An Upstream
  BGP version 4, remote router ID 192.0.2.1, VRF default
  Inherits configuration from and member of peer-group FOO
  Last read 00:00:02, last write 00:00:14
  Hold time is 90, keepalive interval is 30 seconds
  Configured hold time is 180, keepalive interval is 60 seconds
  Effective minimum hold time is 3 seconds
  Send failure hold time is 0 seconds
  Hold timer is active, time left: 00:01:28
  Keepalive timer is active, time left: 00:00:09
  Connect timer is inactive
  Configured idle-restart time is 300 seconds
  Idle-restart timer is inactive
  BGP state is Established, up for 48d12h
  Number of transitions to established: 1
  Last state was OpenConfirm
  Last event was ReapplyInboundPolicy
  Last sent notification:Cease/connection collision resolution, Last time 48d12h
  Types of communities advertised: standard extended large
  Neighbor Capabilities:
    Multiprotocol IPv4 Unicast: advertised and received and negotiated
    Four Octet ASN: advertised and received and negotiated
    Route Refresh: advertised and received and negotiated
    Enhanced route refresh: advertised
    Send End-of-RIB messages: advertised and received and negotiated
    Additional-paths recv capability:
      IPv4 Unicast: advertised
    Additional-paths send capability:
  Restart timer is inactive
  End of rib timer is inactive
    IPv4 Unicast End-of-RIB received: Yes
      Received 48d12h
      Number of stale paths removed after graceful restart: 0
      Number of paths received before End-of-RIB: 363
  AIGP attribute send and receive for IPv4 Unicast are disabled
  AIGP attribute send and receive for IPv4 with MPLS Labels are disabled
  AIGP attribute send and receive for IPv6 Unicast are disabled
  AIGP attribute send and receive for IPv6 with MPLS Labels are disabled
  BGP session driven failover for IPv6 Unicast is disabled
  BGP session driven failover for IPv4 Unicast is disabled
  Message Statistics:
                         Sent      Rcvd
    Opens:                  1         2
    Notifications:          1         0
    Updates:               30   8489393
    Keepalives:        164001    139688
    Route Refresh:          0         0
    Total messages:    164033   8629083
[...]

PR #2116, opened 2024-07-15
Rework `eos.get_bgp_neighbor`

bewing:
Use all JSON output for `get_bgp_neighbor` to remove any text-parsing and prevent CLI output updates from breaking the code.  Supercedes #2112 

Since I'm touching the code and the tests, I'd really love some more eyes on this before a merge.


bewing:
Test cases removed:
* issue1168 - invalid, 4.23 uses different field that is in seconds natively
* issue1356 - invalid, 4.23+ has descriptions in `show ip bgp neighbors`
* issue1759 - may be valid, will re-add and generate new mock outputs
* issue58 - not sure what the test was here, can't find a doc

napalm/eos/eos.py
```diff
@@ -672,140 +673,84 @@ def get_interfaces_counters(self):
...
-                bgp_counters[vrf]["router_id"] = vrf_data["routerId"]
-                for peer, peer_data in vrf_data["peers"].items():
-                    if peer_data["peerState"] == "Idle":
+        bgp_counters = defaultdict(
+            lambda: models.BGPStateNeighborsPerVRFDict(
+                peers=models.BGPStateNeighborDict()  # type: ignore
+            )  # type: ignore
+        )  # type: ignore
+        # Iterate IPv4 and IPv6 neighbor details
+        for cmd in cmd_outputs[2:]:
```
ktbyers, line 693: Why do we skip the first entry in `cmd_outputs` (not apparent from the code)?

napalm/eos/eos.py
```diff
@@ -672,140 +673,84 @@ def get_interfaces_counters(self):
...
-                "local_as": napalm.base.helpers.as_number(local_as.group("as")),
-                "description": str(desc),
-                "address_family": {
-                    "ipv4": {
-                        "sent_prefixes": int(get_re_group(v4_stats, "sent", -1)),
-                        "received_prefixes": int(
-                            get_re_group(v4_stats, "received", -1)
+                    is_up = peer["state"] == "Established"
+                    description = peer.get("description", "")
+                    uptime = int(peer.get("establishedTime", -1))
```
ktbyers, line 717: Is uptime just an integer second here?

test/eos/mocked_data/test_get_bgp_neighbors/issue1759/expected_result.json
```diff
@@ -1,112 +1,70 @@
 {
   "global": {
     "peers": {
-      "fe80::a8c1:abff:fe0b:7b5f%Et5": {
+      "fe80::a8c1:abff:fe94:4171%Et3": {
+        "local_as": 4259973120,
+        "remote_as": 4259906564,
```
ktbyers, line 6: Is that the standard form that we represent AS numbers (or do we do asdot)? I don't know the answer...just wanted to make sure we are being consistent with what we do across drivers.

test/eos/mocked_data/test_get_bgp_neighbors/normal/expected_result.json
```diff
@@ -1,91 +1,91 @@
...
+        "remote_id": "192.168.0.1",
+        "is_up": true,
         "is_enabled": true,
-        "uptime": "...",
-        "remote_as": 65001,
         "description": "",
-        "remote_id": "192.168.56.3",
-        "local_as": 65002,
-        "is_up": false,
+        "uptime": "...",
```
ktbyers: Why "..." for uptime?

ktbyers (commented):
One other question, what is the difference between when "global" VRF vs "default" is used on Arista?

napalm/eos/eos.py
bewing: The idea is we want to run 4 commands.  Rather than doing 2 roundtrips, we do all 4 at once (ipv4 bgp sum, ipv6 bgp sum, ipv4 bgp neigh, ipv6 bgp neigh) and then process the pairs (both bgp sum, both bgp neigh)

napalm/eos/eos.py
bewing: Correct, uptime is an integer second in the JSON output

test/eos/mocked_data/test_get_bgp_neighbors/issue1759/expected_result.json
bewing: Standard is to return ASNs as `int`: https://github.com/napalm-automation/napalm/blob/develop/napalm/base/models.py#L185-L186

test/eos/mocked_data/test_get_bgp_neighbors/normal/expected_result.json
bewing: `...` is "don't care" in the expected_result checker:  https://github.com/napalm-automation/napalm/blob/develop/napalm/base/test/getters.py#L42

You are probably right though, I think it was intended mostly for things that change (dates, etc) not based on the input JSONs, not static test content relevant to the transformation of the input.  I was being lazy, and I'll change it.

bewing:
> One other question, what is the difference between when "global" VRF vs "default" is used on Arista?

`global` is specifically called out in the base driver as the "no vrf" key:  https://github.com/napalm-automation/napalm/blob/develop/napalm/base/base.py#L446-L447

VRF-aware Arista output uses the the key `default` for the "no vrf" key, so I do a quick rename via pop

test/eos/mocked_data/test_get_bgp_neighbors/normal/expected_result.json
ktbyers: Yeah, looking at other platforms it looks like -1 is the value we typically use for `uptime` if the neighbor is down or admin down.

ktbyers:
Everything looks good to me (after fixing the uptime to be -1 for BGP peers that are down/admin down).

One other thing I saw that was odd in the test data was the remote_id of `0.0.0.0`.

```
      "10.0.0.3": {
        "local_as": 65001,
        "remote_as": 65003,
        "remote_id": "0.0.0.0",
```

Is that just a config error in the BGP setup?

test/eos/mocked_data/test_get_bgp_neighbors/normal/expected_result.json
bewing: fixed in latest commit

bewing:
> Is that just a config error in the BGP setup?

If a peer is down, you don't know his router ID.  Same logic from before this change:  https://github.com/napalm-automation/napalm/pull/2116/files#diff-c33f60453329944c74a7bdfb2416f90a5fe0debffbe70a7aefb83a07dc6c81d0L771

Same behavior in other drivers: https://github.com/napalm-automation/napalm/blob/f0efc0bd15109f2bd6876f7adcead3e7b501dcf7/test/ios/mocked_data/test_get_bgp_neighbors/normal/expected_result.json#L68

ktbyers:
@bewing Ooops, missed that the BGP peer was down...

ktbyers:
LGTM.

ktbyers (approved):

PR #2135, opened 2024-07-30
[eos] fix banner when transport is ssh

derekdainys:
When setting transport="ssh" the heredoc commands cause an issue because they are not a string.
This allows for heredoc commands to not be converted when using "ssh" transport

### DIFF
--- a/test/eos_ssh/test_heredoc.py
+++ b/test/eos_ssh/test_heredoc.py
@@ -38,11 +38,10 @@ class TestConfigMangling(object):
             "HEREDOC conversion",
             "EOF",
             "management ssh",
-            "idle-timeout 15"
+            "idle-timeout 15",
         ]
 
         self.device.device.run_commands.assert_called_with(expected_result)
-
 
     def test_mode_comment(self):
         raw_config = dedent(
@@ -86,7 +85,7 @@ class TestConfigMangling(object):
             "This is a multi-line HEREDOC",
             "comment for standard ACL test3",
             "EOF",
-            "permit host 192.0.2.3"
+            "permit host 192.0.2.3",
         ]
 
         self.device.device.run_commands.assert_called_with(expected_result)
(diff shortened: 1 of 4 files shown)


ktbyers:
@bewing You know a lot more about this area of the code? Any objections?

bewing:
Can we add a SSH test double class to https://github.com/napalm-automation/napalm/blob/develop/test/eos/conftest.py and possibly start testing some of the differences in behavior between the SSH and eAPI transports/drivers/devices?

ktbyers:
@bewing Do we want to do this in `conftest.py` or have a separate `test/eos` and `test/eos_ssh` folder.

The latter is probably more consistent with what we did for other drivers (though we also have more code separation at the driver level between `nxos` and `nxos_ssh` (and between `iosxr` and `iosxr_netconf`).

Ultimately, I am fine with either/both (though probably have a preference for separate test folders for the different drivers+transport).

bewing:
I also think `eos_ssh` folder makes more sense.  I'll look at cribbing what's necessary from the existing _ssh folders and the standard EOS test double to arrange the boilerplate.  @derekdainys if you can actually collect the data for the unit test it would be great - let me know on slack if you need any guidance

derekdainys:
@bewing  Would like to pick this up again
Noticed that we didnt have a test for heredoc that has blank lines for example in a banner and I added it
This caused me to remove the blank line skip in `_load_config`
Added a `test/eos_ssh` folder which sets a transport=ssh and does some heredoc testing

test/eos_ssh/test_heredoc.py
```diff
@@ -0,0 +1,190 @@
+import mock
```
bewing: should be `from unittest import mock`

https://docs.python.org/3/library/unittest.mock.html

derekdainys: Thanks! Anything else you see?
Some of my builds were passing before but had I had to rebase.

derekdainys:
bump @bewing @ktbyers

bewing:
Apologies for the delays.  Tests look good, much thanks!

bewing (approved):

PR #2150, opened 2024-10-22
Junos get_config patch

SenorRodriguez:
If no format attribute is specified get_config returns config as a plain text, but in json is specified, it returns a dict. 
Probably it still needs further polishing. e.g I cast it to a str, but maybe i could stay as dict.  It's my first pull request ever, Sorry if I messed something up.

### DIFF
--- a/napalm/junos/junos.py
+++ b/napalm/junos/junos.py
@@ -2459,12 +2459,17 @@ class JunOSDriver(NetworkDriver):
         }
         if retrieve in ("candidate", "all"):
             config = self.device.rpc.get_config(filter_xml=None, options=options)
-            rv["candidate"] = str(config.text)
+            if options["format"] == "text":
+                rv["candidate"] = str(config.text)
+            else:
+                return str(config["configuration"])
         if retrieve in ("running", "all"):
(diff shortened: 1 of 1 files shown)


ktbyers:
@SenorRodriguez Can you provide some more details on the problem you are trying to solve i.e. what use case you are running into issues and how Netmiko currently behaves in that use case?

SenorRodriguez:
I was trying to write some code for validating get_config, when I noticed that getting JSON was also not possible (as mentioned in pull req  comment)  This was my attempt to fix that. Didn't even try netmiko yet.

ktbyers:
@SenorRodriguez Yeah, the Netmiko comment was just my mistake...I should have said `napalm` there.

SenorRodriguez:
I'm off my work PC, but in case I try to get a config in json format, napalm throws something like "config" dosn't have 'text' property", and sure enough - it returns dict instead of string
-edit-
dammit! I already noticed an error in my 4 lines of code :D After the the second 'else' there  should be obviously  ` rv["running"] = ...`
Sorry, if i ever make a pull request again, ill try to be more cautious :)

PR #2160, opened 2025-01-22
Support NTP server details in base, EOS

bewing:
Based on models from openconfig-system, return optional additional data regarding NTP servers for the first supported driver (EOS).

### DIFF
--- a/napalm/base/models.py
+++ b/napalm/base/models.py
@@ -223,17 +223,7 @@ IPV6NeighborDict = TypedDict(
 
 NTPPeerDict = TypedDict(
     "NTPPeerDict",
-    {
-        "address": str,
-        "port": int,
-        "version": int,
-        "association_type": str,
-        "iburst": bool,
-        "prefer": bool,
-        "network_instance": str,
-        "source_address": str,
-        "key_id": int,
-    },
+    {},
     total=False,
 )
 
--- a/napalm/base/models.py
+++ b/napalm/base/models.py
@@ -224,7 +224,15 @@ IPV6NeighborDict = TypedDict(
 NTPPeerDict = TypedDict(
     "NTPPeerDict",
     {
-        # will populate it in the future wit potential keys
+        "address": str,
+        "port": int,
+        "version": int,
+        "association_type": str,
+        "iburst": bool,
+        "prefer": bool,
+        "network_instance": str,
+        "source_address": str,
+        "key_id": int,
     },
     total=False,
 )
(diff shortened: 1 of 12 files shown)


napalm/base/models.py
```diff
@@ -224,15 +224,31 @@
 NTPPeerDict = TypedDict(
     "NTPPeerDict",
     {
-        # will populate it in the future wit potential keys
+        "address": str,
```
bewing: Adding the same keys here was a quick and dirty fix to avoid refactoring nxos's mypy typing for same.  I may revisit.

https://github.com/napalm-automation/napalm/actions/runs/12915579137/job/36017755840#step:7:14

napalm/base/test/helpers.py
```diff
@@ -4,23 +4,26 @@
 def test_model(model, data):
     """Return if the dictionary `data` complies with the `model`."""
     # Access the underlying schema for a TypedDict directly
-    model = model.__annotations__
-    same_keys = set(model.keys()) == set(data.keys())
+    annotations = model.__annotations__
+    if model.__total__:
+        same_keys = set(annotations.keys()) == set(data.keys())
+    else:
+        same_keys = set(data.keys()) <= set(annotations.keys())
```
ktbyers, line 9: We generally haven't allowed there to be a subset of keys in data. Is there a way in testing that we can specifically allow this in this case.

The equivalent of:

```python
def test_model(model, data, allow_subset=False):
```

But where we pass in `allow_subset=True`, not sure how we would actualize that in the testing code.

napalm/base/test/helpers.py
```diff
@@ -4,23 +4,26 @@
...
             "model_keys: {}\ndata_keys: {}".format(
-                sorted(model.keys()), sorted(data.keys())
+                sorted(annotations.keys()), sorted(data.keys())
             )
         )
 
     correct_class = True
-    for key, instance_class in model.items():
-        correct_class = isinstance(data[key], instance_class) and correct_class
+    for key in data.keys():
```
ktbyers: Again we are switching over from the `annotations` to the `data` which is dangerous. Can we call out the `allow_subset=True` to specifically allow this in this case?

ktbyers (commented):
I only reviewed part of the testing here...but it is a start.

napalm/base/models.py
```diff
@@ -223,16 +223,22 @@
...
+        "port": int,
+        "version": int,
+        "association_type": str,
+        "iburst": bool,
+        "prefer": bool,
+        "network_instance": str,
+        "source_address": str,
+        "key_id": int,
```
ktbyers, line 241: So these are all optional key-values which may be returned?

ktbyers (approved):
LGTM

napalm/base/models.py
bewing: That is correct.    If we get to the point where all drivers return it in all situations, we could make it non-optional.

PR #2163, opened 2025-01-28
use plain dict in IOSDriver.get_bgp_neighbors_detail()

### LINKED ISSUE #2161: The IOSDriver's BGP neighbor details cannot be pickled, which causes issues with peering-manager
Cheers,

we are running [peering-manager](https://github.com/peering-manager/peering-manager), which uses the wonderful NAPALM library to fetch data from routers.

For some reason we are still running IOS devices from the stone age in some locations, which causes us a little issue - NAPALM is of course perfectly fine talking to these devices, as the built-in `IOSDriver` "just works" (thanks again to all the wonderful people contributing to this great piece of software!).

BUT peering-manager then wants to process that data, and stores that data in redis for caching - and this is where the trouble starts, because apparently, the data that is returned by `IOSDriver.get_bgp_neighbors_detail()` does not like to be pickled. It is my understanding that this is because it uses a defaultdict in combination with a lambda function to generate a nested defaultdict, and only named functions can be pickled, not lambdas.

As far as I can see, there are two simple solutions to this:
1) replace the lambda function with a named function:
```py
# OLD
bgp_detail = defaultdict(lambda: defaultdict(lambda: []))

# NEW
def inner_defaultdict(): return defaultdict(list)
bgp_detail = defaultdict(inner_defaultdict)
```
or
2) don't use defaultdicts at all, and create the missing dicts / list manually (as is done e.g. in the `IOSXRDriver`)
```py
bgp_detail = {}
...
for neigh in bgp_sum:
    ...
    vrf_name = details["routing_table"]
    if vrf_name not in bgp_detail.keys():
        bgp_detail[vrf_name] = {}
    remote_as = details["remote_as"]
    if remote_as not in bgp_detail[vrf_name].keys():
        bgp_detail[vrf_name][remote_as] = []
```

happy to implement this, just let me know which one you prefer.

paketb0te:
fixes #2161 by omitting `defaultdict` and creating the missing empty dicts / lists manually, when required.

### DIFF
--- a/napalm/ios/ios.py
+++ b/napalm/ios/ios.py
@@ -22,7 +22,6 @@ import socket
 import telnetlib
 import tempfile
 import uuid
-from collections import defaultdict
 
 from netmiko import FileTransfer, InLineTransfer
 
@@ -2076,7 +2075,7 @@ class IOSDriver(NetworkDriver):
         return bgp_neighbor_data
 
     def get_bgp_neighbors_detail(self, neighbor_address=""):
-        bgp_detail = defaultdict(lambda: defaultdict(lambda: []))
+        bgp_detail = {}
 
         raw_bgp_sum = self._send_command("show ip bgp all sum").strip()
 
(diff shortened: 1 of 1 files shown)


ktbyers:
Looks good.

ktbyers (approved):

Issue #2166, opened 2025-02-18
EOS 4.32+  Pending Commit Syntax Change


brandondrm:
Beginning in EOS 4.32,   the syntax for pending configurations changes has been changed when using commit confirm workflows.    
Config sessions immediately show  `"state": "committed"` for each session.   The following top level keys are present only when a config session is pending confirmation:
 - `commitTimerSessionName`  name of the session with pending commit
 - `commitTimerExpireTime`  replacement for `commitBy`

Example command:   `configure session troubleshooting commit timer 300` & `show configuration sessions detail | json`

4.31 and before:
```
{
    "sessions": {
        "troubleshooting": {
            "state": "pendingCommitTimer",
            "completedTime": 1739832190.8418295,
            "commitBy": 1739835851.8256721,
            "commitUser": "adminuser",
            "description": "",
            "instances": {}
        }
    },
    "maxSavedSessions": 1,
    "maxOpenSessions": 5,
    "mergeOnCommit": false,
    "saveToStartupConfigOnCommit": false
}
```

4.32 and later:
```
{
    "sessions": {
        "troubleshooting": {
            "state": "committed",
            "completedTime": 1739832020.1816368,
            "commitUser": "adminuser",
            "description": "",
            "instances": {}
        }
    },
    "maxSavedSessions": 1,
    "maxOpenSessions": 5,
    "mergeOnCommit": false,
    "saveToStartupConfigOnCommit": false,
    "commitTimerSessionName": "troubleshooting",
    "commitTimerExpireTime": 1739835681.1816053
}
```

ktbyers:
@brandondrm Did you test these changes on 4.32+ EOS system?

I can regression test them on an older system (since I have that available).

brandondrm:
@ktbyers  - Yes, I was testing on a pair of 7280CR3K-36S running EOS 4.32.4M, both 32-bit and 64bit variations.   I was able lock it down to a change in 4.32 after trying 4.29.10M, 4.30.9M, and 4.31.6M.   

Afterward, we found in Arista's `EOS-4.32.4M-CommandApiGuide.pdf` section 2.278 highlights the changes in 4.32.0 - as well as some changes in 4.29.1.  The 4.29.1 changes did not have a noticeable impact on our workflows.

orphu:
Thanks for this! We are looking to deploy 4.32 in our environment soon.

tlund42:
[...]

PR #2167, opened 2025-02-18
fix for 4.32 pending config session changes

brandondrm:
Potential fix for #2166

### DIFF
--- a/napalm/eos/eos.py
+++ b/napalm/eos/eos.py
@@ -318,23 +318,36 @@ class EOSDriver(NetworkDriver):
         {'napalm_607123': 522}
         """
         config_sessions = self._run_commands(["show configuration sessions detail"])
-        config_sessions = config_sessions[0]["sessions"]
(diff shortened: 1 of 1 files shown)


brandondrm:
Not sure we need to keep the  logic for error checking `commit_by` prior to 4.32.  The new data structure only allows for a single pending session name and timer - but I left the data duplication in place.

bewing:
Can we add a unit test for this change?  Reach out to me on Slack if you're not sure how to create a new one.

indy-independence:
What is the status of this work? I'm also hitting this issue

indy-independence:
I did some manual testing with this patch and I can confirm it's working both for EOS 4.32.5M and with old syntax on EOS 4.31.5M

Qwiko:
Any update regarding this?

ktbyers:
Superseded by:

https://github.com/napalm-automation/napalm/pull/2281

PR #2208, opened 2025-04-02
Fix LLDP neighbors detail retrieval on ASR9K devices

### LINKED ISSUE #2207: Fix LLDP neighbors detail retrieval on ASR9K devices
### Problem Description
When attempting to retrieve LLDP neighbor details on Cisco ASR9K devices using the `get_lldp_neighbors_detail()` method, the operation fails with an XMLCLIError. This occurs because ASR9K devices require a more specific XML path in the RPC call.

### Current Behavior
The current implementation uses a generic LLDP query:
```xml
<Get><Operational>
<LLDP><NodeTable></NodeTable></LLDP>
</Operational></Get>
```

This works on most IOS-XR devices but fails on ASR9K models with an XMLCLIError.

### Proposed Solution
I've implemented a try/except pattern that:
1. First attempts the generic query
2. If that fails, falls back to a more specific query that includes the exact node path:
```xml
<Get><Operational>
<LLDP><NodeTable><Node><Naming><NodeName>
<Rack>0</Rack><Slot>0</Slot><Instance>CPU0</Instance>
</NodeName></Naming></Node></NodeTable></LLDP>
</Operational></Get>
```

This approach maintains compatibility with all IOS-XR devices while adding support for ASR9K models.

Kani999:
The original implementation of the get_lldp_neighbors_detail method fails with an XMLCLIError exception when used on Cisco ASR9K devices. This is due to the ASR9K requiring a more specific XML path in the RPC call.

Added a try/except block that:
1. First attempts the generic LLDP query
2. If that fails with XMLCLIError, tries a more specific query that includes the exact node path (Rack 0, Slot 0, Instance CPU0)
3. Also added an informative log message to indicate when fallback occurs

This change maintains backward compatibility with standard IOS-XR devices while adding support for ASR9K devices.

Closes #2207

### DIFF
--- a/napalm/iosxr/iosxr.py
+++ b/napalm/iosxr/iosxr.py
@@ -798,7 +798,20 @@ class IOSXRDriver(NetworkDriver):
 
         rpc_command = "<Get><Operational><LLDP><NodeTable></NodeTable></LLDP></Operational></Get>"
 
-        result_tree = ETREE.fromstring(self.device.make_rpc_call(rpc_command))
+        try:
+            result_tree = ETREE.fromstring(self.device.make_rpc_call(rpc_command))
+        except napalm.pyIOSXR.exceptions.XMLCLIError:
+            logger.info(
+                "Attempting more specific ASR9K rpc call for LLDP neighbors query"
+            )
+            rpc_command = (
+                "<Get><Operational>"
+                "<LLDP><NodeTable><Node><Naming><NodeName>"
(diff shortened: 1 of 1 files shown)


Kani999:
Hi, just checking in on this PR. The fix is still relevant — the `get_lldp_neighbors_detail()` method on develop still fails on ASR9K devices with an `XMLCLIError` due to the generic RPC path.

The branch merges cleanly against current develop (no conflicts). Would appreciate a review when someone has a chance. Thanks!

Issue #2213, opened 2025-04-17
pyIOSXR connections are not closing properly on version <7.10


justinbrink:
Summary:
After IOSXR updates to versions to somewhere around 7.10 (I don't know the exact version that triggered this issue), ssh handling has changed on the Cisco side, and SSH sessions from Napalm are staying open until the thread that called it is closed (handled by `__exit__`). This shows on the Device side as an open defunct SSH session in "show ssh", and the sessions will continue to sit active until no more SSH connections are allowed. Once the python module is closed with exit(), the sessions clean. This is not an issue with the CLI module or ad hoc scripts, due to the fact that it is closed properly after each call. However, napalm connections called from inside parent daemons will persist until the parent process is terminated, potentially leaving sessions open for days or months.

Cause:
While the change was on the Cisco side to SSH handling, the root of the issue is the `close()` method on the napalm.pyIOSXR.iosxr IOSXR class. After the lock cleanup it runs the `self.device.remote_conn.close()` method, which does not actually close the session fully. The proper cleanup on the Netmiko BaseConnection (the inherited class) is `.disconnect()`, but that only runs on `__exit__()`. If the thread isn't exited, the connection never clears.

Fix:
Update pyIOSXR to use the `.disconnect()` method instead of just the direct `.remote_conn.close()`
```
    def close(self):
        """
        Close the connection to the IOS-XR device.

        Clean up after you are done and explicitly close the router connection.
        """
        if self.lock_on_connect or self.locked:
            self.unlock()  # this refers to the config DB
        self._unlock_xml_agent()  # this refers to the XML agent
        if hasattr(self.device, "remote_conn"):
            self.device.remote_conn.close()  # close the underlying SSH session
        # run netmiko cleanup for session before __exit__
        self.device.disconnect()
```

moogzy:
G'day

Any thoughts on when this can be reviewed and merged? We're currently considering whether we need to fork to solve problems we're facing the SSH session exhaustion. 

We would prefer to pull this in rather than fork though.

ktbyers:
@moogzy The PR has changes that need made on it. Someone needs to update the PR and fix these issues.

The changes needed are referenced in the PR.

ktbyers:
@moogzy Okay, this issue has been fixed/merged in:

https://github.com/napalm-automation/napalm/pull/2214/

You should be able to use that commit to fix your issue. Let us k now if you run into issues.

moogzy:
[...]

PR #2214, opened 2025-04-17
Explicitly close IOSXR connection

justinbrink:
Explicitly disconnect napalm.pyIOSXR.iosxr class IOSXR before __exit__
This is for issue #2213

### DIFF
--- a/napalm/pyIOSXR/iosxr.py
+++ b/napalm/pyIOSXR/iosxr.py
@@ -510,7 +510,7 @@ class IOSXR(object):
         if self.lock_on_connect or self.locked:
             self.unlock()  # this refers to the config DB
         self._unlock_xml_agent()  # this refers to the XML agent
-        self.device.disconnect()  # close the underlying SSH session        
+        self.device.disconnect()  # close the underlying SSH session
 
     def lock(self):
         """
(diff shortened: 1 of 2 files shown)


napalm/pyIOSXR/iosxr.py
```diff
@@ -512,6 +512,8 @@ def close(self):
         self._unlock_xml_agent()  # this refers to the XML agent
         if hasattr(self.device, "remote_conn"):
             self.device.remote_conn.close()  # close the underlying SSH session
+        # run netmiko cleanup for session before __exit__
+        self.device.disconnect()
```
ktbyers: Shouldn't the fix be to replace this (line 514):

```
# Remove
            self.device.remote_conn.close()  # close the underlying SSH session
```

With line 516:

```
# Add
        self.device.disconnect()
```

Those two things seem redundant and problematic if they both execute.

justinbrink: I agree, but I just put it in with the same pattern it hit with .close() and exit. They can both execute (and do now) but it's entirely unnecessary, since disconnect() contains the remote_conn.close() call as well. I'll update.

ktbyers (approved):

PR #2223, opened 2025-05-14
Clean up IOS-XR SSH sessions correctly address regressing from #2214

moogzy:
This addresses the regression introduced by https://github.com/napalm-automation/napalm/pull/2214 and solves https://github.com/napalm-automation/napalm/issues/2213

Without the two cleanup paths the Parmiko channel doesn't get closed correctly and it appears to make the processing hang. Combined
this cleans up the Paramiko channel and correctly closes and cleans up the SSH session on the IOSXR device.

Signed-off-by: Adrian Arumugam <adrian@a6m.au>

### DIFF
--- a/napalm/pyIOSXR/iosxr.py
+++ b/napalm/pyIOSXR/iosxr.py
@@ -510,7 +510,10 @@ class IOSXR(object):
         if self.lock_on_connect or self.locked:
             self.unlock()  # this refers to the config DB
         self._unlock_xml_agent()  # this refers to the XML agent
-        self.device.disconnect()  # close the underlying SSH session
+        # Both of the following methods are needed to cleanup the IOSXR connection successfully.
+        if hasattr(self.device, "remote_conn"):
+            self.device.remote_conn.close()  # close the underlying Paramiko/Netmiko channel.
+        self.device.disconnect()  # close the Netmiko session, inclusive of SSH cleanup if used.
 
     def lock(self):
         """


moogzy:
@ktbyers @justinbrink if you could review and merge please.

The original fix while it cleans up the SSH sessions it introduces hung processing within Saltstack and likely other systems that make use of this library. When patching as provided in the diff here, the hung processing is solved and we maintain the SSH session clean up correctly.

My reading of the netmiko code suggests this is needed as we need to clean up the Paramiko channel and the SSH session. IOS-XR is great :D

ktbyers:
@moogzy Netmiko disconnect code calls close()

https://github.com/ktbyers/netmiko/blob/develop/netmiko/base_connection.py#L2495

You probably need a separate issue with details on your problem.

ktbyers:
I guess technically Netmiko closes remote_conn_pre, but I have not seen that matter. Anyways if a separate call on .close() is need on both `remote_conn` and `remote_conn_pre`, then that fix should be in Netmiko.

```python
    def paramiko_cleanup(self) -> None:
        """Cleanup Paramiko to try to gracefully handle SSH session ending."""
        if self.remote_conn_pre is not None:
            self.remote_conn_pre.close()
        del self.remote_conn_pre
```

moogzy:
> @moogzy Netmiko disconnect code calls close()
> 
> https://github.com/ktbyers/netmiko/blob/develop/netmiko/base_connection.py#L2495
> 
> You probably need a separate issue with details on your problem.

The Paramiko cleanup is not hitting the right object as we're discussing here.

https://github.com/napalm-automation/napalm/issues/2213#issuecomment-2880567944

PR #2274, opened 2025-11-12
Fix issue with sanitized and filtered EOS configs

### LINKED ISSUE #2273: Inconsistent behavior on get_config with sanitized=True on Arista EOS
In get.config() for Arista EOS, `sanitized=True` is ignored if we also set the `retrieve` parameter.

Running napalm 5.1.0
network_driver: eos

```python
with driver(**conn_options) as device:
    config = device.get_config(sanitized=True, retrieve="running")
    running = config["running"]
    print("'removed' count with retrieve=running: ", running.count("removed"))
    config = device.get_config(sanitized=True)
    running = config["running"]
    print("'removed' coun without retrieve set: ", running.count("removed"))
    with open(output, "w") as fd:
        fd.write(config['running'])
```
Sample output:

```
'removed' count with retrieve=running:  0
'removed' coun without retrieve set:  15
```

bewing:
Fixes #2273


ktbyers (approved):

ktbyers:
LGTM

PR #2292, opened 2026-03-12
Convert from requirements.txt files to uv

ktbyers:
NOTE: I temporarily disabled the auto-formatting with black as it wanted to auto-format a bunch of files and I didn't want to clutter up the PR with this (exclude-files = "*" in pyproject.toml). I will convert over to ruff check and cleanup the auto-formatting at that point (in a subsequent PR).


poetry.lock
bewing, line 1: Can we remove this file if we're settling on UV?

ktbyers: Yeah, I removed it in one of the follow-on PRs...didn't see it was still there.

bewing (approved):

PR #2308, opened 2026-04-01
Expose pyEAPI `send_enable` via optional_args

bewing:
If using NAPALM with a non-privileged user with a custom role, allow the end-user to skip sending "enable", as it would return an error. 

This allows a locally configured read-only user with an extremely limited command set to be defined that bypasses any existing TACACS authorization and accounting, speeding up execution

### DIFF
--- a/napalm/eos/eos.py
+++ b/napalm/eos/eos.py
@@ -127,6 +127,8 @@ class EOSDriver(NetworkDriver):
         self.optional_args = optional_args or {}
 
         self.enablepwd = self.optional_args.pop("enable_password", "")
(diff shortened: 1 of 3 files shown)


ktbyers (approved):

ktbyers:
LGTM

bewing:
Just looked at the IOS driver, going to rename kwarg to `force_no_enable` for compat

ktbyers:
@bewing Okay, let me review it again (I didn't realize you were still working on it).

ktbyers (approved):

ktbyers:
Looks good.

PR #2310, opened 2026-04-08
Fix get_route_to() crash on BGP origin code suffix

surudhi:
`get_route_to` with `longer=True` issues a follow-up `show ip bgp <dest> longer-prefixes` for every BGP route it finds in the routing table, to retrieve BGP-specific attributes (AS path, communities, peer info). On EOS with BGP confederations, the asPath field in that response ends in an origin code (e/i/?) rather than a number. Taking split()[-1] and passing it to int() crashes with `ValueError: invalid literal for int() with base 10: 'e'`

Fix: Strip origin code characters from the AS path string before splitting.


bewing (commented):
Can we add a unit test for this?

bewing (approved):
LGTM

PR #2315, opened 2026-04-20
Cache EOS multi-agent in bgp details

bewing:
Whether or not a switch is running in multi-agent mode won't change in between BGP peers.


bewing:
Relates to #1573

ktbyers (approved):

PR #2321, opened 2026-05-17
fix(nxos): correct is_up/is_enabled field mapping in get_interfaces

### LINKED ISSUE #1544: NXOS: Interface's is_up and is_enabled are flipped

NXOS devices report `is_up` as the port being `admin up` (ie: `no shutdown`) and `is_enabled` as having link.

ios, eos and mos (not 1st party) report this the other way around.

- [x] Yes
- [ ] No

### napalm version
(Paste verbatim output from `pip freeze | grep napalm` between quotes below)

```
napalm==3.3.1
```

(Paste verbatim output from `show version` - or equivalent - between quotes below)

```
Software
  BIOS: version 5.4.0
  NXOS: version 7.0(3)I7(8)
  BIOS compile time:  10/23/2019
  NXOS image file is: bootflash:///nxos.7.0.3.I7.8.bin
  NXOS compile time:  3/3/2020 20:00:00 [03/04/2020 05:49:49]

```

Rad [NXOS implementation](https://github.com/napalm-automation/napalm/blob/develop/napalm/nxos/nxos.py#L897) vs [EOS implementation](https://github.com/napalm-automation/napalm/blob/develop/napalm/eos/eos.py#L480)

deep-blue-pulsar:
## Summary

Fixes #1544.

The NX-API `nxos` driver's `get_interfaces()` populated `is_up` from
`admin_state` and `is_enabled` from `state` — the inverse of every other
NAPALM driver and the spec. A port that is admin-up but operationally
down (no link) was returned as `is_up=True, is_enabled=False` — the
opposite of reality.

## Fix

Swap the variable name and dict-key assignments in `napalm/nxos/nxos.py`
so each output key reads its correct source field:

- `is_enabled` ← `admin_state` / `svi_admin_state` (else `state`)
- `is_up`       ← `state` (or `svi_admin_state` for SVIs)

No logic change — only the field-to-output-key mapping.

## Test plan

The existing fixtures (`normal/`, `vlan_interface/`) do not exercise the
bug because every interface in them has `admin_state == state`. Added a
new mocked-data scenario `admin_up_oper_down/` with two interfaces:

- `Ethernet1/1` — `admin_state=up`, `state=down` → expects
  `is_enabled=true, is_up=false`
- `Ethernet1/2` — both down → expects `is_enabled=false, is_up=false`

All 532 existing tests still pass; the new fixture is the regression test.

Verified against a real Nexus 9000v running NX-OS via NAPALM 5.1.0
(NX-API transport).
- Before fix: `{'is_up': true, 'is_enabled': false}` for an admin-up,
  link-down port.
- After fix:  `{'is_up': false, 'is_enabled': true}`.
- `mgmt0` (admin-up, oper-up) was unaffected.

### DIFF
--- a/napalm/nxos/nxos.py
+++ b/napalm/nxos/nxos.py
@@ -976,11 +976,11 @@ class NXOSDriver(NXOSDriverBase):
             interface_speed = float(float(interface_speed) / 1000.0)
 
             if "admin_state" in interface_details:
-                is_up = interface_details.get("admin_state", "") == "up"
+                is_enabled = interface_details.get("admin_state", "") == "up"
             elif "svi_admin_state" in interface_details:
-                is_up = interface_details.get("svi_admin_state", "") == "up"
+                is_enabled = interface_details.get("svi_admin_state", "") == "up"
             else:
-                is_up = interface_details.get("state", "") == "up"
+                is_enabled = interface_details.get("state", "") == "up"
             if interface_details.get("eth_hw_addr"):
                 mac_address = interface_details["eth_hw_addr"]
             elif interface_details.get("svi_mac"):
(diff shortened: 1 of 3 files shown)


bewing (approved):
LGTM

bewing:
Thank you for your contribution!

PR #2323, opened 2026-05-22
Improvements to network driver import behavior

ktbyers:
Improve get_network_driver() behavior with respect to unsafe module loading.

### Issue:                                                                                                              
get_network_driver() in napalm/base/__init__.py passes a caller-supplied string directly to importlib.import_module().

                                                                                                              
This PR adds (some) input validation to this process.
                                                                     
### Changes:                                                                                                                
                                                                                                                         
 - Add _validate_driver_name() which enforces that dotted names are restricted to the two documented namespaces          
   (napalm.<driver> and custom_napalm.<driver>) with no deeper namespace traversal allowed.
 - Add a guard that rejects not namespaced names when prepend=False, since that flag implies the caller has already supplied a fully-qualified name containing "napalm".        
 - When a dotted name is supplied, restrict the candidate list to that single module.                                     
 - Improve the tests related to this.


bewing (approved):
LGTM

PR #2326, opened 2026-07-12
Fix config templates that fail to render on Python 3

ggiesen:
## What

The shipped per-driver configuration templates fail to render on Python 3.

`snmp_config`, `delete_snmp_config`, `set_probes`, `delete_probes`,
`schedule_probes` and iosxr's `delete_users` use `dict.iteritems()`, which was
removed in Python 3, so they raise `jinja2.exceptions.UndefinedError` at render
time on every supported Python (3.10-3.14). This affects all six drivers that
ship them: eos, ios, iosxr, junos, nxos and nxos_ssh (27 occurrences across 18
templates). This PR replaces `.iteritems()` with `.items()`.

Separately, `nxos/templates/delete_users.j2` and
`nxos_ssh/templates/delete_users.j2` have a mis-nested `{% endfor %}` placed
before the `{% endif %}` that closes the enclosing `{% if user_details %}`,
which is a `TemplateSyntaxError` -- the template never rendered at all -- plus an
empty `else` branch. They are reordered to close the `if` before the loop and to
delete the whole user in the `else`, matching `eos/templates/delete_users.j2`.

## Why

Any consumer that renders these templates (e.g. `load_template`, or Salt's
`net.load_template` via `netusers`/`netsnmp`/`netntp`) gets a hard failure
instead of config. Users of `set_snmp_config`, the probe templates, etc. cannot
generate config on any supported Python.

## Testing

Adds `test/base/test_config_templates.py`, which renders every shipped config
template through the same Jinja setup `load_template` uses (a `FileSystemLoader`
over the driver's `templates` dir plus the NAPALM custom filters) and asserts it
renders under Python 3. It fails on `develop` (21 of 56 cases raise) and passes
with this change (all 56).

```
uv run pytest test/base/ -q
168 passed, 31 skipped
```

## Pull Request Checklist

- [x] Code passes `ruff check` and `ruff format`
- [x] Code passes `mypy` type checking
- [x] All tests pass locally with `pytest`
- [x] New functionality includes appropriate tests
- [x] Documentation is updated if necessary (n/a -- internal templates)

### DIFF
--- a/napalm/eos/templates/delete_snmp_config.j2
+++ b/napalm/eos/templates/delete_snmp_config.j2
@@ -9,7 +9,7 @@ no snmp-server chassis-id "{{chassis_id}}"
 {% endif %}
 {% if (community is defined) and community %}
 {%- if community is mapping -%}
-{% for comm_name, comm_details in community.iteritems() %}
+{% for comm_name, comm_details in community.items() %}
 no snmp-server community {{comm_name}}
 {% endfor %}
 {%- elif community is string -%}
--- a/napalm/eos/templates/snmp_config.j2
+++ b/napalm/eos/templates/snmp_config.j2
@@ -8,7 +8,7 @@ snmp-server contact "{{contact}}"
 snmp-server chassis-id "{{chassis_id}}"
 {% endif %}
 {% if (community is defined) and community %}
-{% for comm_name, comm_details in community.iteritems() %}
+{% for comm_name, comm_details in community.items() %}
 {% if (comm_details is defined) and comm_details %}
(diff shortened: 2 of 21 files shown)


bewing (approved):
LGTM

bewing:
Thank you for your contribution!

ktbyers:
Note, one clarification here--I am pretty sure Jinja2 Version2 had helper code that made iteritems() work on Python3. 

Consequently, it was only going to Jinja2 Version3 that caused these to break. Though Jinja2 Version3 has also been around for quite awhile...

This is potentially (at least partially) why this issue wasn't observed earlier.

PR #2330, opened 2026-07-27
Switch to dynamic versioning

bewing:
Have uv set the pyproject.toml to the value of github.ref_name

Since the publish workflow triggers on tag pushes, this will be set to the bare tag name, which we can pass directly to `uv version`


ktbyers:
LGTM.

Obviously, there is some risk of supply chain attack on the repository, but we already had that with the dynamic publishing based on tags (so doesn't really change the risk meaningfully).

ktbyers (approved):

